PR #2, opened 2018-01-26
Adding Delete/Undelete to Community site. case 3279766

judeallred:


public/deleted-projects.styl
```diff
@@ -0,0 +1,118 @@
...
+    padding-bottom: 8px
+    border-radius: 5px
+    border: 0
+    margin-left: 6px
+    
+    .down-arrow
+      margin-left: 0
+      
+      
+@keyframes slideUp
```
pketh: could you use the slideUp defined elsewhere (used for project pinning?), i dunno if it's global or not. But if it's not maybe it makes sense to make it that way so that `animation: slideUp` can be globally referenced

judeallred: How would you feel about an 'animations.styl' where we can consolidate the @keyframes ?

pketh: makes sense to me

PR #11, opened 2018-03-22
Add hello-sqlite to starter apps

jennschiffer:
on desktop (mac ff): 

screenshot 2018-03-22 15 44 08

on mobile (ios safari):

img_7953

### DIFF
--- a/public/styles/projects.styl
+++ b/public/styles/projects.styl
@@ -267,7 +267,6 @@ project-avatar-size = 42px
       white-space: nowrap
 
       li
-        width: 285px
         display: inline-block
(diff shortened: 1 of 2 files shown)


pketh:
can you move sqlite to the second position? The reasoning being that the easiest one (website) should go last because for inexperienced users it’s like the ‘ok’ button in a dialog box

judeallred:
I think that sounds like a good idea

jennschiffer:
@pketh rearranged!

judeallred:
sweet. shipit.

PR #21, opened 2018-04-10
errorpage: 404 page's stars should extend to the max width (case 3294539)

ischwarz:
This should fix up the starfield not extending to the entire width of the page (https://our.manuscript.com/f/cases/3294539/full-screen-the-black-stars-on-our-404-page).

It also updates a couple gomix.com links to glitch.com and removes the beta tag

### DIFF
--- a/public/styles/pages/error.styl
+++ b/public/styles/pages/error.styl
@@ -8,7 +8,8 @@
   font-size: 80%
   position: relative
   height: 100vh
-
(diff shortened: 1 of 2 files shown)


ischwarz:
Here's my remix: https://glitch.com/edit/#!/easy-latex
...and a helpful link: https://easy-latex.glitch.me/not-the-droids-were-looking-for

PR #25, opened 2018-04-25
Steady taurus -- reduce the bundle to 530kb and use modules exclusively

judeallred:
implements case 3303352 via https://github.com/nolanlawson/cjs-to-es6  ,  switches to using moment-mini,  and enables lodash tree-shaking and minification


judeallred:
also added in case 3302430

PR #29, opened 2018-05-02
Superb guide: reactify the footer

GregWeil:
also fixed a bug because the category id wasn't included in asprops but we still try to use it as a key


src/presenters/footer.jsx
```diff
@@ -0,0 +1,45 @@
+import React from 'react';
+import PropTypes from 'prop-types';
```
judeallred: nice use of {children}!  Super clean

PR #30, opened 2018-05-02
Superb guide: reactify user search results

GregWeil:
next up is the team search results, then the whole search page (excepting layout) can be made reacty


src/presenters/user-item.jsx
```diff
@@ -0,0 +1,38 @@
...
+  };
+  return (
+    <a href={user.userLink}>
+      <div className="item" style={style}>
+        <div className="content">
+          <img className="avatar" src={user.userAvatarUrlLarge} alt={"@"+user.login}></img>
+          <div className="information">
+            {user.name ? <h3 className="name">{user.name}</h3> : null}
+            <div className="button">@{user.login}</div>
+            {user.thanksCount > 0 ? <p className="thanks">{user.userThanks} <span className="emoji sparkling_heart"></span></p> : null}
```
judeallred: React will also render nothing if the {} returns 'false', so for example in the case of
`{user.thanksCount > 0 ? <.../> : null}`
you can also write it more concisely as:
`{user.thanksCount > 0 && <.../>`

just be careful of falsey values;  e.g. 
`{user.thanksCount && <.../>`
can render "0"

src/presenters/user-item.jsx
```diff
@@ -0,0 +1,38 @@
...
+export default function UserItem({user}) {
+  const style = {
+    backgroundImage: `url('${user.coverUrlSmall}')`,
+    backgroundColor: user.coverColor || '',
+  };
+  return (
+    <a href={user.userLink}>
+      <div className="item" style={style}>
+        <div className="content">
+          <img className="avatar" src={user.userAvatarUrlLarge} alt={"@"+user.login}></img>
```
judeallred: As we move from jadelet into react and touch all the lines, I'd like to pull things up to a good a11y standard as we go.  Thankfully we have the a11y linter to help us with some stuff.

I think better alt text here would be either `alt=""` (meaning: this image is for display purposes and should be deliberately omitted by screen readers.) or perhaps something like "Avatar image for {user.login}".   The former might be the best idea here, since it's a noninteractive thing and doesn't convey any information that the next line (the user name) doesn't also include.

@jennschiffer is more up to date than me on a11y too, so she might have some good resources for us both

src/presenters/user-item.jsx
```diff
@@ -0,0 +1,38 @@
...
+          <div className="information">
+            {user.name ? <h3 className="name">{user.name}</h3> : null}
+            <div className="button">@{user.login}</div>
+            {user.thanksCount > 0 ? <p className="thanks">{user.userThanks} <span className="emoji sparkling_heart"></span></p> : null}
+            {user.description ? <p className="description" dangerouslySetInnerHTML={{__html: user.truncatedDescriptionHtml}}></p> : null}
```
judeallred: hahahaha, dangerouslySetInnerHTML!   indeed!

Unless i'm missing something, I believe the security practice here should be 'last minute encoding', which implies to me that we should be passing user.truncatedDescriptionHtml through an html sanitizer here.  (or sanitizing it in props and labeling it as such).  

there might already be one in our package.json (we have one for markdown), but if not we can toss one in or find a different way to do this ;-P

src/templates/includes/user-item.jade
```diff
@@ -1,12 +0,0 @@
...
-    .item(@style)
-      .content
-        img.avatar(src=@avatarUrl alt=@login)
-        .information
-          h3.name(class=@hiddenUnlessName)= @name
-          .button
-            span= @login
-          p.thanks(class=@hiddenUnlessThanks)= @thanks
-            span.emoji.sparkling_heart
-          p.description(class=@hiddenUnlessDescription)= @truncatedDescription
```
judeallred: ah yeah,  Jadelet has a scrubber built into it,  so we need to add that piece back to React before we can safely insert that description here

PR #31, opened 2018-05-03
Future riddle -- popover-container.jsx

judeallred:
I was able to make a general purpose, reusable Popover container with no (visible) external state.

check out https://glitch.com/edit/#!/future-riddle?path=src/presenters/pop-overs/popover-container.jsx:47:80  for how it works, and https://glitch.com/edit/#!/future-riddle?path=src/presenters/pop-overs/project-options-pop.jsx:105:1 for how it's used

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,8 @@
     "webpack-cli": "^2.0.15",
     "lodash-webpack-plugin": "^0.11.5",
     "babel-plugin-lodash": "^3.3.2",
-    "lodash": "^4.17.10"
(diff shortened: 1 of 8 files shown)


judeallred:
This is still a WIP, it's just far enough that i wanted to pause for feedback on the design.

My next steps are to refactor all of the other popovers to use this structure (and coordinate with pirijan on the ones he's modified)

GregWeil:
I had no idea props.children could be a function, but the usage ends up really clean :+1:

PR #32, opened 2018-05-08
Superb guide: reactify team search results

GregWeil:
Nested <a> tags are not valid html, and the warning react creates can not be disabled. This seems like the cleanest workaround, though I'm not sure how a11y it is.

Looking around the most common pattern seems to be `<div onClick>` outer and `<a href>` inner. Generally there will also be a 'true' `<a href>` link somewhere in there. An example is videos on the youtube homepage where the author is a link within the video link. It all acts like a link but only the video title and author link are `<a href>`s and show the url on hover.

I opted not to do that because it seems to need extra stuff for ex pointer mouse styling and preventing inner link clicks from bubbling out to the fake link.

I'm totally open to opinions though, I'm not sure what's more accessible


judeallred:
Reading up on https://stackoverflow.com/questions/18666915/why-are-nested-anchor-tags-illegal,  it seems right that we stop nesting anchors-- they're more unpredictable than I had thought.

I think the mask approach is clever and it works well!  Something I'm confused by is how the mask knows that the hrefs "puncture" the mask but the other elements do not;  what's causing that effect? can we rely on that in all browsers?

GregWeil:
That happens because we style links/buttons/tooltips with `position: relative`, which pushes them to a separate layer (and dom ordering puts them above the mask)

It should be reliable in all browsers, but it's a hidden dependency that could trip us up in the future.

judeallred:
got it, thanks for the explanation

judeallred:
lgtm!

GregWeil:
Updated to do the onclick handler instead. Seems more 'correct' and wasn't as much extra code as I expected

PR #34, opened 2018-05-17
Superb guide: split the bundle and better js caching

GregWeil:
We now have three bundles:
- client-bundle.js is everything in src/
- react-bundle.js has react modules (react/react-dom/react-router/etc)
- dependencies.js is everything else from node_modules

Each file has the hashed js appended to the querystring in the output html file. I think it might be safer to put the hash in the actual file name, but that would leak js files. (There's a plugin to clean the output directory but that doesn't work with how we send the previous js until the new build is done)


pketh:
@GregWeil are all three of these bundles downloaded by the user when loading glitch.com? In the case of analytics, if I wanted to load a larger analytics lib that'd only be used on the team page, is there a recommended way to do so in this scheme?

GregWeil:
These all get downloaded on any page load. Webpack supports the `import('asdf')` form which returns a promise that resolves as the package, but that only works for stuff pulled from (p)npm and not external js files.
I haven't tried it out, but seems like the way to go: https://webpack.js.org/api/module-methods/#import-

pketh:
cool i'll try that out for my charting lib

routes.js
```diff
@@ -1,27 +1,14 @@
...
-  FACEBOOK_CLIENT_ID = "1858825521057112";
-} else {
-  APP_URL = 'https://glitch.com';
-  API_URL = 'https://api.glitch.com/';
-  EDITOR_URL = 'https://glitch.com/edit/';
-  CDN_URL = 'https://cdn.glitch.com';
-  GITHUB_CLIENT_ID = "b4cb743ed07e20abf0b2";
-  FACEBOOK_CLIENT_ID = "660180164153542";
-}
+const API_URL = ((process.env.RUNNING_ON === 'staging')
```
judeallred: why calculate this here instead of in webpack like the others?

webpack.config.js
```diff
@@ -24,9 +44,26 @@ module.exports = () => {
...
+          react: {
+            name: 'react-bundle',
+            test: /[\\/]node_modules[\\/]react[-\\/]/,
+            chunks: 'all',
+          },
+          modules: {
+            name: 'dependencies',
+            test: /[\\/]node_modules[\\/]/,
+            chunks: 'all',
+            priority: -1,
```
judeallred: why priority -1?

webpack.config.js
```diff
@@ -24,9 +44,26 @@ module.exports = () => {
...
-      filename: '[name].js',
+      filename: '[name].js?[chunkhash]',
       path: PUBLIC
     },
+    optimization: {
+      splitChunks: {
+        cacheGroups: {
+          react: {
+            name: 'react-bundle',
+            test: /[\\/]node_modules[\\/]react[-\\/]/,
```
judeallred: why bundle react separately?

GregWeil: React-dom in particular is half of our dependency weight, this division results in two bundles of about 600kb each (in dev, not sure about prod). It gets us a bit closer to avoiding the 'large chunk' warnings and is less for users to redownload when we add new modules

webpack.config.js
GregWeil: That way modules prefer any other group over that one. In this case it just means they prefer the react bundle, but it's more flexible if we add more groups in the future

routes.js
GregWeil: We use the api url in the server code to cache teams/categories. The html file gets that url from webpack.

PR #35, opened 2018-05-17
Future riddle -- reactifying the user profile

judeallred:


judeallred:
profile.jsx is the big focus here, but there's some other fixes in play:

 -> anon user backgrounds weren't rendering, added that
 -> team avatars and profile covers wouldn't update has[avatar,cover]image
 -> categories would render a little strangely, i nudged them toward json a bit
 -> .anon-user-avatar was unused, kill it
 -> feature: added a 'clear cover' button.   (clear avatar works too, but i left the button off since it doesn't fit well.)
 ->extracted ANON_AVATAR_URL to a const on user.js; use that as a fallback url if images fail to load

src/presenters/random-categories.jsx
```diff
@@ -47,12 +47,11 @@ export default class CategoryContainer extends React.Component {
...
   
   componentDidMount() {
-    this.props.getCategories().then((categoryModels) => {
-      let models = categoryModels.filter(model => !!model.projects.length);
-      models = sampleSize(models, 3);
-      
-      const categories = models.map(categoryModel => {
-        const {...category} = categoryModel.asProps();
+    this.props.getCategories().then((categoriesJSON) => {
+      const categoriesWithProjects = categoriesJSON.filter(category => !!category.projects);
```
GregWeil: `!![] === true` so this new check will only remove categories with no array at all

GregWeil:
I left a comment on the by file view but it doesn't show up here?

The change from `!!category.projects.length` -> `!!category.projects` won't work because empty arrays are truthy (unlike empty strings)

src/presenters/random-categories.jsx
judeallred: correct, and that matches the api response.  It is confusing though, as that's different from the rest of what's up...  i'll add a comment

PR #38, opened 2018-05-29
Future riddle -- improved oauth error pages

judeallred:
for case 3312135

we now display an error page instead of a blank white screen, and we log the oauth failure details to the console.


src/client.js
```diff
@@ -25,129 +23,154 @@ console.log("#########");
...
+  // OAuth Handling
+  //
+  if (normalizedRoute.startsWith("login/")) {
+    const provider = normalizedRoute.substring("login/".length);
+    const code = queryString.code;
+   
+    return application.login(provider, code)
+      .then(() => {
+        history.replaceState(null, null, "/");
+        window.location = "";
```
GregWeil: (This was mostly here before your refactors)
I'm not sure what's supposed to happen here, could it just use `location.replace('/')`?

PR #39, opened 2018-05-30
Nova chance: refactor css and class names for team/user pages

GregWeil:
shuffled around the jsx a bunch and consolidated css. main highlight is the flatter css

there's now 3 main parts at the top of the profile page:  the cover-container, the profile-info container (the gray box with a max width pushing it to the left size) and the avatar-container (splits avatar|content)

each one's styling is top level, so they can be used independent of each other


judeallred:
lgtm, great cleanups!

judeallred:
...'close' was the wrong button...

judeallred:
:shipit:

PR #40, opened 2018-05-31
Superb guide: promote projects from the overlay to a page

GregWeil:
looks like https://www.are.na/block/2237044 but without the pinned projects underneath

Next up I want to bring back the non embed readme so the embed can start in app mode.
After that markdownify the description (and make it editable on the page?).
Then start checking out related projects or stats, will need to start exploring the api beyond just using what things are already calling.

Side note: the view/edit/remix buttons all open in the current tab. Should I switch them to open in a new tab so you don't lose the community site context?


pketh:
@GregWeil are the additional changes you mentioned (description, etc.) going to be included in the scope of this PR? or later?

GregWeil:
Later. This is enough to match the overlay's functionality (plus the description) so I figured adding stuff can happen separately

pketh:
looks great, just a couple tiny design tweaks

artboard

GregWeil:
- Refactored the project buttons to share more code
- Standardized the margin between each section
- Changed 'preview' to 'show'
- De-emphasized the feedback button

I left view source/edit project as is (rather than renaming to fullscreen like the embed) because I think it could get confusing with some apps. For ex clicking fullscreen on the page for a game I'd expect to get the game running fullscreen, not the source code for it.

GregWeil:
Oh I also started markdown rendering project descriptions across the site. It's a bit confusing right now because you can only edit descriptions in the editor which does not render the markdown, but it means urls are linkified and :asdf: emojis work which is nice

pketh:
Looks like you still need to remove the remix button

pketh:
Can we also change the ‘what do you think?’ Button to ‘report abuse’? I remember people complaining that’s it’s unclear and we want to encourage non critical convos to happen on the forum instead

GregWeil:
- Removed remix (and removed the conditional highlight on edit project)
- Renamed 'what do you think' to 'report abuse'
- Wording changes to the feedback template to match the button
- Added a lock icon to the page if the project is private

pketh:
got a mixed content error for the readme at http://superb-guide.glitch.me/~space-invaders

screen shot 2018-06-04 at 10 57 31 am

pketh:
now that the readme extends page content, I think it'd make sense to move the `report abuse` button to the bottom of the page (below the readme)

pketh:
also what do you think about moving the `buttons` into the `profile-info profile-information` (or an adjacent section called `profile-actions` maybe)?  I think it makes the page easier to parse. And will probably be a closer match to how we do a 'follow' button on a user profile page in the future.

screen shot 2018-06-04 at 11 02 29 am

GregWeil:
The readme failing to load seems like an api bug. If you are signed out it gives a 'you need the authorization header' error and omits the cors header.

The space to the right of the profile seems like a good place for _something_. Maybe the buttons + owner config that doesn't fit inline (ex make private)? For now I'll put the buttons in the profile space like your mockup since they do look nicer there than floating on their own

GregWeil:
- Render readmes on project pages
- Moved feedback button to the bottom
- Move show/edit buttons into the profile area
- The embed starts as app only
- Added a new DataLoader component, which makes a request and shows a loader until it is done. Only used on the project page and readme right now, but should be applicable elsewhere

pketh:
That's weird, the readme endpoint should not need authorization :(

anyways, this lgtm. The only change for now I'd suggest is changing the `readme` and `embed` sections to `margin-top: 2rem` (it looks a little less cramped, and matches the <section> spacing on the homepage)

one way to accomplish this is to remove the generic section override ```
.project-page > section {
    margin: 1rem 0 ```

in favor of ```
.info {
  margin-top: 1rem
}
```

**Comment by GregWeil** [comment]
Removed the extra reduced space css. I originally reduced it because the buttons looked bad with so much space around them so it makes sense to bring it back now.

**Comment by pketh** [comment]
I think you still want to reduce the space above the info container and the header , to match the homepage distance between the glitch logo and the introductory text

eg ```.info {
margin-top: 1rem
}
```

pketh:
yup good catch, actually lets leave that alone then. it looks crowded as 1rem in that context and i'd rather be consistent

pketh:
screen shot 2018-06-04 at 1 44 06 pm
I'm also noticing that some project avatars have weird wrapping visual glitches

`background-size: contain` fixes this issue

GregWeil:
That doesn't repro on windows chrome, but it does change how non square avatars render (wide/tall images get empty space around them rather than being cropped)

pketh:
what does `background-size: cover;` do?

GregWeil:
Both `cover` and `contain` preserve the aspect ratio. For an image that is wider than tall, `cover` would cut off the left and right sides of the image so it appears square, while `contain` will shrink the image to fit entirely within the square, with empty space above and below the image to fill out the square.
(the test-team avatar shows it off pretty well)

`no-repeat` is already set (And it doesn't repro on my end), so I think this is a browser rendering bug.

pketh:
👍

PR #41, opened 2018-06-01
Future riddle:  User-editable name and login fields

judeallred:
Most of the way there! play with it at:
  https://future-riddle.glitch.me/@gregweil
  https://future-riddle.glitch.me/@pketh

missing is some additional input validation (emanuele is adding this on the api side)...   so if you hose your account while you're playing with it, here's a console command you can use to fix yourself.
`application.currentUser().updateUser(application, {login:`fixed`})`

image

image


GregWeil:
My display name field was previously unset, when I go to my user page I get

Warning: Failed prop type: The prop `value` is marked as required in `EditableField`, but its value is `null`.

and the name field starts set to the string 'null'

pketh:
imo think we should use the editor `field-errror-message` and `input field-error` styles for input errors
- used in a similar context in the editor (and assets)
- uses standard error colors
- more visually tied to the input

judeallred:
Agreed, I prefer keeping the UI consistent! I wasn't able to find the precedent for it, so thanks for the screencap

judeallred:
there are some lingering api bugs: https://our.manuscript.com/f/cases/3313580/limits-for-login-on-person-entity  but the frontend part of this should be done

GregWeil:
It's still getting an error for my profile (login only, no name) which causes the field to default to 'null'

```Warning: Failed prop type: The prop `value` is marked as required in `EditableField`, but its value is `null`.
    in EditableField (created by NameAndLogin)
    in NameAndLogin (created by LoadedUserProfile)```

**Comment by pketh** [comment]
some little css issues:
-  rounded corners at the ends of the line 
- left alignment is 10px in from edge, unlike textarea or the thanks line
```
.content-editable
  border-radius: none
  padding-left: 0 // (i think the padding-left: 10px on input was for the search box)
```

**Comment by pketh** [comment]
instead of putting the `@` in the field and having logic to prevent it's deletion, maybe it'd make sense to put it right before the field instead? eg
`@__________`

**Comment by judeallred** [comment]
@greg fixed;  my mistake was that I thought react default values would handle 'null', but they do not- only undefined.
@pketh  styles fixed;  I'll play with the @ some more.  It's really hard to get the positioning right if it's anything but part of the input... I'll see what else I can come up with.

**Comment by pketh** [comment]
coolness, lemme know when that last bit is done so i can re-review :)

**Comment by judeallred** [comment]
okie doke, ready to be played with:
https://future-riddle.glitch.me/@JudeAllred
https://future-riddle.glitch.me/@pketh
https://future-riddle.glitch.me/@gregweil

**Comment by pketh** [comment]
screen shot 2018-06-07 at 11 47 09 am
 misalignment of  🚒on safari

**Comment by pketh** [comment]
you already poke garethx about this upcoming change? (prob some faq and support forum updates)

**Comment by pketh** [comment]
maybe in this pr or seperately you might also want to prepare a new-stuff-update about this change, it's likely a big one for a lot of users

File: src/presenters/includes/prefixed-input.jsx [hunk 1]
```diff
@@ -0,0 +1,23 @@
...
+    return <input {...props}/>;
+  }
+  
+  const className = props.className || "";
+  const inputId = props.id || uniqueId('prefixed-input-');
+  return (
+    <div style={{display: 'flex'}}>
+      <label htmlFor={inputId} style={{width: 'auto'}} className={className}>{prefix}</label>
+      <input id={inputId} {...props}/>
+    </div>
```
GregWeil: This could instead put the label as the wrapper, and use a span around the prefix text for flex support, so you wouldn't have to override the jsx rule.

As a side stylistic thing, I'd rather see PrefixedInput as a wrapper element that takes an input element via children. That way it wouldn't be included in every input when it's really only needed in one place.

GregWeil:
fire engine is low on windows chrome too, but it works really smoothly, having the url update when you rename is nice, though it does feel odd that the links to your profile around the page become broken.

Not sure how that could be fixed, though it is worth thinking about since it's something the observable pattern handled pretty naturally

judeallred:
oof, good catches, thanks.   the prefix input broke the fire engine :p

src/presenters/includes/prefixed-input.jsx
judeallred: 1--  good idea, will do
2-- also good idea...  need to figure out how to inject the id prop. looks like https://stackoverflow.com/questions/32370994/how-to-pass-props-to-this-props-children 

i can put the styles into stylesheets... i was thinking to myself "these are dedicated functional styles for a specific element, so maybe they should live here?"...   but I probably shouldn't step into that precedent lightly.

fixin'

GregWeil:
Seems good! you can still create conflicts with team names, but that's an api side issue

PR #42, opened 2018-06-04
Nova chance: Show placeholder on empty description fields

GregWeil:
empty markdown now renders as `<span></span>` which breaks the `:empty` check used to show the placeholder. Now it will render completely blank


GregWeil:
Updated to disable smart quotes

PR #44, opened 2018-06-06
community-new-analytics: New analytics

pketh:
https://www.are.na/block/2272620

noodle around with it at https://community-new-analytics.glitch.me/test-team


pketh:
@garethx just a heads up that we have a new analytics ui ready soon in case you need to update any docs, etc.

judeallred:
if i view the page while i'm not logged in, i see some errors in the console and the loaders spin forever: 
image

judeallred:
not sure if intentional, but looks like spacing is missing between the buttons here?
image
(chrome/windows)

judeallred:
the spacing on this arrow feels a little off:
image

judeallred:
image
some quirks--
  the totals up top, "21471 app views; 20 remixes" don't mach the details down below: app view, remixes.    The nature of the display (a chart and then a table) makes me really expect them to line up

because app views dwarfs remixes by so much (and I suspect this will generally be true),  putting them both on the same graph kind of puts 'remixes' in trouble because it comes though as a flat line.   Could you make it so that clicking on the 'app views' or 'remixes' at the top of the chart will toggle that line and redraw the scale?

judeallred:
can you split part total app/code views from unique app/code views?  I remember our meeting with mouse where they only wanted the distinct viewers and we gave them the total, which distorted their numbers a lot because of all the visits from their internal people.

(hmm, another neat metric might be visits by members/non-members...)

pketh:
re toggling scales: you can toggle the chart by clicking the legend, i could try manually triggering the click on the c3 element if you click on the totals too

pketh:
re the totals count disparity mentioned in https://github.com/FogCreek/Glitch-Community/pull/44#issuecomment-395573282

re unique views mentioned in https://github.com/FogCreek/Glitch-Community/pull/44#issuecomment-395573717

@etamponi I know you've mentioned it before, but it's unclear from the api property names. Can you elaborate on what's going on/possible with these issues?

judeallred:
quirk in the 'latest remixes' section
steps:
 1. remix a project
 2. notice it appears in the section (yay!)
 3. delete the remix
 4. 'latest remixes' says "no remixes yet"

expected
  ...I still want to know that the remix happened,  even if it's deleted.
actual:
  it says "no remixes yet", which isn't true - nobody remixed my project :sadpanda: (but truly they did!)

judeallred:
what would you think about putting the user list to the right of project details?  like maybe as a row of usernames to the right of the project icon?

judeallred:
re: toggling -- oh cool!   Yeah that's exactly the behavior i was looking for. Since the colors match up I expected that clicking either would work, and I tried the big ones up top first.

pketh:
re recent project userlist: to do this with the data I currently have, I would have to do a /project GET for each project, which would certainly add up

judeallred:
got it.  seems ok if you punt on it, but since it's in response to a user action "tell me about this project" instead of just fetching all of them up front, it'd probably be ok to send those requests

GregWeil:
image
Remix name tooltips are all the way on the right if you hover over the first remix in any line after the first

pketh:
new fixes/features:

- analytics no longer renders if not on team
- added default project fallback src on broken images https://www.are.na/block/2301417
- added spacing between buttons
- can click summary data to toggle graph views https://www.are.na/block/2301286
- moved `↑` alignment upwards in project details message
- added self-referrers data to explain the discrepancy between totals and referrers (previously was only showing referrers that came were actually referred from another place)

re latest remix quirk `https://github.com/FogCreek/Glitch-Community/pull/44#issuecomment-395575118`:
if the remix is deleted it's not returned in the endpoint. For our goal of analytics being more about vanity/feel-good-roi over deep data diving this seems sufficient. Also this particular 1 or 0 scenario sounds like an edge case

re 'tell me about this project' `https://github.com/FogCreek/Glitch-Community/pull/44#issuecomment-395580232`:
it's simpler to have the ~/project page be the source of truth on that, which shows the user list

re tooltips `https://github.com/FogCreek/Glitch-Community/pull/44#issuecomment-395861708`:
 ya that's a weird one, maybe greg's new pr addresses this though

@judeallred @GregWeil could you give this another review when you get the chance? thx!

pketh:
todo: 
- merge conflicts
- mobile fixes

GregWeil:
If a project has never been remixed, the 'last remixed' date is 'a few seconds ago'

GregWeil:
a few small comments:
- projects with no remixes were last remixed 'a few seconds ago'
- original remixed from should include the project title in the link, not just the avatar
- in the four week view the x axis shows every other day. is that a data limitation, and if so can you make it more clear where the unlisted days fall?

GregWeil:
also there seems to be some flicker on adding/removing projects to teams

pketh:
- re: projects with no remixes were last remixed 'a few seconds ago', that appears to be an api error. Because the rest of the analytics reports 0 remixes and because someone wanting stats on 0 remix projects in a teams context is going to be rare, I'm leaning towards not fixing for now.

i'm looking into fixing the others

GregWeil:
This is probably an api bug, but the last 24hrs analytics seems to only have the first datapoint filled out. 12:40PM yesterday is ~27000, but everything from then until now is flat 0.

GregWeil:
I'm having trouble reproducing it, but I've had it happen a few times where the analytics section never loads, and the network tab shows that the analytics api request never happened. It seems fine 9/10 times, but maybe there's a race condition somewhere?

pketh:
this fix includes performance, race-condition, a11y fixes

specifically also fixes: graph should always load data now

pketh:
re 24hours analytics bug: 
I can't seem to repro it, switching to 24 hour view gives me the expected graph

screen shot 2018-06-20 at 11 54 40 am

do you have steps to repro?

GregWeil:
Nope, it did happen every time and does not happen now.

Seems good, but I would like to get the seo deploy fixed before anything else, in case we need to roll back again

PR #45, opened 2018-06-07
Familiar text: add proxying for glitch.com/help/ case 3312768

judeallred:


judeallred:
hey github bot, please tell the chatroom that there's a new PR here.  Go go magic github....

judeallred:
@GregWeil  this seem like an ok strategy to you?

GregWeil:
It seems to work pretty well, but why is this page treated differently than /about or other 'glitch project as route' pages?

Also, could you update the /faq link in the footer to /help? /faq is a redirect now.

judeallred:
footer: good spot, fixed

why is it different: we're dipping our toe in;  previously we were doing routing with cloudfront config (see aws/production.yml in the Glitch repo) and this was really painful for gareth and daniel.  This is an experiment to see if we can just do all the routing here instead.

Signs point to yes.  A next task here is to pull in all of the other special routes (...which is actually just about.glitch.me and a few dozen routes pointed at it) except for the editor itself.

since they're all mostly static content and we've got cloudfront caching ahead of them all, this should be inconsequential in terms of load.

So all that said.. the idea is to start small with this route.  Seem ok?

judeallred:
here's the routing file: https://github.com/FogCreek/Glitch/blob/master/aws/production.yml
here's the change that was made for /help/  https://github.com/FogCreek/Glitch/commit/9a502cb1d3d6998447e8ce5119d743d00a90e29d

all of the entries with TargetOriginId: AboutOrigin  would be on the chopping block next, if we wanna proxy those too

GregWeil:
In that case this seems all good

judeallred:
awesome, will deploy monday @garethx

PR #46, opened 2018-06-07
Nova chance: Related projects v1 'more by user'

GregWeil:
image

Picks a team and user on the project to show three random pinned projects from. If the user/team has no pinned projects then the cover area is omitted but they keep the 'more by' link. If the project is not associated with any teams then two users are shown.

also some fixes for project list styling


src/presenters/pages/project.jsx
```diff
@@ -105,9 +112,12 @@ ProjectPageLoader.propTypes = {
 // Let's keep layout in jade until all pages are react
 export default function(application, name) {
   const props = {
-    get: () => application.api().get(`projects/${name}`).then(({data}) => (data ? Project(data).update(data).asProps() : null)),
+    get: () => application.api().get(`projects/${name}`).then(({data}) => console.log(data)||(data ? Project(data).update(data).asProps() : null)),
```
judeallred: debuggin' line, i'd guess?

pketh:
a 'more by' for each team up to like n(3?) teams seems like it'd be a good middle ground

src/presenters/pages/project.jsx
GregWeil: whoops, will remove

pketh:
if you're feeling a bit extra ambitious could you also add a 'remix this' button right below the embed based on https://our.fogbugz.com/f/cases/3314902/remix-button-for-the-project-page#BugEvent.21217400 ?

thx

GregWeil:
The intended behavior if a project is part of multiple teams is that it'll pick one at random to show

Will add the remix button

pketh:
i'm noticing that when i hover over user avatars in http://nova-chance.glitch.me/~community-new-analytics , they no longer show up?

pketh:
it looks like this is a regression on prod as well

GregWeil:
remix button added.

Tooltips don't show up anywhere that avatars are constrained to one line. We use a fixed height and overflow: hidden, which cuts off the tooltip. There's probably a better way to do that?

pketh:
how were we doing it before and why did we change? Why does the avatar line need 'overflow: hidden'?

GregWeil:
not sure when, but I think the reason was so projects with a lot of members don't push adjacent projects around

GregWeil:
ok I have a fix, will add to this

src/presenters/pages/project.jsx
```diff
@@ -74,20 +88,29 @@ const ProjectPage = ({
             authorized={userIsCurrentUser} description={description}
             update={desc => updateDescription(id, desc)} placeholder="Tell us about your app"
           />
-          <p id="buttons"><ProjectButtons domain={domain} isMember={userIsCurrentUser}/></p>
```
pketh: are these `id`s being used for anything?

pketh:
on mobile the info box buttons seem like they need margin-top seperation?

Or maybe they should render from outside the container on small screens? 
Based on media queries for `smallViewport` or `460px` you could have two sets of identical `buttons` groups on the page: one in the `profile-information` container, and another outside of it. And toggle which to display:none with a media query

i think either approach makes sense, both involve media queries

screen shot 2018-06-08 at 1 33 44 pm

src/presenters/pages/project.jsx
GregWeil: No, I put them there for `#links` but on testing now it seems like they don't really work with client rendered stuff

pketh:
also height differences between the two profile info buttons (might also be caused by class differences between the overlay versions and these ones)

screen shot 2018-06-08 at 1 56 20 pm
screen shot 2018-06-08 at 1 56 25 pm

GregWeil:
Added a remix button, fixed button sizing/spacing
Project tile user lists still hide tooltips, they need a bit more investigation.

PR #47, opened 2018-06-08
Somber crane

judeallred:


judeallred:
@etamponi created a PR so i could review the diff :-)

judeallred:
I added a couple touchup commits, see above @etamponi

judeallred:
In this branch, i see a bunch of users who have generic avatars now.

Is that a bug, or are these users who aren't backfilled, or something else?

image

etamponi:
I backfilled another 1000 avatars :) We should be good to go with this change now.

judeallred:
case 3302440 -- editable profile avatars!  paired with emanuele

PR #49, opened 2018-06-13
Nova chance: Edit project name and make it private

GregWeil:
image
image
image

Not included in this pull: upload project avatar, add/remove members
For the future:
Should project pages have direct member adding, or match the editor with join requests and invite links?
Should project titles always be shown with ~ in front like user logins are? (what about teams)


GregWeil:
The embed also gets reloaded when you change the name, which I figured was preferable to broken links.

pketh:
looks super cool :), just a couple design bits

readme: 
- because you're already reading it from left to right having the button centered is a bit awkward. 
- it'd be way easier if clicking anywhere in the collapsed readme expanded it (not just the button)
screen shot 2018-06-14 at 7 46 25 am

privacy:
- i don't think the background should change to yellow once you click this button. The bit color change transition feels awkward on click because of the button affordance/border of the thing you're clicking on. how does it feel if just the button changes?
- the privacy button being yellow alone will be visually prominent enough for ppl to notice it if that's the goal
- the privacy button doesn't match colors with the button on the editor in the idle or hover state
screen shot 2018-06-14 at 7 51 16 am
screen shot 2018-06-14 at 7 51 22 am

error state:
- 'domain in use' doesn't uses our terminology (domain) and doesn't give the user an actionable alternative. Maybe 'Name is already used, try another'
screen shot 2018-06-14 at 7 58 37 am


re questions:
> Should project pages have direct member adding, or match the editor with join requests and invite links?
- for abuse prevention, direct member adding without a verification step (either in real time like the editor, or through email) should not be allowed. e.g. i can add user X to my projects about how much I hate X, and those would show up on Xs profile. 
- In the future, email based invites will be added to the editor as well, with a ui probably similar to the email based add a user functionality on the teams page spec.

> Should project titles always be shown with ~ in front like user logins are? (what about teams)
- I don't think they should on the project page (but should on the teams page):
- the teams page should eventually have a name and login/domain as distinct fields just like users do. lots of orgs want spaces or weird characters (`*` `.`) or specific casing in their names that don't work in a url.
- on the teams and user page the @ makes sense for domain/login because it's more about differentiating between the permissive full name fields and the shorter login/nickname (ie 'pirijan ketheswaran' vs '@pirijan'). Perception wise, it's not as much about url as it is about identity (it's a subtle distinction)
- on the project page, there is no alternate name field, and perception wise it's also more about the identity of your project name than it's url. 
- additionally, no `~` matches and makes sense in the editor which also shares that same perception

pketh:
re error messages, can community use it's own error messages based on the status code we hear back from the api?

GregWeil:
Every error has the same 400 status code.

pketh:
made a pr to pass good errors to this https://github.com/FogCreek/Glitch/pull/661

pketh:
the animation when revealing the full readme, could you reduce this to 200 or 100ms? In general, here's my thinking about animation:
- it has to be purposeful (and it is here, because it connects the action of the button to the feeling of expanding)
- the animation shouldn't draw attention to itself, but rather feel like it's 'reacting' immediately to the user's intent
- it should be designed as if it's being seen for the 1000th time (i.e. would it be annoying to wait for the rest of the readme if you're used to the animation)

pketh:
because of it's context underneath the project name, the privacy tooltip text could be shortened:
- only members of [project-name] can see it's code -> "Only members can view code"
- [project-name] is visible to everyone -> "Visible to everyone"

(later i'll add those tooltips to the editor too)

src/presenters/pages/project.jsx
```diff
@@ -76,19 +90,34 @@ const ProjectPage = ({
...
+  updatePrivate,
 }) => (
   <main className="project-page">
     <section id="info">
       <InfoContainer>
         <AvatarContainer style={{backgroundImage: `url('${avatar}')`}}>
-          <h1>{domain} {project.private && <PrivateBadge domain={domain}/>}</h1>
+          <h1>
+            {(userIsCurrentUser
+              ? <EditableField value={domain} update={updateDomain} placeholder="What's it called?"/>
```
pketh: 'it' might be too unspecific if a user just stumbles around into a project page and deletes the name. Maybe a more actionable placeholder would be "Name your project"

src/presenters/includes/expander.jsx
```diff
@@ -0,0 +1,62 @@
...
+        )}
+      </div>
+    );
+  }
+}
+
+Expander.propTypes = {
+  children: PropTypes.node.isRequired,
+  height: PropTypes.number.isRequired,
+};
```
pketh: is it possible to do the expander based on css instead?
pseudocode:

html
```
<markdown-readme onClick=expandMe>
```

js
```
expandMe = (event) => {
  // $(event.target.src).addClass('.expanded')
  // there's probably a better react way to add classes
}
```

css
```
.markdown-readme
  max-height: 250px
  &.expanded {
    max-height: 100%
    button {
      display: none
    }
  }
```

GregWeil: For the unexpanded height yes, but transitions don't work for auto (and max-height 100% would be relative to the parent element, aka the height of the page). That's why the expander does the scrollHeight calculation on click.

I opted to make this a separate thing since we might use it elsewhere in the future

GregWeil:
Made all of the wording changes and sped up the animation
Also did some refactoring around related projects (now each block loads individually)

GregWeil:
small addition to ignore cachedUser if it looks like junk data, and to make sure login responses are actually logged in

judeallred:
cool!

PR #50, opened 2018-06-19
Nova chance: very basic server side rendering for seo

GregWeil:
also shuffled around server js. it lives in /server now

old:
image

new:
image

also supports `@user` urls. No teams support, can add when we support `@team` urls


judeallred:
LGTM!  One next step to consider would be adding in a sweet basic cache like https://www.npmjs.com/package/memory-cache for those project fetches.  I think that even very minimal caching would have a tremendous impact on the scale-ability of the site.

PR #59, opened 2018-06-29
Nova chance: react team page, other cleanups

GregWeil:
the user page is still totally busted, I just wanted to get this in place for feedback now that teams are 100% functioanl before I got too deep
https://nova-chance.glitch.me/test-team

Basic design is that at the bottom of team.jsx there's `TeamPageLoader`, which makes the api request for the team and sends it through `EntityEditor` and `Uploader`.

`EntityEditor` defines some generic methods for `PATCH`ing fields and updating the state to match. `Uploader` wraps assets.js, and renders the progress notifications.

The stateful team json from `EntityEditor` plus the generic edit and upload functions go to `TeamPageEditor` which defines the actual team editing methods used in the page (for example `() => updateFields({hasCoverImage: false})`).

Lastly the team and team methods go to the TeamPage which actually renders everything to the page.

This feels like a decent way to spread out 'render stuff' 'edit stuff' and 'gather stuff', but the user page has a few new problems that were handled for us with the models/observables but need to be solved again in a more reacty way.

Edits being made don't go beyond the EntityEditor for the page, so stuff in the header never sees the changes (even if they were in the same react hierarchy). We could shim something for the current user to use a larger editor around everything, but that wouldn't handle situations like removing yourself from a team, which does not happen through the user.

The current user gets loaded several times as part of unrelated requests (for ex getting their list of projects). We have to do some kind of normalization so that user lists use the current user rather than treating it like any other user.


GregWeil:
My current strategy idea:

Expose an 'updatecurrentuser' function alongside the current user object, which simply replaces the entire object. The user page editor will make sure to call that after any change gets made, and the team page can add a bit of logic to the 'removeuserfromteam' func to update the user if they remove themselves from a team.

Normalizing users isn't that bad (we can just map them through a func that returns the current user if the ids match) but it is an extra step that has to be repeated in several places with a very subtle bug if we forget.

pketh:
if `updatecurrentuser` replaces the entire object is there a risk that it could remove some data for the current user that we get on /boot , but not from a normal /user call. looks like there's some extra stuff for billing being adding here too

/boot user info
screen shot 2018-06-29 at 3 14 44 pm

authenticated /user/<id>
screen shot 2018-06-29 at 3 15 10 pm

GregWeil:
Good catch! In that case lets treat it more like setState (or maybe have it call setstate indirectly)

pketh:
you mean the react `setState` where we're updating state properties individually that we expect to change after the GET user/id ?

pketh:
i wonder if there's like a `replace the things that are different in this object only` type of thing ?

pketh:
(stuff like this, and other issues i'm having with teams, makes me reeeally wish we had a graphql api :( )

pketh:
re
> Edits being made don't go beyond the EntityEditor for the page, so stuff in the header never sees the changes

scenario where this will feel janky is when you leave a team, open your user options menu from the header, and see the team you've left listed there. Is there a reason to isolate state to the entity-page? Is this a potential use case of react contexts? (https://reactjs.org/docs/context.html)

judeallred:
just playing with the app, it's SO MUCH FASTER... that's really nice.

GregWeil:
the tricky thing about leaving a team is that the page is operating on the team object, and not the users in it (aka remove user affects team.members, the header uses user.teams) so we have to update two things

As the team page gets more features (renaming) there will be more stuff that has to get to the current user than just the presence of the team, so I don't know how well it will expand to that.

I think context makes sense for the current user and the api, but I'm iffy on it for general data loading/normalization.

Each page already renders the header and footer for itself (but in jadelet land above the react page). We could move the layout rendering within the editor component for the page.

GregWeil:
Ok, rendering the header/footer within the page is no good, because that means it goes away until the page finishes loading.

GregWeil:
Ok everything is now fully functional.

I used react contexts to send around application.currentUser() in the user page. Anything using the consumer/helper functions will noop if the provider isn't in the page. Future work is to move user store/load there along with the api, but I don't think that makes sense as long as the application.currentUser model is around.

I'm not that happy with how entityeditor vs userpageeditor worked out. I think I'll just flatten it out and copy the behavior entityeditor provided to the different page specific editors. It makes it awkward to track/edit state that isn't in the model being edited specifically, for ex deleting/undeleting projects. That'll probably happen after this though, I want to get it merged before I start doing more refactors if possible.

The deletedprojects list also needs some work, the list of deleted projects is split between the component and the user editor, and there's some wonky logic to build a single list out of everything. I'm going to poke at it a bit.

GregWeil:
Reworked the deleted projects list, added some error handling to the seo rendering (5s timeout after which it'll just act like the user doesn't exist)

pketh:
oh wait now it works , looking now

src/presenters/entity-page-projects.jsx
```diff
@@ -1,125 +1,116 @@
...
-      this.setStateFromModels(projectsModel, pinsModel, this);
-    });
+    this.ensureProjects(this.props.projects);
   }
   
-  componentWillUnmount(){
-    this.aggregateObservable && this.aggregateObservable.releaseDependencies();
-    this.aggregateObservable = null;
+  componentDidUpdate() {
+    this.ensureProjects(this.props.projects);
```
judeallred: I'm not totally sure, but I think getDerivedStateFromProps() is the proper thing to use here, as opposed to componentDidMount plus componenteDidUpdate

https://reactjs.org/docs/react-component.html#static-getderivedstatefromprops

judeallred: oh, it looks like componentDidUpdate() is preferred when it's possible to use,  so I think you're good. :ignoreme:

judeallred: I'm concerned about the infinite loop warnings from https://reactjs.org/docs/react-component.html#componentdidupdate ,  but it seems like it's working ok atm. shrug?

src/presenters/includes/uploader.jsx
```diff
@@ -0,0 +1,94 @@
...
+    } catch (e) {
+      this.setState({error: true});
+      throw e;
+    } finally {
+      this.setState({uploading: false});
+    }
+    return url;
+  }
+  
+  async uploadAssetSizes(blob, policy, sizes) {
```
judeallred: we gtg to use async/await instead of promises going forward?

GregWeil: I really want to be... looks like IE doesn't actually support it though. we could use a babel transform/polyfill to add support, or I can go through and switch them back to standard .then logic.

src/presenters/entity-page-projects.jsx
GregWeil: yeah, I guess I should throw a comment on ensureProjects about being careful. Since it always sets everything in one go (and only if necessary) it shouldn't call itself more than once or twice.

src/presenters/pages/project.jsx
```diff
@@ -140,50 +141,35 @@ ProjectPage.propTypes = {
...
+  updateFields: PropTypes.func.isRequired,
 };
 
-const ProjectPageLoader = ({name, get, ...props}) => (
+const ProjectPageLoader = ({name, get, api, ...props}) => (
   <DataLoader get={get} renderError={() => <NotFound name={name}/>}>
-    {project => project ? <ProjectPageEditor initialProject={project} {...props}/> : <NotFound name={name}/>}
+    {project => project ? (
+      <EntityEditor api={api} initial={project} type="projects">
+        {({entity, ...funcs}) => <ProjectPageEditor project={entity} {...funcs} {...props}/>}
```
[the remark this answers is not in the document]
GregWeil: The children as a function thing is called render props https://reactjs.org/docs/render-props.html
I'm actually refactoring out the entity editor right now, it made it really difficult to edit adjacent state like undeleting projects. I do like the idea of moving the dependencies into a 'composed' editor though, so you don't have to worry about getting all the stuff when you just want the editor.

PR #62, opened 2018-07-18
Superb guide: teams and user share @urls

GregWeil:
the teams cache has been removed, in favor of `src/curated/teams.js` which just stores a map of url -> id for all existing teams (we can always add new entries).

`@urls` will prefer to load teams, and if the team isn't found they will load the user with that name. The team api is case sensitive, so for the moment `/@glitch` will work but `/@Glitch` will say nothing found.

If you are looking at a team that has the same (ignoring case) url as your login, a persistent notification appears with a link to your /user/id url. You can test that out by renaming yourself to `Test-Team`, due to a current api bug.

also added seo stuff for `@team` urls, and fixed that rare rendering issue with team pages. I think the issue is that the new project dialog loads projects and sometimes its id only models take precedence over what we get from the team api. You still get a prop types warning but the page resolves itself


GregWeil:
The stored map is only used for backwards compatibility on root level /team urls, but if we want to add something to it later we can always just poke in and change it in the community site.

src/presenters/includes/name-conflict.jsx
```diff
@@ -0,0 +1,48 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import {CurrentUserConsumer} from '../current-user.jsx';
+import Notifications from '../notifications.jsx';
+
+const NameConflictWarning = ({id}) => (
+  <React.Fragment>
+    <p>This team has your name!</p>
+    <a className="button button-small button-tertiary button-in-notification-container" href={`/user/${id}`}>Go to your profile</a>
```
judeallred: nice handling

pketh: mild copy suggestions:
- we shouldn't use `!` in most copy. Depending on the person, it can come across as excitement or anger. Neither is a good fit for an error message. 
- the profile button can just be 'Your Profile' to match the same button in the user info pop

but, how would this scenario happen tho? 
- In the case of the current teams, because there's like 125 we can manually make sure there are no existing conflicts. 
- in the case of future teams? renaming/patching a team or user url/domain should check to make sure that domain is  not in use by another team or user (we can request support for this from the api)

src/presenters/pages/team.jsx
```diff
@@ -75,38 +68,25 @@ TeamPage.propTypes = {
...
-      ) : <NotFound name={name}/>}
-    </DataLoader>
-  </Notifications>
-);
-TeamPageLoader.propTypes = {
-  get: PropTypes.func.isRequired,
-  name: PropTypes.node.isRequired,
+const teamConflictsWithUser = (team, currentUserModel) => {
+  if (currentUserModel.login()) {
+    return currentUserModel.login().toLowerCase() === team.url;
```
pketh: re comment above, do conflicts currently exist? If not, I think the better solution is to prevent those conflicts from happening in the first place (at the name creation/change step)

src/presenters/pages/team.jsx
```diff
@@ -75,38 +68,25 @@ TeamPage.propTypes = {
...
-    myProjects: application.currentUser().projects().map(({asProps}) => asProps()),
-    get: () => application.api().get(`teams/${id}`).then(({data}) => (data ? TeamModel(data).update(data).asProps() : null)),
-    searchUsers: (query) => UserModel.getSearchResultsJSON(application, query).then(users => users.map(user => UserModel(user).asProps())),
-    getProjects: (ids) => application.api().get(`projects/byIds?ids=${ids.join(',')}`).then(({data}) => data.map(d => ProjectModel(d).update(d).asProps())),
-  };
-  const content = Reactlet(TeamPageLoader, props, 'teampage');
-  return LayoutPresenter(application, content);
-}
\ No newline at end of file
+const TeamPageContainer = ({api, currentUserModel, team, ...props}) => (
```
pketh: sweet – i think `container` is way more clear than the previous `loader` convention

src/presenters/pages/user.jsx
```diff
@@ -1,24 +1,15 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
-import LayoutPresenter from '../layout';
-import Reactlet from "../reactlet";
-
-import ProjectModel from '../../models/project';
 import {getAvatarStyle, getProfileStyle} from '../../models/user';
 
-import {DataLoader} from '../includes/loader.jsx';
```
pketh: this might be confusing with the existing loader at `src/presenters/includes/loader.jsx` (that draws these guys: 
screen shot 2018-07-19 at 1 43 46 pm )

pketh: would it make sense to make the file name more explicit, to `data-loader` as well?

GregWeil: it might be worth kicking that to a separate file. dataloader basically renders the circles while an async function is running, then renders children or an error once it finishes

src/presenters/includes/name-conflict.jsx
GregWeil: the api checks for conflicts already, but there is a race condition if a team and user try to set the same name at the same time. (user vs user and team vs team are enforced by the db, but not user vs team). the answer i heard from everyone on the api side was that preventing it adds a lot of complexity for a rare case, so throwing a warning like this is probably enough of a safety net in case it does happen. that said allowing team name and user NAME to coexist like it does now is a bug that I'm waiting on before releasing this.

src/presenters/pages/user.jsx
pketh: not sure i understand what you mean by `kicking that to a separate file` : `loader.jsx` is already a separate file

src/presenters/includes/name-conflict.jsx
```diff
@@ -0,0 +1,48 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import {CurrentUserConsumer} from '../current-user.jsx';
+import Notifications from '../notifications.jsx';
+
+const NameConflictWarning = ({id}) => (
+  <React.Fragment>
+    <p>This team has your name</p>
```
pketh: is there an action we want you to do with this information? Based on the button, it sounds like we're asking you to go to your own profile and change your own name?
If that's the case, adding some clarifying text like 
> This team has your name. You should update your info to remain unique ❄️

PR #64, opened 2018-07-19
Nova chance: react category page, remove category model

GregWeil:
also moved project list loading from `EntityPageProjects` to a separate component so categories can use it too


src/presenters/random-categories.jsx
```diff
@@ -46,34 +47,34 @@ export default class CategoryContainer extends React.Component {
...
-        category.projects = sampleSize(category.projects, 3);
-        return category;
-      });
-      this.setState({categories});
+  async loadCategories() {
+    // The API gives us a json blob with all of the categories, but only
+    // the 'projects' field on 3 of them.  If the field is present,
+    // then it's an array of projects.
+    const {data} = await this.props.api.get('categories/random?numCategories=3&projectsPerCategory=3');
+    const categoriesWithProjects = data.filter(category => !!category.projects);
```
judeallred: @GregWeil could you please check with @garethx on if we could take this a step further?
Specifically I'm thinking that we could move all the categories data into a file in /curated/, and that would let us deprecate and remove the db/api stuff while perhaps also being a better experience for both us and for Marketing

PR #71, opened 2018-07-25
Thundering way -- build stylus with webpack, case 3317684

judeallred:


judeallred:
this bumps our stylus build time up to like ~7-12 seconds...  which is a bummer.  i'll play some more and try to smooth that out.

judeallred:
agh ok, so the build time with the watcher was a <2 seconds,  which was tremendously better than without.   I'll work on keeping the watcher but making the output more robust....

judeallred:
I have successfully made things more complicated without necessarily fixing anything.  Hurrah? hmm.  I'll sit on this for a little bit.

judeallred:
A solution that would definitely work would be:  in production, build without watch and use the mv command as needed to deploy the 'asset';   in dev, --watch.

I've been avoiding starting down this precedent (because maintaining _two_ build scripts is a bummer), but it may be needed here.

judeallred:
okay, this is showing promise.

I've got the webpack build working for css.  In webpack 4, this came along: 
  https://github.com/webpack-contrib/mini-css-extract-plugin
...and deprecated all the old ways to do css.

I've got it running stylus and have the stylus autoprefixer.

it's working well; it's speedy, webpack watch is powering it, etc., etc.,

missing stuff: 
->  css minification/compression isn't working.  i don't know why.
-> source maps are weird;  we successfully get a css source map, but it just points to a single .styl file instead of to the whole collection of style files;  it's like the maps are being generated by the CSS loader but not by the stylus loader,  whereas we need it to listen to the stylus loader's output and go from there.

judeallred:
ok got it,
 -- stylus has its own compression lib, and i'm able to invoke that as before.
 -- i had to specifically turn on source maps for the css-loader module to get the stylus maps to come through.  Don't know why that was the key point, but hey presto

GregWeil:
Having to exclude the style.js manually seems odd, but I guess that's just a quirk of how webpack works. either way having one less watcher bouncing around seems good.

judeallred:
@GregWeil okay, this is ready for review now, lmk what you think

highlights:
 -> /public/styles has moved to /styles/
 -> we now build stylus via webpack, outputting styles.css as before
 -> we also generate styles.json via the manifest plugin, so we use that to far future our css files

judeallred:
@pketh WB!  fyi ^^

judeallred:
@GregWeil  yup, the css .js file is odd!  There's no getting around generating it though :-(

https://github.com/webpack-contrib/extract-text-webpack-plugin/issues/518
https://medium.com/webpack/the-new-css-workflow-step-1-79583bd107d7

Looks like WebPack vNext is targetting including CSS natively, which will do this better.

judeallred:
@GregWeil far futuring based on the manifest contents is a good idea! I'll note that down.

PR #72, opened 2018-07-25
Nova chance: react questions

GregWeil:
also removed the question model


src/presenters/question-item.jsx
```diff
@@ -0,0 +1,52 @@
...
+
+import {getEditorUrl} from '../models/project';
+
+const iconHelp = 'https://cdn.glitch.com/f7224274-1330-4022-a8f2-8ae09dbd68a8%2Fask-for-help.svg?1494954687906';
+
+function truncateQuestion(question) {
+  const max = 140;
+  if (question.length > max) {
+    return question.substring(0, max - 3) + '…';
+  }
```
judeallred: the ellipsis is only one character, so i think this should be max -1

GregWeil: the old one was max - 5 + three dots :man_shrugging: but yeah that probably makes more sense

GregWeil:
fixed new project pop breaking project models elsewhere, only max-1 truncation

judeallred:
lgtm

PR #73, opened 2018-07-25
Useful sting --s/updateProjectdomain/updateProjectDomain/ case 3321570

judeallred:
- also update a typo'd variable name which was a red herring


judeallred:
undid my shim; kept the rename.

PR #74, opened 2018-07-26
Nova chance: reactify the index page

GregWeil:
- removed some dead code form application
- moved some small jsx files into index.jsx

the index page has some `<script>` tags that it adds. react wasn't running them so I made a function that adds them to the dom. this will need revisiting when we go single-page-app, but it works for now


src/presenters/pages/index.jsx
```diff
@@ -0,0 +1,129 @@
...
+        <h2>How It Works</h2>
+        <span>
+          <a href="https://glitch.com/about">
+            <figure title="How Glitch works">
+              <img className="wide" src={whatsGlitchWide} alt={whatsGlitchAlt}/>
+              <img className="narrow" src={whatsGlitchNarrow} alt={whatsGlitchAlt}/>
+            </figure>
+          </a>
+          <div>
+            And it's <img className="free" src={free} alt="free"/>.{' '}
```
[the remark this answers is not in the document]
GregWeil: Yup

src/presenters/pages/index.jsx
```diff
@@ -0,0 +1,129 @@
...
+      We're <a href="https://www.fogcreek.com">a friendly, self-funded company</a> that's
+      been helping people make stuff for over {age} years.
+    </p>
+  </section>
+);
+
+const MadeInGlitch = () => (
+  <section className="made-in-glitch">
+    <p>Of course, this site was made on Glitch too</p>
+    <a href={getEditorUrl('community')} className="button button-link has-emoji">
```
judeallred: Hmm...  this style of using the utility functions is going to make unit testing a lot harder.  I'd prefer more of a DI model, and I'm not sure on the best way to go about that here.

Maybe we should hurry on getting a unit test skeleton in quickly so that we can actively use DI for testing rather than as an academic ideal.

pketh: what's DI?

src/presenters/recent-projects.jsx
```diff
@@ -0,0 +1,61 @@
...
+    {(user, fetched) => (
+      <RecentProjectsContainer user={user}>
+        {fetched ? (
+          <ProjectsLoader api={api} projects={user.projects.slice(0,3)}>
+            {projects => <ProjectsUL projects={projects}/>}
+          </ProjectsLoader>
+        ) : <Loader/>}
+      </RecentProjectsContainer>
+    )}
+  </CurrentUserConsumer>
```
judeallred: I love how this is coming together.  Used to be a huge pain to pull this off, and now it's clear, concise, and waaaay less bug prone

src/presenters/pages/index.jsx
```diff
@@ -0,0 +1,129 @@
...
+import Questions from '../questions.jsx';
+import RandomCategories from '../random-categories.jsx';
+import RecentProjects from '../recent-projects.jsx';
+
+const play = "https://cdn.hyperdev.com/6ce807b5-7214-49d7-aadd-f11803bc35fd%2Fplay.svg";
+const whatsGlitchWide = "https://cdn.glitch.com/f7224274-1330-4022-a8f2-8ae09dbd68a8%2Fwhats-glitch-wide.svg?1499885209761";
+const whatsGlitchNarrow = "https://cdn.glitch.com/f7224274-1330-4022-a8f2-8ae09dbd68a8%2Fwhats-glitch-narrow.svg?1499884900667";
+const free = "https://cdn.glitch.com/2bdfb3f8-05ef-4035-a06e-2043962a3a13%2Ffree.svg?1499350845981";
+
+const whatsGlitchAlt = "Create a node app, or remix one. It updates as you type. Code with Friends!";
```
pketh: what's the advantage of having `whatsGlitchAlt` and `ByFogCreek` inline here, vs separate files like before?

src/presenters/pages/index.jsx
GregWeil: I think the container pattern is the shortest path to dependency injection. The current component would have an extra `getEditorUrl` prop that gets preset by the container using getEditorUrl. (maybe the project model exports should actually be a component that calls children with all of the getX functions?) It seems a bit verbose when we're just going to call the function and send the results downward.

Another option would be to make an `<EditorLink>` component that creates the `<a href>` and passes through the other props. On the other hand we might want to create a `<ButtonLink>` that does the classes. Not sure which is better.

src/presenters/pages/index.jsx
judeallred: my 2c: Each file is essentially a "global" in the sense that it's a top level thing that anyone can import.  When something is small and only used in one place, like these, I think it's nice to keep it in place since it helps avoid adding to the mental weight of the whole project.  It might make this file a little harder to read (although these are small enough that I don't think that's the case), but it helps make the whole project easier to read.

src/presenters/pages/index.jsx
```diff
@@ -0,0 +1,129 @@
...
+const age = current - FOUNDED;
+const logo = "https://cdn.glitch.com/2bdfb3f8-05ef-4035-a06e-2043962a3a13%2Ffogcreek.svg";
+
+const ByFogCreek = () => (
+  <section className="by-fogcreek" role="complementary">
+    <h2>Made By Fog Creek</h2>
+    <img src={logo} alt="Fog Creek logo"/>
+    <p>
+      You might know us for making Trello, FogBugz, and co-creating Stack Overflow. 
+      We're <a href="https://www.fogcreek.com">a friendly, self-funded company</a> that's
```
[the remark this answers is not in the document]
GregWeil: I'm going to take this as an excuse to convert it to a stateful component that auto refreshes the age

src/presenters/pages/index.jsx
judeallred: DI == dependency injection

Greg and I chatted a bit at lunch and followed up on this;  I don't think we came to a strong conclusion other than 'yeah adding in tests is a good thing to do'.   There are ways we can write these utility functions more verbosely in order to get to more of a DI model (or just always use react components),  but that's a lot of boilerplate that doesn't seem to buy us all that much.  We were thinking instead that so long as we keep the utility functions pure, then they can have their own unit tests fairly simply.

PR #75, opened 2018-07-30
adds teams admin assignment, views

pketh:
https://community-pirijan.glitch.me/

includes a bunch of new ui for
- adding, removing admins
- seeing who's an admin
- shows the free team limits (based on a feature stub)
- allows users to create teams (hidden)
- allows admins to delete teams (hidden)
- some stub work for adding templates to teams (incomplete and hidden)

these are the ui state expectations I QAed against:
artboard


pketh:
there's prob lots of gross hacks in this code, i'm probably not using async/await where I could for example. still a wip but should be overall reviewable now

server/proxy.js
```diff
@@ -8,32 +8,27 @@ const url = require('url');
...
+  // Proxy all the requests to /help/ over to help-center:
+  app.use('/help/', proxy('help-center.glitch.me', {
     preserveHostHdr: false, // glitch routes based on this, so we have to reset it
     https: false, // allows the proxy to do less work
     proxyReqPathResolver: function(req) {
-      const path = routeWithLeadingSlash + url.parse(req.url).path;
+      const path = '/help' + url.parse(req.url).path;
       console.log("Proxied:", path);
       return path;
     }
```
judeallred: it seems like the merge is going the wrong way here;   is this branch up to date with master?

judeallred:
visiting http://community-pirijan.glitch.me/fogcreek while not logged in gives:
```
raven.js:58 Warning: Failed prop type: The prop `currentUserId` is marked as required in `TeamPage`, but its value is `undefined`.
    in TeamPage (created by TeamEditor)
    in TeamEditor
    in Uploader
    in ErrorHandler (created by TeamEditorContainer)
    in TeamEditorContainer (created by DataLoader)
    in DataLoader (created by TeamPageLoader)
    in Notifications (created by TeamPageLoader)
    in TeamPageLoader
```

judeallred:
UI feedback:  When the progress bar is full, (to me) it doesn't look like a progress bar anymore, just a thick line.  Maybe if it had a different color or some sort of "i'm full" convention?  I'm not sure how to fix it, just sharing my impression :-)
image

judeallred:
image

I'm thinking we should remove the user profile link from this section.  It's kind of like the shopping cart billing idea-- once you click 'check out!', you don't want to then have buttons that lead you away from the checkout funnel.   In this case, if someone clicked on one of those pictures, they'd leave the page and clicking 'back' wouldn't get them back to the upsell part because it was a popover.  

So I'd suggest either removing the links from these images, or adding target="_blank" so that they open in new tabs and don't take the user away from the upgrade popover

judeallred:
image
The admin oval doesn't feel very crisp to me... I think perhaps because it's close tone the skin tone of some of the people pictured, therefore its edges tend to blur a little bit?

judeallred:
bug: page reload is needed before UI expresses changes in admin status

steps:
 1. visit http://community-pirijan.glitch.me/fogcreek
 2. click on a user
 3. click either 'make an admin' or 'remove admin status' (depending on whichever one they currently are
 4. click outside the popover to close it

expected:
  user display in the list has now been toggled between admin status

actual:
  user display is as before, admin status is unchanged.  If you click the user again, the popover indicates their old status.

5. refresh the page
6. changes are now in effect

judeallred:
@pketh what's the secret incantation to hide the hidden add/delete team ui?

server/proxy.js
GregWeil: hmm, looks like it is up to date per git, but most of the files are outdated and most of the diffs are reverting stuff we want to keep

pketh:
@judeallred it's just commented out in the code itself

server/proxy.js
pketh: are the overall files changed losing code? I've made individual commits where i reverted merge attempts so commit view is misleading

GregWeil: yup (for ex the file this comment is on lost /featured routing)

pketh:
re https://github.com/FogCreek/Glitch-Community/pull/75#issuecomment-408965456 if i abandon the purity of `<progress>` I can do something inline like 
screen shot 2018-07-30 at 4 44 59 pm

pketh:
made some more changes to the banner, 
- shows total projects instead of # left, for easier layout and to not discourage ppl from adding that last project
screen shot 2018-07-31 at 10 29 23 am

- bar changes color at limit(5)
screen shot 2018-07-31 at 10 26 30 am

pketh:
users in `upgrade-team-pop` are no longer links

pketh:
added a border around the admin bubbles so that they'll always have contrast
screen shot 2018-07-31 at 10 44 40 am

pketh:
fixed admin status propagation issues

pketh:
can you try breaking the ui some more @judeallred whenever?

judeallred:
@pketh will do! :-D

pketh:
added a blank slate for when a team has no projects yet
screen shot 2018-08-01 at 11 52 48 am

pketh:
screen shot 2018-08-02 at 9 27 45 am
Most of these conflicts was just a weird merge and not actually conflicts. In almost every case I picked up changes `HEAD`/`master`. In some cases I hadn't made changes to a file, but the conflicts were so mangled and stuff seemed lost that I just copied the raw version off master. the only real exception being `teams.jsx`

pketh:
oh jeez there's like entire new files missing still :@

pketh:
i hosed this branch with the power of git, moving to a new one

PR #79, opened 2018-08-02
Team admin

pketh:
https://community-pirijan.glitch.me/

includes a bunch of new ui for
- adding, removing admins
- seeing who's an admin
- shows the free team limits (based on a feature stub)
- allows users to create teams (hidden)
- allows admins to delete teams (hidden)
- some stub work for adding templates to teams (incomplete and hidden)

these are the ui state expectations I QAed against:
artboard


src/models/team.js
```diff
@@ -96,13 +98,26 @@ export default Team = function(I, self) {
       }
       return {backgroundColor: self.backgroundColor()};
     },
-    
+
+
     asProps() {
+      // const adminIds = (users) => {
```
judeallred: some scratch code to clean up

pketh: removed in upcoming commit

pketh:
^ - no id users can view page
- cover clear button returns

pketh:
^ removing/adding projects updates analytics

pketh:
^ 
screen shot 2018-08-03 at 1 25 12 pm

pketh:
ok this is feature complete now. now i'm just QAing till I hear more

GregWeil:
small bug:
add a user, make them an admin, remove them, add them back
they'll appear as an admin, but on refresh revert back to a normal member

pketh:
added blank slate for when you have no projects to add 
screen shot 2018-08-03 at 2 24 03 pm

src/presenters/includes/editable-field.jsx
```diff
@@ -9,15 +9,41 @@ export default class EditableField extends React.Component {
...
-      this.handleSuccess.bind(this),
-      this.handleFailure.bind(this, value));
+    if (this.props.fieldOnlyUpdatesOnSubmit) {
+      this.props.update(value);
+    }
+    else {
+      this.props.update(value).then(
+        this.handleSuccess.bind(this),
+        this.handleFailure.bind(this, value));
+    }
```
GregWeil: EditableField is kind of hard to follow now that it has conditional behavior. If I'm reading it correctly, it seems like the changes are to let you use it without the debounced update and be able to provide your own error messages.

If that's the case it could probably exist as a stateless render component that just takes value/update/error/placeholder that would also be called by EditableField's render function. (So instead of EditableField having two modes, there'd just be a stateless fieldwitherror and the EditableField, which debounces update and tracks errors to render a fieldwitherror)

pketh:
fixed https://github.com/FogCreek/Glitch-Community/pull/79#issuecomment-410336136

src/presenters/includes/editable-field.jsx
pketh: this is used by the unused `create-team-pop` so i'm going to remove the condition. But also I want to change the behaiviour of the component to not need a `fieldOnlyUpdatesOnSubmit` in the first place

pketh: removed conditions in https://github.com/FogCreek/Glitch-Community/pull/79/commits/047f1c4f749cce0122f410bcbe4946907cb11c2e

src/presenters/team-editor.jsx
```diff
@@ -1,3 +1,5 @@
+/* global notify */
```
GregWeil: `notify` doesn't exist at the window level. You have to add a `<Notifications>` down in the `TeamEditorContainer` to give the `createErrorNotification` func as a prop to the editor. (error-handlers.jsx does the same thing, so maybe it should just send createErrorNotification along so the team editor can use it too?)

pketh: like this ? ```const TeamEditorContainer = ({api, children, currentUserModel, initialTeam}) => (
  <ErrorHandlers>
    {errorFuncs => (
      <Uploader>
        {uploadFuncs => (
          <Notifications>
            <TeamEditor {...{api, currentUserModel, initialTeam}} {...uploadFuncs} {...errorFuncs}>
              {children}
            </TeamEditor>
          </Notifications>
        )}
      </Uploader>
    )}
  </ErrorHandlers>
);
```

Comment (pketh) [reply src/presenters/team-editor.jsx]: fixed in https://github.com/FogCreek/Glitch-Community/pull/79/commits/f826920551a0592f063486346e67aec275f5aef1 , hope i did that right - seems to work as expected tho

**Comment by GregWeil** [comment]
It looks like layout.js came back form the dead, but everything else looks properly merged now.
Only code comment I have is that the team page shouldn't use a class if it has no state. Other than that it looks good.

File: src/presenters/pages/team.jsx [hunk 4]
```diff
@@ -2,75 +2,230 @@ import React from 'react';
...
+        }
+
+        {/* billing info section goes here */}
+
+        {/* Temporary: enable once team creation is public
+        { currentUserIsTeamAdmin &&
+          <DeleteTeam api={() => this.props.api}
+            teamId={this.props.team.id}
+            teamName={this.props.team.name}
+            teamAdmins={this.teamAdmins}
```
GregWeil: This should send in the list, not a function to get the list. That way react will update this when the set of users changes, and the deleteteam component won't have to recompute the set of admins when its own state changes.

pketh:
@GregWeil addressed the two issues above (didn't have time to un-class the team page, let me know if that's a blocker for you, but I figure it's a noop thing that can happen later if necessary)

src/presenters/pop-overs/delete-team-pop.jsx
```diff
@@ -0,0 +1,78 @@
...
+        </section>
+        <section className="pop-over-actions danger-zone">
+          <button className="button button-small has-emoji opens-pop-over" onClick={this.deleteTeam}>
+            <span>Delete {this.props.teamName} </span> 
+            <span className="emoji bomb" role="img" aria-label="bomb emoji"></span>
+            { this.state.teamIsDeleting && <Loader /> }
+          </button>
+        </section>
+        <section className="pop-over-info">
+          <UsersList users={this.props.admins()}/>
```
GregWeil: `DeleteTeamPop` is still expecting a func rather than an array. and it's also expecting the prop as `admins` rather than `teamAdmins` (same for the `DeleteTeam` component).

pketh:
as discussed irl, i'll fix up deleteteams stuff after it's uncommented/exposed-to-users for easier debugging. gonna merge this soon

PR #81, opened 2018-08-07
steep-kettle: Join and leave team projects

pketh:
https://steep-kettle.glitch.me/

team page: lets team members join/leave projects in their team

user page: `leave project` option only shows up if you have other project members


src/presenters/pop-overs/project-options-pop.jsx
```diff
@@ -9,85 +10,140 @@ const PopoverButton = ({onClick, text, emoji}) => (
...
+  }
+  
+  function currentUserIsOnProject() {
+    let projectUsers = props.project.users.map(user => {
+      return user.id;
+    });
+    if (projectUsers.includes(props.currentUser.id)) {
+      return true;
     }
   }
```
GregWeil: This doesn't have to be a function since the result is being used right here. You can define a variable and reference that. Also it looks like you have a leftover proptype for `currentUserIsOnProject` defined.

GregWeil: It looks like the caller to this also defines a function and passes in `currentUserIsOnProject`, but it isn't used since it gets recomputed here

CONTRIBUTING.md
```diff
@@ -95,31 +93,43 @@ After you either wait or run the above commands, go to step 5 to update the pull
...
-Ok, now the GitHub repository is updated and stable. 
-  
-3. Inside of https://glitch.com/~community-staging, open up the console.
+git checkout master
+git pull
+git checkout my-branch
+git merge master
+git push origin my-branch
+```
+3. Go through code/design review, to get your PR merged into master on [`Glitch-Community`](https://github.com/FogCreek/Glitch-Community)
```
GregWeil: Going through code/design review should be a precondition for these steps, otherwise you'll have to pull live into the  github again and basically start over.

judeallred: right; merging in from Live a deploy step that helps reconcile gareth's changes with github,  it's not something that should come up while we're iterating on PR's.

CONTRIBUTING.md
```diff
@@ -1,26 +1,24 @@
...
 ------------
 
-Hey, welcome!  This document should give you all the steps you need to make contributions to the Glitch community site. This guide applies to full time makers of Glitch as well as all members of our Glitch community.
+Hi, this document should give you all the steps you need to make contributions to the Glitch community site. This guide applies to full time makers of Glitch as well as all members of our Glitch community.
 
-Please note we have a [code of conduct](https://glitch.com/edit/#!/community?path=CODE_OF_CONDUCT.md), please follow it in all your interactions with the project and the denizens thereof.
+👀 [Code of Conduct](https://glitch.com/edit/#!/community?path=CODE_OF_CONDUCT.md)
 
-We also having a [style guide](https://glitch.com/edit/#!/community?path=STYLE.md) discussing our coding conventions, styles, and goals.
+👀 [Code Style Guide](https://glitch.com/edit/#!/community?path=STYLE.md)
```
judeallred: I don't love this shortening -- this doc is written for our community users first, and ourselves second.   I think that doing things like asking people to follow a code of conduct is imporant

CONTRIBUTING.md
```diff
@@ -95,31 +93,43 @@ After you either wait or run the above commands, go to step 5 to update the pull
...
 #### Deployment
 
-Only employees of Fog Creek will be able to do this step, and here it is!  This generally happens immediately after we merge in any pull request.
+Only employees of Fog Creek will be able to do this step. 
+
+1. Locally, get [~community-staging](https://glitch.com/~community-staging) on Github up to date with prod
+```
+  git checkout master
+  git pull https://api.glitch.com/community/git
+  git push origin master
```
judeallred:
```
1. Locally, get [~community-staging](https://glitch.com/~community-staging) on Github up to date with prod
```
  git checkout master
  git pull https://api.glitch.com/community/git
  git push origin master
```
that's not what this does;  this code says "update github from the live site", and the instruction says "update staging on github" (which doesn't exist,  staging is a glitch project)

what problem are trying to solve with this document?

**Comment by judeallred** [comment]
I'm pretty uncomfortable with the proposed changes to contributing.md here -- it goes against the flow I've been using successfully and stably, and introduces a few thing that don't make sense to me. How about if we take a look at the original document together and chat about what was bothering you about it, then go after fixing it together?

File: src/presenters/entity-page-projects.jsx [hunk 5]
```diff
@@ -32,6 +32,13 @@ const EntityPageProjects = ({projects, pins, isAuthorized, addPin, removePin, pr
     </React.Fragment>
   );
   
+  projectOptions = _.mapValues(projectOptions, function(projectOption) {
+    return async (projectId, userId) => {
+      await projectOption(projectId, userId);
+      reloadProject(projectId);
+    }
+  });
+
```
judeallred: you can use Object.entires for this sort of thing -- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/entries 

I think it'd turn into something like:
```
projectOptions = Object.entries(projectOptions).map(
 ([name, func]) => {
   const wrapped = async (projectId, userId) => {
      await func(projectId, userId);
      reloadProject(projectId);
  }
  return {[name]: wrapped};
```

i think both are ok, mostly just an FYI.  I try to avoid using libraries when there is good native behavior to replace them;  this one is borderline for me because it is handy not to muck with they keys in your version :-)

src/presenters/entity-page-projects.jsx
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import {partition} from 'lodash';
+import _ from 'lodash';
```
judeallred: why switch to _ instead of to {partition, mapValues}?   We mostly use the {} convention because it's a way to declaring what parts of lodash you intend to use rather than the whole thing

pketh: because the build tools now do the work of using only the parts of a lib needed by the project, it seems more straightforward to just use it straight (less to extend if we need another lodash thing in here in the future, code examples more closely match their docs)

src/presenters/pop-overs/project-options-pop.jsx
pketh: removed func in upcoming commit

src/presenters/pop-overs/project-options-pop.jsx
```diff
@@ -9,85 +10,140 @@ const PopoverButton = ({onClick, text, emoji}) => (
...
+  }
+  
+
   return (
     <dialog className="pop-over project-options-pop">
       <section className="pop-over-actions">
-        {!!addPin && <PopoverButton onClick={clickAddPin} text="Pin This" emoji="pushpin"/>}
-        {!!removePin && <PopoverButton onClick={clickRemovePin} text="Un-Pin This" emoji="pushpin"/>}
+        {!!props.addPin && <PopoverButton onClick={addPin} text="Pin " emoji="pushpin"/>}
+        {!!props.removePin && <PopoverButton onClick={removePin} text="Un-Pin " emoji="pushpin"/>}
```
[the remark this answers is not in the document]
pketh: they both reference different things (one a prop, one a func) , but i'll name them distinctly in an upcoming commit

pketh:
@judeallred reverted contributors and addressed issues
@GregWeil added commits for reported issues
if you could give this a quick pass that'd be great, thx

GregWeil:
Seems good to me

PR #83, opened 2018-08-13
Hazel liquid -- updates to contributing.md

judeallred:
Trying to address some good concerns that Pirijan had,  in particular:

- Lack of clarity on which steps are one-time versus during-PR's versus during-deployments
- When there is a sequence of console commands,  not having them in easily copy/pasteable form.

To this end, I've broken apart the sections a bit more, so you should be more able to go just to the section you need (e.g. "update my PR...") , and within sections I've batched the console commands together

Along the way, I've also encoded our undocumented convention of keeping a 'live' remote pointed at ~community,  and of naming the branches the same as the remixes. 

Thoughts/feels/suggestions?

Pirijan, is this better for you? are there specific sections that you'd like to see more work on?


judeallred:
Rather than viewing the diff and rendering markdown in your imagination, i suggest reading the rendered content here: https://glitch.com/edit/#!/hazel-liquid?path=CONTRIBUTING.md:48:21

CONTRIBUTING.md
```diff
@@ -7,19 +7,15 @@ Please note we have a [code of conduct](https://glitch.com/edit/#!/community?pat
 
 We also having a [style guide](https://glitch.com/edit/#!/community?path=STYLE.md) discussing our coding conventions, styles, and goals.
 
-Wish List
---------
-Check out our [Wish List](https://glitch.com/edit/#!/community?path=WISH_LIST.md) for ideas of things to get started on.
```
judeallred: This was created for the purpose of Manuscript staff having some suggestions on things to jump into, without needing to talk to the team.   I've removed it, as we no longer have that need and it's starting to get stale.  It did get a couple of things worked on though :-)

pketh:
the distinction between the 'on your side' vs 'on our side' seems unclear to me. ie stuff like 
> All interactions between this site and Glitch-the-application are using public API endpoints
seems like it'd also be useful to outside contributors ('on your side')

pketh:
i like the comments next to the code blocks, much more direct 👍

CONTRIBUTING.md
```diff
@@ -43,86 +37,139 @@ Accessible technologies are important to allowing people to contribute.  As we i
...
+
+Satisfied?  Deploy it.
+
+#### Deploy
+
+Run the _special command_ to swap ~community with ~community-staging.
+
+(there's a special endpoint for this, ask us about it and sit beside a team member the first time you use it.   It's the same thing as renaming the the two projects, but it does this atomically and without downtime, which you can't do if you rename then manually.)
+
+#### Revert the Deploy.
```
pketh: the way  'revert the deploy' section is phrased/sequenced it sounds like it's the thing you do after the deploy. 

It might be clearer to use `If You Need to Revert the Deploy` instead

pketh:
thanks for the updates 🥇

CONTRIBUTING.md
GregWeil: Agree with this, also it should mention checking for live site changes when you revert, so we can preserve featured item changes.

judeallred:
Thanks all!  I agree with the comments above, will do another pass

judeallred:
Updated to address comments;  I collapsed the /you/us/ section into one, and expanded the 'If you need to revert...' section.

pketh:
:+1:

CONTRIBUTING.md
```diff
@@ -7,122 +7,169 @@ Please note we have a [code of conduct](https://glitch.com/edit/#!/community?pat
...
-```git checkout origin/master``` 
-2. then pull from the base community site to get any live changes
 ```
-git pull https://api.glitch.com/community/git
+  # Make sure we're in the right place and up to date.
+  git checkout master
+  git pull
+
+  # Pull in any live changes that aren't yet in Master:
+  git pull live master
```
pketh: I tried to do this step and got ```fatal: 'live' does not appear to be a git repository
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```
the `git pull` above pulled from master as expected

Comment (judeallred) [reply CONTRIBUTING.md]: that'll happen if you don't do the first time setup steps;  do those and then it'll work

File: CONTRIBUTING.md [hunk 4]
```diff
@@ -7,122 +7,169 @@ Please note we have a [code of conduct](https://glitch.com/edit/#!/community?pat
...
-   ```   
    
-Note: Glitch apps make git commits (we call them checkpoints) every 10 minutes. If you make some quick changes, they may not show up in your diff on the Github pull request. If that is the case, you can either wait several minutes for the checkpoint to be made or manually commit your changes in your Glitch app's console (under "Advanced Options") by running:
-  ```
-  git add .
-  git commit -m "A short message about what your commit does"
-  ```
-After you either wait or run the above commands, go to step 5 to update the pull request.
+  # Update the PR in Github
+   git push origin my-remix
```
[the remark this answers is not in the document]
judeallred: PR merging into master happens in github w/ the completion of the code review.  I'll see if i can clarify that.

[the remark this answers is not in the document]
judeallred: I strengthened the wording.

Going to go ahead and deploy this version, but we'll keep iterating on the doc!  move further comments to #glitch-community :-)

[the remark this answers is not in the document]
pketh: :+1:

PR #84, opened 2018-08-14
Superb guide: invite users to teams, whitelist domains

GregWeil:
Users are not added directly, instead they will get an email with a join link. Note that the apis are not entirely there yet, so no emails actually get sent (aka there are blockers to releasing this)

- The add user popup will detect email addresses and show an 'invite {email}' option
- Emails and `@domain.com` will show an option to whitelist the domain
- You can remove yourself from a team without having to be an admin
- If you match the whitelist, a join team button appears on the page

Pending api support:
- Inviting users by id or email
- Having more than a single whitelisted domain for a team

Stuff I'm iffy about
- Whitelisted domains have a hover and a popover, which both say the same thing

Also I stuck `currentUser` and `api` on the window. Actual code shouldn't use them from there, but it's way more convenient for testing stuff.


judeallred:
image
The formatting on the 'allow anyone with...' line doesn't feel great to me.  I can't quite put my finger on it but something about the line height and text flow.  

related;  I'm wondering if we can fetch the favicon for a given domain?   I think we'd have to write a little proxy for it, but it'd work something like:
 1. request from favicon.glitch.me/?source={url}
 2. favicon.glitch.me CURL's url, finds <favicon> in html and extracts path
 3. returns path to icon _or_ proxy it directly

I'm worried about script injection into the img src="", but other than that I think this would be slick.  It'd enable this sort of thing:
image

judeallred:
typo: 
image

(also would be cool to use the .ico here as well...)

judeallred:
code lgtm

judeallred:
I haven't tested the full join flow (and need to),  but so far so good.   I'll add @ttseng here so she can either comment in place or meet with you and walk through it.

judeallred:
looking into IMG src --  OWASP claims that modern browsers are ok with remote img src uses now. 
 https://security.stackexchange.com/questions/135513/what-could-an-img-src-xss-do suggests that it's still terrifying, but perhaps not if we limit to .ico or otherwise scrub the source in some way.

src/presenters/current-user.jsx
```diff
@@ -90,6 +90,10 @@ class CurrentUserManager extends React.Component {
     )) {
       this.load();
     }
+    
+    // hooks for easier debugging
+    window.currentUser = currentUser;
+    window.api = this.api();
```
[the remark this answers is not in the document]
pketh: does making it a global window object obviate the need for the whole react context thing? (I'm also cool with keeping it, esp if window is just how we do global things in the future)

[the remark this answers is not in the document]
GregWeil: These should not be referenced off of window in real code, since that won't trigger any of react's update/rendering stuff. They're just on the window so you can get to them from the console without having to poke around the react debugger

pketh:
you can add line-height: 16px to .results-list .result-name in results-list.styl - it doesn't look like it should affect the single line result entries
screen shot 2018-08-15 at 11 23 13 am


if an email is in the input we should make sure the email entry is above any user matches.
screen shot 2018-08-15 at 11 23 01 am


@ttseng if you're looking at this, here's the spec/mockups https://www.notion.so/glitch/77466b6f4229453bbcf7722358c929c2?v=e91a07a4cb9a486885d7aa9ea04c0377&p=c5c7e224e44f4d939b1e35c8a446a7cc

positioning of the letter for non-favicon domains is off (maybe this is mac specific related to our weird font baseline) 
screen shot 2018-08-15 at 11 28 08 am
try 
```
.results-list .whitelisted-domain
height: 21px
padding-top: 1px
```

.whitelisted-domain letter should be medium(`font-weight: 600;`) instead of bold

ttseng:
Really like the favicon for whitelists - super effective.

A list of suggestions:

1. Should the Profile link from the team members tooltip be removed?  It seems redundant.
image

2. The whitelist feature seems a bit hidden, since you would need to type @domain.com to be able to use it.  Maybe we could add something to the bottom of the search field, when the add btn is clicked, that gives a tip:
image

(I don't really like the current wording I have in the mockup above - any suggestions?  I'm guessing we are trying to avoid the word "whitelist" as a user-facing term)

I also imagine that the whitelists tip automatically goes away once you type something in the search field.

3. The notifications about sending invites to added users is far away from the input field, especially when viewing the page on a large screen.  I missed the notification the first few times I tried adding someone:

image

What if we had the pop-up show up directly below the Add btn, such as

image

4. Copy suggestion for the Upgrade team btn:
"Paid teams are coming soon.  In the meantime, feel free to add as many projects as you want — they won’t go way."

image


(Currently, it's "feel free to add as many project as you want, they won't go away."  comma-usage is not grammatically correct)

pketh:
code lgtm :+1:

pketh:
@ttseng re:
1: the links are redundant but across both the editor and community site we've set an expectation that clicking the user avatar takes you to your profile page. In that context, I think it might feel awkward if this was the one place it didn't. The 'your profile' button is technically redundant, it may or may not be an affordance that's helpful to ppl (it was added based on feedback that ppl weren't sure how to get to their profile from the 'recent projects' list on the homepage, so I figured the same logic applied here).

2. +1 I like the idea of giving an indication that you can add a while list. Instead of the emoji pointing up, text like "You can also whitelist with @your-domain.com" (in other similar dialogs, the grey info sections content is generally applicable to the above white action section so the extra visual emphasis from the emoji here might not be needed)

3. I agree with the notification issue. Because notifications are defined to trigger in a consistent place, my recommendation is keeping the existing the notification but also adding additional visual feedback closer to  your content. 
For example, the add button could transition temporarily (for like 1 or 2s) to an 'invite sent' success state. It'd still be interactable in the same way during this time.
screen shot 2018-08-15 at 11 58 23 am

I've been planning to implement a pattern similar to this with the 'copy' text field buttons in the editor as well

4. +1 to copy suggestion

ttseng:
@pketh 
1. I'm not sure I fully follow - it seems like clicking on avatars on Glitch generally leads you to the user profile, unless it triggers a dropdown where you have more options.  I couldn't find another example of a "Profile" or "Your Profile" link other than for the user avatar in the navbar.  For consistency, it may be worth having the same layout for the user dropdown from the navbar + user dropdown from teams page, i.e.,
image

(this isn't really essential though - just a note)

2. Cool - I like that wording.
3. Ok -  keeping existing notification UI and adding redundant message closer to the input sounds good.  I think it might be disorienting to temporarily remove the add btn – we could also just add a pill notification close immediately next to the add btn, such as
image

pketh:
1. I think that'd be really cool (I don't think it should be done in this PR tho - could you make a card with these screens in the community board so we can do it later?)
3. I think the disorientingness may be a factor of how that transition is implemented (this is a case where animation/tweening can maintain the idea that this button is still the same thing). In the case of the textfield copy button context in the editor, or in the mobile viewport usecase here. Adding a separate longer notification inline would instantly shift around the layout (and then shift it back once the notification disappeared) which might be jarring irl.

ttseng:
@pketh 
1. Sure
2. Oh I forgot that the "Add" btn is inline.  
I guess we could either have the Add btn always on a new line so that the notification always appear to the right of it, without shifting any content above or below it, or we can swap the add btn + notification temporarily.  The plus side of always having the add btn visible is that the user wouldn't need to wait until the success notification fades back out to continue adding users.  But perhaps this is something best prototyped + tested before deciding.

pketh:
in the case of an animated transition, it would only change the display of the button. it's behavior when you click on it would remain the same (would open the add user thing), so the user wouldn't need to wait. yeah I'll work on a prototype for this

https://www.notion.so/glitch/context-distance-between-notification-and-user-action-of-adding-a-teammate-is-high-426dba060e214fa4978915ba59d1c2a4

pketh:
also, greg, I like the way you handled the whitelisted domain tooltip/pop description 👍

GregWeil:
Updated with inline invite messages from pirijan

src/presenters/pop-overs/add-team-user-pop.jsx
```diff
@@ -81,37 +64,98 @@ class AddTeamUserPop extends React.Component {
...
+  }
+    
   render() {
-    const isLoading = (!!this.state.maybeRequest || !this.state.maybeResults);
+    const {maybeRequest, maybeResults, query} = this.state;
+    const isLoading = (!!maybeRequest || !maybeResults);
+    const results = [];
+    // this regex is basically (a)?@(b.c)
+    // any match means we have a valid domain in group 2
+    // group 1 is optional, but its presence means we have a full email
```
judeallred: strongly recommend you pull in and use https://www.npmjs.com/package/email-addresses for this

(also, jack wrote it!  cool ;-) )

In the "@domain.com" case, the parser would return null and you could then handle that as a dedicated separate test.

judeallred: (this also means that if somebody copy/pastes one of the more advanced forms of email addresses, this'll handle it just fine)

GregWeil: Ok, I think I can test for domains by appending asdf in front and seeing if that parses.

GregWeil:
- Parse emails and domains using the email-addresses module
- Use favicon-fetcher.glitch.me to get domain favicons

Still waiting on invite emails, but whitelist joining is working

judeallred:
style bug when search results are empty:
image

judeallred:
preexisting bug, but more apparent now:
 -> try to search for someone by name, e.g. "esteban molina"
image

expected:
  you can type in "esteban molina"

actual:
  React is preventing trailing spaces in the input, so you have to first type in 'estebanmolina', then inject a space in the middle to create 'esteban molina'.

( I'm pretty sure this is my fault 🏅  )

GregWeil:
style issues are fixed. Trailing spaces should have already been fixed in this branch?

ttseng:
A few smaller styling issues:

1. Spacing around the Thanks section is a little off right now.  
image
I would suggest this styling change:
image

image


2. When hovering over the Thanks section, the thanks counter does not invert color, which makes it harder to read:
image

GregWeil:
image

:+1:

GregWeil:
commented out the email invite stuff, now this branch is just whitelisting and style cleanup

judeallred:
feels good!  ship it!

judeallred:
here's a tiny bug in edge; not a blocker imho:
image

GregWeil:
fixed that styling bug in edge. looking forward to proper details support

PR #85, opened 2018-08-15
create-team-pop.jsx: element with onClick should be an interactable

remyabel:
Fixes the following errors:

``` 
103:9  error  Visible, non-interactive elements with click handlers must have at least one keyboard listener  jsx-a11y/click-events-have-key-events
103:9  error  Non-interactive elements should not be assigned mouse or keyboard event listeners               jsx-a11y/no-noninteractive-element-interactions```


judeallred:
Thanks!  Your change is live.

PR #88, opened 2018-08-15
Inline update user

pketh:
https://inline-update-user.glitch.me/

builds on top of https://github.com/FogCreek/Glitch-Community/pull/84 (this'll be much easier to review after that's merged)
shows an `inline-notification` when a user is invited to join a team. notifications are position:absolute so shouldn't affect page layout


pketh:
@GregWeil if you approve of this , maybe it makes sense to merge this into your branch, and close this pr

src/presenters/pop-overs/add-team-user-pop.jsx
```diff
@@ -64,6 +64,9 @@ class AddTeamUserPop extends React.Component {
   togglePopoverAnd(func, ...params) {
     this.props.togglePopover();
     func(...params);
+    if (func.name === 'inviteUser' || func.name === 'inviteEmail') {
+      this.props.updateInviteSent()
+    }
```
GregWeil: I don't think we should be identifying the function to call based on its name, it seems pretty fragile. This also means we show the 'sent invitation' inline before we've actually hit the server (and the popup won't correct itself if something goes wrong)

pketh: i can try somethign else. 

re timing, I think it's actually better to show the notification before the 200 comes back. If an error comes back we can display the error, but in 99% of times it's gonna work so assuming success lets us make the ui feel faster

pketh:
@GregWeil split the function into three less clever methods, that each call the inline-notifier and the team-editor

pketh:
now invitees are displayed in the inline notification, replaced the global notification

pketh:
ready for re-review

GregWeil:
Looks good! I guess I'll merge this into superb-guide as an expansion of that review

GregWeil:
Actually, can you clean up the linting errors first?

pketh:
ya sure, one sec

pketh:
ok fixed

PR #89, opened 2018-08-16
Nova chance: Categories are curated and static, more complex currentUser handling

GregWeil:
The list of categories are kept in server memory and put in the page html as `window.CATEGORIES`. I also moved the script tags up into the head with defer, since it kicks off script loading slightly earlier and is supported in pretty much every browser.


judeallred:
This is good if categories were long for this world....    I'm fairly confident that Collections will replace them, and since in the meantime they're very unlikely to change, I'd like to suggest an alternate:

-> create curated/categories.js
-> put [this](https://api.glitch.com/categories/) in it
-> use that instead; delete /cache/ and friends

views/index.ejs
```diff
@@ -66,8 +70,5 @@
         serverName: '<%= PROJECT_DOMAIN %>'
       }).install();
     </script>
-    <% for (let i = 0; i < scripts.length; ++i) { %>
-    <script src="<%= scripts[i] %>" async></script>
-    <% } %>
```
judeallred: my one nervousness about this is that (I think) it no longer guarantees that Raven will load before our bundle, thereby giving us a way for bundle failures to go uncaptured.

Might be ok;  if it works some of the time then that would get us enough data.

judeallred: yeah, on further thought I think what you've done is good,  keep it!

we should also probably put raven into the bundle as its own chunk. that's a different topic.

GregWeil:
Pivot!

So now this review is about currentUser handling. The current user is now stored as `community-cachedUser`. If it ever does not match `cachedUser`'s id or token it get dumped and reloaded.

Also I made the logo fish check the time every five minutes so you don't have to refresh the page for them to switch to sunset/night fish

GregWeil:
Updated with stronger error handling. Now if you have a bad persistentToken or an id that doesn't match the site will either reset your cachedUser to the correct id for your token or log you out if the token isn't valid at all.

src/presenters/header.jsx
```diff
@@ -10,21 +11,43 @@ import NewProjectPop from "./pop-overs/new-project-pop.jsx";
...
+        hour: (new Date()).getHours(),
+      });
+    }, moment.duration(5, 'minutes').asMilliseconds());
+  }
+  
+  componentWillUnmount() {
+    window.clearInterval(this.interval);
+  }
+  
+  render() {
```
judeallred: fun :-)

PR #90, opened 2018-08-20
Abounding witness -- fix LINT errors in prod; remove 'profile' link from team user pop

judeallred:


judeallred:
something went wrong with this pr; tryin again

PR #91, opened 2018-08-20
Abounding witness -- fix LINT errors in prod, remove 'profile' link from team pop

judeallred:


ttseng:
new team profile popover:
image

user-options-popover:
image

Note that the styling is not exactly the same as user-options-pop is more narrow than team-user-info-pop

GregWeil:
On first glance I thought the name/login were too low compared to the avatar for non admins, but now it seems ok so maybe I just had to get used to it.

I think the header dropdown might look better with the light gray background that the team user dropdown has.

Should the whole profile area act as one large link?

Having a header in the user popover looks really nice, though.

ttseng:
@GregWeil I also thought that it might be good to have a grey background in the navbar user dropdown, like there is for the teams dropdown.  

But there is also a grey section at the bottom of the navbar user dropdown (for new stuff / support / sign out), and I didn't think those sections were related enough to have them have the same background color.

The only potential issue with having the whole profile header act as a link is that there is a tooltip for the Team Admin badge under the teams popover.  We could, however, just have the entire profile header act as a link just for the navbar.  

Going to implement these changes and add to the pull request...

ttseng:
ok - I was able to get a version I was happy with. a summary of changes:
* consolidated the user avatar + info to a single row for the user-info pop-over
* change background color of profile header on hover
* clicking anywhere in the header brings user to profile page
user-options-pop-over

judeallred:
I like this look;  On *my* username I feel like it looks a little off:
image

I'm OK with it, but I think my instinct is that the "@JudeAllred" should be perhaps a smaller font size or otherwise inset in some way?   I think the vibe i'm getting is that there's a header (my name) and a subheader (my login), and the subheader feels too weighty

judeallred:
I might just need to get used to it, though

judeallred:
I might be imagining it,  but did the 'Team Admin' callout get closer to the @username?  the spacing feels a little off there to me:
image

judeallred:
(@ttseng ping me when you want to deploy)

GregWeil:
There's some space around the profile that highlights but isn't actually clickable, which feels odd but isn't a huge deal

pketh:
is there a specific reason behind why the whole section is clickable? (the bigger target is nice, but it seems like you're having to do a lot of backflips re background highlighting and padding that might not be useful if the main goal here is "clicking on avatar = go to profile page")

judeallred:
my 2c: I like the whole area being clickable

ttseng:
I think having the avatar-only be clickable to go to the user's profile is useful if other sections in the header should link elsewhere.  However, if there is only one link in the entire container, my instinct is to make the entire header clickable.  

I will fix the issue with parts of the header not being clickable and remove the background cover on hover - the change of the cursor to pointer is probably enough of a visual queue.

And will include some small padding updates to ensure the Admin badge is not too close to the user handle.

styles/pop-overs.styl
```diff
@@ -148,18 +148,32 @@
...
   left: 2px
-  top: 28px
+  top: 28px  
+  
+.team-user-info-pop, .user-options-pop
+  section:first-child
+    border-bottom: 1px solid secondary
     
-  .pop-over-info
+  .pop-over-actions.user-info
```
pketh: in stylus , it's cleaner to nest classes instead of declaring the multiple classes together
ie
```
.pop-over-actions
  rule1
  .user-info
     rule2
     rule3
```

or if `.user-info` is used elsewhere, and you just want to override bits for the `user-options-pop`, you can do
```
.user-options-pop
   .user-info
       myoverrides
```
this'll prevent styles for this from affecting other components

styles/pop-overs.styl
```diff
@@ -148,18 +148,32 @@
...
+  .pop-over-info.user-info
     .avatar
       width: 44px
       height: 44px
+  .pop-over-actions.user-info, .pop-over-info.user-info
+    .avatar
       border-radius: 8px
       vertical-align: middle
       margin-right: 10px
+      margin-bottom: 10px;
```
pketh: having a `margin-bottom`, breaks the intent of `vertical-align: middle` (ie having the avatar image vertically centered next to the text)

src/presenters/pop-overs/user-options-pop.jsx
```diff
@@ -95,17 +107,22 @@ const UserOptionsPop = ({toggleUserOptionsPop, userLink, avatarUrl, avatarStyle,
...
-      <section className="pop-over-actions">
-        <a className="button-link" href={userLink}>
-          <div className="button button-small has-emoji button-tertiary">
-            <span>Your Profile </span>
-            <img className="emoji avatar" src={avatarUrl} style={avatarStyle} alt="Your avatar"></img>
-          </div>
-        </a>
-      </section>
+      <a href={userLink} className="user-info">
+        <section className="pop-over-actions user-info">
```
pketh: now that you're not doing background color hover changes anymore, this should probably be a `pop-over-info` section again (matches other profile areas, visually grounds the content below it better)
screen shot 2018-08-21 at 8 20 59 am

src/presenters/pop-overs/user-options-pop.jsx
```diff
@@ -82,7 +82,19 @@ TeamList.propTypes = {
...
+  userLink,
+  avatarUrl,
+  avatarStyle,
+  teams,
+  signOut,
+  showNewStuffOverlay,
+  toggleCreateTeamPop,
+  userIsAnon,
+  userName,
+  userLogin,
```
pketh: this is less of a big deal, but when we've got components that specify a bunch of params like this, we should convert it to use {...props} instead:

```
const UserOptionsPop = ({...props}) => {
// replace stuff like {toggleUserOptionsPop} with {props.toggleUserOptionsPop}
}
```
this makes the component less brittle to future refactors

styles/pop-overs.styl
pketh: the margin-bottom also adds extra margin on the team-user-info pop

src/presenters/pages/user.jsx
```diff
@@ -17,7 +17,7 @@ function syncPageToLogin(login) {
   history.replaceState(null, null, `/@${login}`);
 }
 
-const NameAndLogin = ({name, login, id, isAuthorized, updateName, updateLogin}) => {
```
pketh: does this work with anon users (where we don't have a name, or login . but if we wanted could link them to their anon profile page with /@123)

pketh: this is another one of those components that can benefit from using {...props}, instead having to declare all params. see comment below

styles/pop-overs.styl
pketh: (you might also want to try `vertical-align: top` , not sure if that'll look better with long names that wrap)

src/presenters/pages/user.jsx
ttseng: this one I wasn't sure how to best fix - @judeallred maybe we can talk about it when you get in.

PR #92, opened 2018-08-20
Profuse tea: rename teams and change their url

GregWeil:

Added a team name input, focusing it makes the url field visible (only for team admins)
Verified teams show the icon at the end of the name input, same with the project privacy toggle. I think it matches the unauthenticated page more, but let me know what you think.


GregWeil:
Forgot to mention: will have to wait on case 3323310 to block empty string team names.

pketh:
i'm seeing an alignment issue on the url field 
screen shot 2018-08-21 at 8 35 18 am
I also see this on the user page one too. Using mac safari

pketh:
i want to redo that verified icon later, but in the meantime can you remove the `margin-left: 5px;` from `.verified` ?

ttseng:
Ok, I thought more about it, and I am still not completely convinced that having a separate short-url is a good idea.

The only pro I could think of is that the team name could be unique, while the short-url needs to be unique.

But it leads to these issues:

**Issue 1**
Say someone on a team renames the short-name but maintains the original team name, e.g. @test-team to @hello-world, but keeps the team name as "Test Team"

Now [https://glitch.com/@test-team](https://glitch.com/@test-team) no longer exists, even though the team still appears to be named "Test Team" when I open up the team page (now at [https://glitch.com/@hello_world](https://glitch.com/@hello_world)).  
The only way I can reveal what the short-url is is to edit the team name, which is a bit confusing (I guess I could also look at the URL of the page too).

Everywhere else in the UI, such as the user-options-pop-over, I would still see the team name as "test team", e.g.,
screen shot 2018-08-21 at 4 25 59 pm

So generally the incongruence between the team name + short-url can lead to confusion about where to actually find the team

**Issue 2**
^ the case I described in the comment above, where when you click on the team name, the short-url input field suddenly appears.  Which do I edit?  And how are they related to one another?  This is not immediately obvious, as the placeholder to tell you the @ input field is for the short-url does not actually appear on default (need to remove all characters from @ input field to see placeholder)

I would suggest maintaining a link between the team name and short-url, and always display the short-url explicitly under the team name (read-only, not editable), something like:
image

pketh:
that makes sense to me and lines up pretty well with the original proposed design at https://www.are.na/block/2395341 , where the url is just an auto kebab cased version of the name. I think @ names may be prevalent/common enough on the internet that listing them with just `@test-team` like we do in the pop overs may read clearly enough. 
We've been using lighter colors (eg light grey) to demarcate secondary info rather than use font-sizes smaller than 14/13px. This approach may apply similarly here.

ttseng:
@pketh got it - based on your feedback, here's a revised suggestion

image

GregWeil:
I agree that the @ name should always be visible, but I think teams are going to want to separate their name from their url. We already have a bunch of teams with urls that drop words or are abbreviations.

pketh:
we can treat those as special cases, where we display as is what we get from the db. However if one of those teams changes it's name, it's url gets changed as normal.

saleswise, teams that are concerned about specific vanity urls that don't match their name can go through sales for manual changes. Although, from my observations this is one of those things ppl only care/think about if they're specifically asked to care about it.

pketh:
the possibility of @ mentions for users in the future, and generally higher expressiveness around names for individuals, means that we might want to maintain manual naming for users though?

ttseng:
yeah, the separation of user names & handles makes sense to me! 

for teams, we can keep the team name & handle linked for now and see how many requests we have for manually overriding team names.  I suspect the same as @pketh - I'm guessing this default behavior will work for the majority of cases, and we can deal with special requests manually.

GregWeil:
Updated with auto kebab urls

ttseng:
@judeallred these are good questions.  do we have a list somewhere of all the collections that have been created up to this point?  or can I do a database query or something?

judeallred:
@ttseng http://glitch-team-manager.glitch.me/ has the existing teams

judeallred:
(btw @greg I reviewed the code and it looks good to me.  I'm just poking at the UX; I don't consider it a blocker, though I do imagine UX refinements will be coming)

pketh:
I think it's worth getting a second opinion on meg, specifically about whether this is a deal break or strong concern for potential teams. 

In terms of naming handling I lean to the first option described
> will their url automatically (and irreversably) change if they edit their name?
I think we should optimize for the 1000s+ of new teams over the existing 125, unless we have a good reason to do otherwise.

In terms of error handling, the create-team-pop will do something similar where as you type your new name, a field error indicator on the focused input will show you if the generated team url is taken

GregWeil:
Ok, updated again with separate fields for team name/url. When you type into the name field, the url field gets updated in real time if it matches.

As part of this, I did some heavy refactoring on the EditableField component. It should work exactly the same, but is now made up of two components.

PureEditableField (probably a better name out there) is stateless and just renders the input and an error message
OptimisticValue stores the input value until a debounced api call happens, then it reverts back to the value from props if it succeeded

GregWeil:
Small update: I switched the markdown components from functional to `React.PureComponent`, which bumps isaac's profile page from 27s to 17s on my pc. (18s to 10s for prod builds)

GregWeil:
Updated again per today's meeting. It now works just like the user page, with two uncoupled fields. There's an api bug right now that lets you set a team name to an empty string, but otherwise it should be ok.

GregWeil:
Api bug is fixed

PR #93, opened 2018-08-21
Pewter-stitch: team analytics data bins explicitly grouped by day

pketh:
fixes https://www.notion.so/glitch/teams-analytics-not-showing-up-to-today-s-date-on-team-page-6f378036dc8a47038d2f0369e191106b

for 2week/4week ranges, this replaces d3Array histogram with something less magical, that ensures that each day is in it's own bin for the graph. 

in action: (it's aug 21 today. chart now shows an explicit tick for aug 21, as well as all other days)


src/presenters/includes/team-analytics-activity.jsx
```diff
@@ -2,36 +2,52 @@ import React from 'react';
...
+const groupByRegularIntervals = d3Array.histogram().value(function(data) {
+  return data['@timestamp'];
+});
+
+const createBins = (buckets, currentTimeFrame) => {
+  let bins = [];
+  if (currentTimeFrame === "Last 24 Hours") {
+    return groupByRegularIntervals(buckets);
+  } 
+  bins = groupByTime(buckets, '@timestamp', 'day'); // supports 'day', 'week', 'month'
```
judeallred: you could use `const bins = ...` here, rather than defining it up top

src/presenters/includes/team-analytics-activity.jsx
```diff
@@ -2,36 +2,52 @@ import React from 'react';
...
+      if (!timestamp) {
+        timestamp = data['@timestamp'];
+      }
       totalRemixes += data.analytics.remixes;
       uniqueAppViews += data.analytics.uniqueIps;
       // referrers.push(data.analytics.referrers)
     });
     histogram.push({
-      time: bin.x0,
+      time: timestamp,
```
judeallred: looks like if a given bin is empty, the 'time' in the histogram will be `undefined`;  that ok?

src/presenters/includes/team-analytics-activity.jsx
```diff
@@ -73,10 +89,10 @@ const renderChart = (c3, analytics, currentTimeFrame) => {
...
         tick: {
           format: dateFormat(currentTimeFrame),
-          fit: true,
-          culling: {
-            max: 12
-          },
+          // fit: true,
+          // culling: {
+          //   max: 12
+          // },
```
judeallred: testing code?

src/presenters/includes/team-analytics-activity.jsx
pketh: Ya , it’s just for race conditions

pketh: But I might not need that anymore

src/presenters/includes/team-analytics-activity.jsx
pketh: in the language of stats that I'm a newb at, bins are the containers of data for a single point on a histogram. put another way, disparate data points (buckets) gets 'binned' (aka grouped) into 'bins' for consolidated display on a graph.

buckets are what's returned from the api, I'm not particularly sure why @etamponi chose 'buckets'. I'm sure there's a reason in context. 

That being said, I'm of two minds here
- mind1 = to understand this code you shouldn't have to understand bullshit stat words
- mind2 = using the 'industry' terms is far more googleable, and it's easier to find alternate libraries/approaches that do similar things
- mind2.5 = I could add an explanation of binning as a comment here

what do you think @judeallred ?

pketh:
re difference between buckets and bins:

in the language of stats that I'm a newb at, bins are the containers of data for a single point on a histogram. put another way, disparate data points (buckets) gets 'binned' (aka grouped) into 'bins' for consolidated display on a graph.

buckets are what's returned from the api, I'm not particularly sure why @etamponi chose 'buckets'. I'm sure there's a reason in context. 

That being said, I'm of two minds here
- mind1 = to understand this code you shouldn't have to understand bullshit stat words
- mind2 = using the 'industry' terms is far more googleable, and it's easier to find alternate libraries/approaches that do similar things
- mind2.5 = I could add an explanation of binning as a comment here

what do you think @judeallred ?

src/presenters/includes/team-analytics-activity.jsx
pketh: ya according to the docs it's supposed to limit the number of ticks (useful to ensure the graph is readable on a phone). In practice, it doesn't seem to do anything tho

src/presenters/includes/team-analytics-activity.jsx
pketh: if no one has visited your app on a day, there won't be a bin at all for that day. Because all data points inside a bin have timestamps it should be impossible to have a bin without a timestamp

pketh:
thanks for the review jude, the latest commits should address everything reported. I'm not 100% sold on the comment explanation approach tho. let me know how it feels to you

pketh:
updated to add changes for https://www.notion.so/glitch/77466b6f4229453bbcf7722358c929c2?v=e91a07a4cb9a486885d7aa9ea04c0377&p=20790275512e4ab2bd4579540bd0a38a

pketh:
screen shot 2018-08-23 at 12 23 12 pm
no longer using 'uniques'

PR #95, opened 2018-08-27
proxy about.glitch.me and about-glitch.glitch.me -- case 3323312, case 3317554

judeallred:
https://glitch.manuscript.com/f/cases/3323312/Serve-about-and-legal-from-about-glitch-project

### DIFF
--- a/server/proxy.js
+++ b/server/proxy.js
@@ -16,6 +16,7 @@ module.exports = function(app) {
   
   // Pages hosted by 'about.glitch.me':
   [
+    'faq',
     'react-starter-kit',
     'website-starter-kit',
     'forteams',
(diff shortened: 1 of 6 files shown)


judeallred:
live at https://navy-apartment.glitch.me/  https://navy-apartment.glitch.me/about https://navy-apartment.glitch.me/legal

n.b. '/about' won't show in prod until a cloudfront config change happens, but that can be handled separately from this deplyoment

judeallred:
I'm expanding this to include case 3317554

GregWeil:
It took me a bit to figure out `pathOnTarget` but that was mostly because I mixed up `about-glitch.glitch.me` and `about.glitch.me`.

Could you update the footer and team page to use relative urls instead of glitch.com/

judeallred:
@GregWeil updated

judeallred:
n.b. -- gareth mentioned that there's special redirect logic in the /FAQ/ redirect that's coded in about.glitch.me -- i need to make sure I didn't break that

judeallred:
I'm going to sit on this until I return from XOXO

PR #96, opened 2018-08-27
Aquatic thyme -- developer feature switches

judeallred:
Allyson and whipped up some developer feature switches.

What do you think?

https://aquatic-thyme.glitch.me/secret


src/presenters/pages/secret.jsx
```diff
@@ -0,0 +1,83 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import Helmet from 'react-helmet';
+import {DevToggles} from '../includes/dev-toggles.jsx';
+
+class SecretEffectsOnMount extends React.Component {
+  componentDidMount() {
```
judeallred: hmm, the intent was that people would never navigate away from here...  but since that might happen in the future, we'll want to use a fixed background instead of modifying body. wompwomp.

GregWeil:
The page looks great!

I think the default handling is off though. Once you've modified your keys then future keys will always be disabled, since it'll use the stored list and never check the defaults.

judeallred:
I was thinking about feature switch behaviors last night and I had an idea--
  I think "default state", /enabled/disabled/ is an antipattern.

why?  because you should never need a feature switch whose state is 'enabled', because this indicates that either:
 1. it's being used to _selectively disable_ a feature for some users,
 or 
 2. it's being turned to 'enabled' as part of its deployment, which permits you to enable a feature without removing its features switch... which leads to the delayed task of cleaning out the feature switches.

so i'm thinking this:
 1. I should remove the 'default' capability; all feature switches default to 'off'.
 2. /secret/ can be used to opt-in to features switches.

I believe this preserves the ability for the 'devToggles' to allow people to opt in to dev features, while helping protect us from antipatterns and extra complexity.

GregWeil:
I'm ok with that. If deploying the switch goes badly, swapping back is probably better than disabling the switch anyways.

judeallred:
@GregWeil  @whimsicallyson  updated to fix up above;  thoughts?

judeallred:
(hmm, changesets aren't showing up... stand by.)

judeallred:
ah, github is having issues: https://status.github.com/messages   

ok, just gotta wait for the changesets to show up

judeallred:
@GregWeil  ok changesets are in

PR #97, opened 2018-08-28
Aquatic thyme -- konami code activates /secret

judeallred:


styles/pages/secret.styl
```diff
@@ -22,11 +22,6 @@
        text-align: center
        &.lit
          background: #FB994A
-  
-  .reset-to-defaults
```
judeallred: dead css cleanup

src/presenters/footer.jsx
```diff
@@ -34,6 +38,9 @@ export default function Footer() {
         <img className="for-platforms-icon" src={srcForPlatforms} alt=""/>
         <span className="for-platforms-text">Glitch Teams</span>
       </FooterLine>
+      <Konami>
+        <Redirect to={"/secret"}/>
```
GregWeil: This should use `push={true}` so you can back button out of the secret

GregWeil: Also the footer seems like an odd place for this, I'd suggest bumping it up to at least the layout if not client.jsx

src/presenters/includes/konami.jsx
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
```
pketh: I think it  would make sense to call this file `secret.jsx` , to match the `/secret` route and the secret.styl file

pketh:
what does /secret do? I couldn't find it in the diffs

src/presenters/includes/konami.jsx
judeallred: not really, since secret.jsx exists?

src/presenters/footer.jsx
judeallred: Layout was my other candidate, I'll go for that.

judeallred:
@pketh see https://github.com/FogCreek/Glitch-Community/pull/96

src/presenters/pages/secret.jsx
```diff
@@ -8,9 +8,13 @@ class SecretEffectsOnMount extends React.Component {
   componentDidMount() {    
     //try to play the secret sound:
     const audio = new Audio('https://cdn.glitch.com/a5a035b7-e3db-4b07-910a-b5c3ca9d8e86%2Fsecret.mp3?1535396729988');
-    audio.play().then( () => {
+    const maybePromise = audio.play();
+
+    // Chrome returns a promise which we must handle:
+    maybePromise && maybePromise.then( () => {
```
GregWeil: If you `await audio.play()` instead you can try/catch it instead, which might be a bit cleaner. not a blocker

judeallred: Agreed;   gonna ignore it tho.

PR #98, opened 2018-08-28
remove-team-users: Remove users from a team and projects in it

GregWeil:
https://remove-team-users.glitch.me/@test-team

I had to shuffle around project loading on team/user pages so we could reload their users, but I think it's better this way. Team and user pages now have project members everywhere rather than just in `EntityPageProjects`. The popover reloads the user being removed rather than relying on project members, in case they've joined any projects after the page loaded.

Side note: can someone give me admin back on test-team?


judeallred:
@GregWeil admin restored

judeallred:
title looks a little weird of the name is long: 
image

judeallred:
other than that, functionality looks good to me.

I noticed `warning.js:33 Warning: Can't call setState (or forceUpdate) on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in the componentWillUnmount method.
    in AddTeamUserPop (created by PopoverContainer)` in the logs.

'select all' looks disabled to me,  but i think that's an existing style convention?   not sure.  in this spot it pops out to me more, maybe because the background is white

src/presenters/includes/team-users.jsx
```diff
@@ -14,13 +14,17 @@ export const TeamUsers = (props) => {
   let userIsTeamAdmin = (user) => {
     return props.adminIds.includes(user.id);
   };
+  
+  // let removeUserPopVisible = () => {
```
judeallred: ?

src/presenters/entity-page-projects.jsx
```diff
@@ -32,12 +32,14 @@ const EntityPageProjects = ({projects, pins, isAuthorized, addPin, removePin, pr
     </React.Fragment>
   );
   
+  /*
   projectOptions = _.mapValues(projectOptions, function(projectOption) {
     return async (projectId, userId) => {
       await projectOption(projectId, userId);
       reloadProject(projectId);
     };
```
judeallred: ?

pketh:
re https://github.com/FogCreek/Glitch-Community/pull/98#issuecomment-416736322 : we can avoid this issue if we use their user avatar instead in the title
```
< remove [avatar]
-------------------
Also remove [user-name] from these projects
```

src/presenters/pop-overs/team-user-info-and-remove-pop.jsx
```diff
@@ -0,0 +1,331 @@
...
+                  checked={this.state.selectedProjects.has(project.id)} value={project.id}
+                  onChange={this.handleCheckboxChange}
+                />
+                <img className="avatar" src={getAvatarUrl(project.id)} alt={`Project avatar for ${project.domain}`}/>
+                {project.domain}
+              </label>
+            ))}
+          </div>
+          {this.state.userTeamProjects.length > 1 && (
+            <button className="button-small button-tertiary"
```
pketh: based on feedback about the button looking 'disabled' in this context, how does it look if we remove `button-tertiary`?

ttseng:
A few comments:

1. Currently, we show the option to remove users from projects even if they aren't part of any projects in the team.  Can we just omit showing them the second screen? 
image

2. When removing users from projects, it looks like the target click area is actually the whole row, not just the checkbox. That's pretty convenient.  Could we change the cursor to a pointer, both when the user hovers over the checkbox but also when they hover over the row?  That will make it clearer that they can click anywhere.  
click-target-area
* Also notice in the example above there is some shifting when you click on a project with a long name.

Also noticed a few styling issues with search results when adding users, but will add to the superb-guide PR.

GregWeil:
image

Removing the second panel means the menu jumps a bit since it shows the spinner for a moment. I was aiming for a confirmation that you don't have to worry about them being in any projects. What if the 'also remove...' text switched to something like 'they aren't in any projects here' rather than appending something?

pketh:
+1 , 
can you do:
`[username/anon or userAvatar] is not a member of any projects`

GregWeil:
image
image

PR #99, opened 2018-08-30
Expose create team pop

pketh:
https://expose-create-team.glitch.me/
https://glitch.com/edit/#!/expose-create-team

updated new-team-pop adds random team names and error checking as you type

requires https://github.com/FogCreek/Glitch-Community/pull/98 and https://github.com/FogCreek/Glitch-Community/pull/84 to merge first


pketh:
still wip, not ready for review yet

pketh:
random name possibilities in https://glitch.com/edit/#!/expose-create-team?path=src/presenters/pop-overs/create-team-pop.jsx:24:1

GregWeil:
When you create a new team the loader goes away for a moment before you get navigated to the new team. It should probably just stay around until the popover is closed by the navigation rather than flickering the input/button again.

src/presenters/pop-overs/user-options-and-create-team-pop.jsx
```diff
@@ -0,0 +1,428 @@
...
+      backgroundColor: '',
+      hasCoverImage: false,
+      isVerified: false,
+    })
+    .then (response => {
+        this.setState({ isLoading: false });
+        window.location = `/@${response.data.url}`;
+      }).catch (() => {
+        this.setState({
+          isLoading: false,
```
pketh: @GregWeil you should be able to fix https://github.com/FogCreek/Glitch-Community/pull/99#issuecomment-419553286 by removing 142. I'm gonna be stuck on editor stuff for a while, can you merge/release for me when you have time? Thx!

GregWeil:
Fixes and cleanup, hidden behind the add-users dev toggle.

GregWeil:
I don't think there are any blockers

judeallred:
other than my comments above, lgtm to ship when you're ready

GregWeil:
If a page is loaded with `#create-team` the create team pop will start open. It'll open even if you're an anon user, but not if you're logged out entirely (since the user popover won't exist, and there wouldn't be any users to put on the team)

pketh:
if a user who has never visited glitch.com visits `#create-team` , what's expected to happen?

GregWeil:
It'll load the page normally and not show the popover

GregWeil:
(it'll open the popover now because a separate pull always creates anon users)

PR #100, opened 2018-08-30
Expose delete team pop

pketh:
https://expose-delete-team.glitch.me/
https://glitch.com/edit/#!/expose-delete-team

exposes the ability for a team admin to delete their team. the ability to undelete a team is hidden in this release (it's blocked by email/backend pieces)

screen shot 2018-08-30 at 4 09 14 pm


pketh:
i think what i'm trying to capture in the copy is that the projects become regular projects. But if you were concerned about team members being able to still join a project why would you be deleting your team?

pketh:
p.s. just like w the other PR, because I'm in editor land, could you edit/merge this for me when you get the chance?

GregWeil:
Fixed merge conflicts, moved behind 'delete-teams' toggle.

src/presenters/pop-overs/delete-team-pop.jsx
```diff
@@ -56,23 +55,32 @@ export class DeleteTeamPop extends React.Component {
...
+         */}
+
       </dialog>
     );
   }
 }
 
+export const DeleteTeamPop = withRouter(props => (
+  <NotificationsConsumer>
+    {notifyFuncs => <DeleteTeamPopImpl {...notifyFuncs} {...props}/>}
```
judeallred: I don't think the "Impl" pattern is a great thing to adopt.  Then when you nest things multiple times, you get an ImplImpl and so on, and it means that the "base "container of "DeleteTeamPop" is kind of vacuous.

I'd be tempted to call this DeleteTeamPop a DeleteTeamPopContainer,  and then the DeleteTeamPopImpl a DeleteTeamPop,  although admittedly that's not quite what the container pattern is meant as.  

Maybe DeleteTeamPop  becomes DeleteTeamPopWithNotifications or DeleteTeamPopNotificationWrapper which then calls to the DeleteTeamPop?

judeallred: This sort of wrapper is pretty common, so I'd like your feedback on settling on a standard way to go about it

judeallred:
PR status:  I had a style comment for @GregWeil , but no blockers.

src/presenters/pop-overs/delete-team-pop.jsx
GregWeil: My thinking was that the exported component is the 'true' DeleteTeamPop, since it's the one everything else will import and use. The inner component is an implementation detail that isn't used anywhere other than the wrapper component in the same file.

PR #101, opened 2018-08-30
ebony-dinosaur -- bugfix for https://www.notion.so/glitch/Anonymous-User-Sign-out-5941dca5ecaa43b8b5ffe7fb73b64f30 

judeallred:
./src/presenters/pop-overs/user-options-pop.jsx:8/1663
./src/presenters/header.jsx:8/3


judeallred:
if this passes, feel free to deploy in my absence,  otherwise i'll ship it after i get back from xoxo :-)

PR #105, opened 2018-09-10
Nova chance: Show the noscript site if the browser doesn't support the latest js

GregWeil:
Most of our new features will cause parser errors if not supported, so the entire js file won't be run at all. Instead I have the main js file set a  global, which an inline script checks before either running the site or showing the fallback.

I flipped around our `<noscript>` usage. What used to be in the noscript is now in a `display: none` div and noscript is around some css that makes it visible. If js is enabled but not supported it's easier to make the div visible than to pull html out of a noscript tag.

I also added a 'bootstrap' tag for sentry that should be put on any errors that happen before the 'real' site starts (aka during feature detection).


views/index.ejs
```diff
@@ -59,15 +56,28 @@
           <p>To use this app, please enable JavaScript</p>
```
pketh: because this isn't strictly a `noscript` thing anymore, should this message be different? Like 'To use Glitch, you'll  need to use a browser with javascript ES6 support. We support the latest versions of Safari, Chrome ... etc.' I think you might want to finesse that to sound better/shorter though

GregWeil: Updated with two messages for js disabled vs not supported. "To use Glitch, you'll need a browser with support for more recent javascript features"

I couldn't figure out a good way to mention other browsers without feeling too  in depth.

judeallred:
This works really well, nicely done!   Yeah this is a huge upgrade from the blank white pages, so let's get it out there!

@ttseng would you be up for taking a whack at building a slightly nicer "old browser :-(" template for us, or would you prefer to postpone that? 

I'm fairly happy with the current one,  it just might be missing out on some opportunities to be a friendlier or have relevant links, etc.

judeallred:
uh oh, Edge @latest seems to be having some trouble with it:
image

I think that's a blocker :/  

repro:
https://live.browserstack.com/dashboard#os=Windows&os_version=10&browser=Edge&browser_version=17.0&zoom_to_fit=true&full_screen=true&resolution=responsive-mode&url=https%3A%2F%2Fnova-chance.glitch.me%2F%40judeallred&speed=1

GregWeil:
So the issue with user pages was that the server uses /users/byLogins?logins={} which is significantly slower than what the client does (/userId/byLogin/{} then /users/{}). For your profile (and a few others I tried) the api was taking over 7s. The server times out after 5s and pretends that whatever it was looking for doesn't exist. (the browser is probably going to repeat the request anyways)

judeallred:
@GregWeil  let's get the ball rollling on faster profile pages...   the api currently returns the profile data plus the data for all the projects.  I'll send a case to get a new api endpoint that omits the projects, so we can do a two-phase load

GregWeil:
Just switching it to use the get id then data requests like the client does drops it down to about 300ms (25ms + 280ms)

judeallred:
ah, nice

GregWeil:
I'm totally lost on that edge crash. It only happens on the index page if you aren't logged in (as a real user, anons crash too). It happens a few seconds after page load, everything renders and the embed is interactive for a second or so. At that point everything different in this pr should have run its course.

judeallred:
Hmm, I'll ping patrick on the edge team....

judeallred:
via google analytics, 77.34% of our Edge traffic is v17,  so we need to resolve this to keep the site up in Edge

judeallred:
Greg and I figured out that the Edge crash occurs with the interplay of:
 the Emoji Garden embed;  the Glitch Community site, and being logged out.

..and only in Edge v17.

This PR is unrelated -- more deets over https://fogcreek.slack.com/archives/CB57483TK/p1536699498000100

ttseng:
@GregWeil @judeallred I'm happy to work on updating the default page.  I'll take a look this morning

GregWeil:
Easiest way is to disable js in your browser. (you can do it from the dev tools settings menu in chrome/ff, not possible in edge, and idk about safari). Otherwise you can use browserstack and run an oldish ie.

ttseng:
just a little bit of styling on that page:
image

remixed nova-chance here: https://glitch.com/edit/#!/telling-rainstorm?path=views/index.ejs:57:16

pketh:
it looks good, 
my only suggestion is to drop the `!`. Especially in an error state/dead-end, the exclamation may not be interpreted as excitement. 

Are you going to include a mention about browser support? ie if you're using an old browser with js turned on, you'll still get this page (greg changed it to detect es6 support)

officially, we support only the latest version of evergreen browsers that can autoupdate themselves. so firefox, safari, edge, chrome

PR #106, opened 2018-09-12
remove use of user.email in favor of user.emails, case 3323347

judeallred:
./src/presenters/current-user.jsx:8/82


judeallred:
This is the only spot I noticed where user.email was in play;  user.emails seems to be in use everywhere else.

GregWeil:
`user.emails.find((email) => email.primary)` is slightly cleaner, but not a big difference

judeallred:
ah, good call-- will fix

judeallred:
thanks, fixed...  also i wasn't properly unboxing email before, so fixed that too.

PR #107, opened 2018-09-12
maze-basilisk -- proxy the _old_ about pages until case 3323312 is ready

judeallred:
per https://glitch.manuscript.com/f/cases/3323312/Serve-about-and-legal-from-about-glitch-project, we're not ready to serve the new pages yet,  however if I comment them out like this then i can go ahead and do the cloudfront config change


judeallred:
i noticed a couple proxy improvements, so i made them to ride along with this.

-> use urljoin instead of trying to do it myself and having to think too hard
-> use https, because even though a MITM between glitch.com and other glitch projects is unlikely, it's possible. 
-> use lower case string comparisons when redirecting (...so that if casing slips in in the future, we don't break stuff.)

judeallred:
i'll update this to merge cleanly with profuse-tea...

PR #109, opened 2018-09-14
Nova chance: cleanup for the user options pop

GregWeil:
Brought in helper functions and simplified props since this hasn't really changed since we removed observables.

Of course I'm realizing now that this is going to create some conflicts with the create team pop changes. 🤷‍♂️

### DIFF
--- a/src/presenters/pop-overs/user-options-pop.jsx
+++ b/src/presenters/pop-overs/user-options-pop.jsx
@@ -2,7 +2,8 @@ import React from 'react';
 import PropTypes from 'prop-types';
 
 import {getAvatarUrl} from '../../models/team';
-import {Link, TeamLink} from '../includes/link.jsx';
+import {getAvatarThumbnailUrl as getUserAvatarUrl} from '../../models/user';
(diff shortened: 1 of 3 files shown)


src/presenters/pop-overs/user-options-pop.jsx
```diff
@@ -66,15 +50,21 @@ const TeamList = ({teams, /*toggleCreateTeamPop, userIsAnon*/}) => {
       <CreateTeamButton toggleCreateTeamPop={toggleCreateTeamPop} userIsAnon={userIsAnon} />
       */}
       {teams.map((team) => (
-        <TeamButton key={team.name} {...team}/>
+        <TeamLink key={team.id} team={team} className="button button-small has-emoji button-tertiary">
```
pketh: do you need to add tabindex or do any a11y stuff now that these are just button classes?

GregWeil: `<Link>` renders as an `<a href>` so it gets all of the same accessibility stuff.

GregWeil:
Closed because I'm merging this into expose-create-team and dealing with the conflicts now

PR #110, opened 2018-09-18
Profuse tea: Added a very basic polyfill for trimLeft/trimRight

GregWeil:


judeallred:
LGTM, :shipit:!  thanks :-)

PR #111, opened 2018-09-19
Nova chance: Standardize NestedPopover

GregWeil:
Pretty much some separate code to handle the 'show an alt menu in place when you click the button'. I also added a popover component that renders a title with a back button. Through the magic of contexts you don't have to give it a function, it'll grab it from the closest menu.

I also split the team user info pop back into two files since it's gotten pretty large. The only differences are at the bottom of teamuserinfo where it uses NestedPopover and in teamuserremove where it uses NestedPopoverTitle.

Once either this or exopose-create-team gets merged I'll incorporate it into that menu.

### DIFF
--- a/src/presenters/pop-overs/popover-nested.jsx
+++ b/src/presenters/pop-overs/popover-nested.jsx
@@ -1,7 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
-export class PopoverNested extends React.Fragment {
+export class PopoverNested extends React.Component {
   constructor(props) {
     super(props);
     this.state = {
--- a/src/presenters/pop-overs/popover-nested.jsx
+++ b/src/presenters/pop-overs/popover-nested.jsx
@@ -13,10 +13,12 @@ export default class NestedPopover extends React.Component {
   }
   
   toggle() {
-    this.setState(({menu}) => ({menu: !menu}));
+    this.setState(prevState => ({menu: !prevState.menu}));
   }
   
   render() {
+    // Only use the provider on the sub menu
+    // Nested consumers want the back button, not the open menu
     if (this.state.menu) {
       return (
         <Provider value={this.toggle}>
--- a/src/presenters/pop-overs/popover-nested.jsx
+++ b/src/presenters/pop-overs/popover-nested.jsx
@@ -37,7 +37,7 @@ export const NestedPopoverTitle = ({children}) => (
   <Consumer>
     {toggle => (
       <button className="button-unstyled clickable-label" onClick={toggle} aria-label="go back">
-        <section className="pop-over-info team-user-remove-header">
+        <section className="pop-over-info">
           <div className="back icon"><div className="left-arrow icon" /></div>
           &nbsp;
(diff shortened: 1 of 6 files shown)


ttseng:
cool, this will be really useful for adding the nested pop-over elsewhere!

src/presenters/pop-overs/popover-nested.jsx
```diff
@@ -0,0 +1,53 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const {Provider, Consumer} = React.createContext();
+
+export default class NestedPopover extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      menu: false, // true for alt page, false for main
```
pketh: because both 'pages'/pop-overs are menus, `menu` as a term for this is a little unclear, maybe something like `initialContentVisible:true` or `secondaryContentVisible: false` would be clearer

src/presenters/pop-overs/team-user-remove-pop.jsx
```diff
@@ -0,0 +1,152 @@
...
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import Loader from '../includes/loader.jsx';
+import NotificationsConsumer from '../notifications.jsx';
+import {NestedPopoverTitle} from './popover-nested.jsx';
+import {getAvatarThumbnailUrl, getDisplayName} from '../../models/user';
+import {getAvatarUrl as getProjectAvatarUrl} from  '../../models/project';
+
+class TeamUserRemovePopImpl extends React.Component {
```
pketh: what's the `Impl` stand for?

src/presenters/pop-overs/team-user-info-pop.jsx
```diff
@@ -0,0 +1,141 @@
...
+
+
+// Team User Remove 💣
+
+
+// Team User Info or Remove
+// uses removeTeamUserVisible state to toggle between showing user info and remove views
+
+const TeamUserInfoAndRemovePop = (props) => (
+  <PopoverNested menu={() => <TeamUserRemovePop {...props}/>}>
```
pketh: oh so `menu` is the `popOverContents` and not the visibility state

GregWeil: It's both, thanks to bad naming :) State menu is visibility state and prop menu is the thing to be rendered instead when state menu is true. I'll update with less ambiguous names.

src/presenters/pop-overs/team-user-remove-pop.jsx
GregWeil: Implementation. I tried flipping the naming around, where we used to have TeamPop and TeamPopContainer it's now TeamPopImpl and TeamPop. That way the internal class has the special name and the one we actually export has the external name. We could probably do better than Impl.

GregWeil: Changed to 'Base', which is at least a complete word.

pketh:
just ping me here or in slack when u want a rereview

PR #113, opened 2018-09-24
update about and legal pages, case 32232212

judeallred:


judeallred:
https://thundering-skink.glitch.me/legal/
https://thundering-skink.glitch.me/about/

PR #114, opened 2018-09-25
Profuse tea: handle join team tokens

GregWeil:
I think this could be done better, but I'm not sure how. It turns out we can't just add you to the team single-page-app-ly because then you miss out on admin info and anything else that doesn't load if you weren't already a team member.

I also removed querystringify in favor of the standard `new URLSearchParams()`


src/presenters/pages/team.jsx
```diff
@@ -266,6 +269,47 @@ TeamPage.propTypes = {
...
+  async componentDidMount() {
+    const params = new URLSearchParams(this.props.location.search);
+    if (params.has('joinToken')) {
+      try {
+        await this.props.api.post(`/teams/${this.props.team.id}/join/${params.get('joinToken')}`);
+      } catch (error) {
+        if (error && error.response && error.response.data && error.response.data.message) {
+          this.props.createErrorNotification(`Failed to accept invite: ${error.response.data.message}`);
+        } else {
+          this.props.createErrorNotification('Failed to accept invitation');
```
pketh: for this and line `288` could we use something in natural language like 
> Invite failed, try asking your teammate to resend the invite
I'm assuming here that the main reason for a failed invite would be a old/expired join token

src/presenters/pages/team.jsx
```diff
@@ -266,6 +269,47 @@ TeamPage.propTypes = {
...
+        }
+      }
+      this.setState({reload: true});
+    }
+    this.setState({pending: false});
+  }
+  
+  render() {
+    if (this.state.reload) {
+      return <Redirect to={this.props.location.pathname}/>;
```
pketh: how does it feel irl? if it's not _too_ janky maybe it's fine for now. In theory, this was the nice thing about global models, where when the model for a team updates the right things change on a team

pketh: (i'm having trouble with a model thing on the editor rn, so that's far from perfect too)

pketh: should we show the user who just joined a new team , some kind of notification that they're not on the team? one way we could do that is a redirect on 200 with a diff queryString for joinSuccess, and then show a success notification when we see that

pketh:
Actually there might be a third way, I can update the email template to send the user to a new community route that serves an extremely simple page with ‘joining teamname...’ then we can do the call there and then just load the team page. That might create an overall less janky experience.

src/presenters/pages/team.jsx
GregWeil: Sure, I'll keep the specific reason in the console

src/presenters/pages/team.jsx
GregWeil: Externally it's not too bad. The only visible thing is that the loader kind of flickers when it switches from accepting the invite to actually loading the team. (It doesn't actually blink out or anything, but the animation resets) The redirect is single page app, so we can create a success notification when we redirect and it'll stay around.

pketh:
ok i'll update the email template later today to point to `/join/@:name/:code`

GregWeil:
Updated to work as a separate /join/@team/token route

PR #116, opened 2018-09-28
Delirious bite: metadata on ~project page

whimsicallyson:
https://delirious-bite.glitch.me/

uses prototype code from calico-spy (further fleshed out, added some error handling, etc) to show popular NPM packages and license types for the ~project page. I imagine there is still work to be done (Jude mentioned project.json files that aren't well-formed in the spec, and I doubt I handle them well/at all), but better to get eyes on it sooner rather than later!


pketh:
Is there a functional spec behind this? something that shows the goals, use-cases etc. for this?

judeallred:
previewing at https://delirious-bite.glitch.me/~fogcreek-dotcom ,

Display:
 - there's some alignment quirks to handle:
 image
 - I don't think using the loader graphic makes sense for this, as it's pretty normal for images to render in after-the-fact;  I think having the loader may call more attention to the area than is necessary
 - Intermixing the License icon with the node packages is a little weird, since they represent different things  (and also the license is missing its alt text)

content:
 - I'm not sure we want 'express' in there, since pretty much every project will have express.  I think it's more interesting to focus on the novel libraries--  this project is _stands out_ because I can see it's using _[tensorflow](https://www.npmjs.com/package/@tensorflow/tfjs),  etc.
 - check with @jennschiffer  for suggestions on other packages to get in the list.

UX treatment--
  After patching up the obvious quirks,  please sync with tiff on ideas around display, as well as a general checkin on "is this adding value to the page?  how to improve?", etc.

pketh:
@judeallred can you give additional context to the design goals for this? I think there are some assumptions being made about visual hierarchy/how this information is presented that could use an actual design pass.

src/presenters/pages/project.jsx
```diff
@@ -81,6 +81,90 @@ ReadmeLoader.propTypes = {
...
+    "underscore": "underscorejs.org",
+    "jquery": "jquery.com",
+    "aws-sdk": "aws.amazon.com/tools",
+    "passport": "www.passportjs.org",
+    "mongodb": "www.mongodb.com",
+    "express-handlebars": "handlebarsjs.com",
+    "googleapis": "developers.google.com",
+    "ghost": "ghost.org",
+    //"bluebird": "bluebirdjs.com", 
+    //"async": "caolan.github.io/async", these are both popular but don't work with favicon-fetcher
```
judeallred: worth sending along in a notion card to @GregWeil so he can fix up the fetcher

src/presenters/pages/project.jsx
```diff
@@ -81,6 +81,90 @@ ReadmeLoader.propTypes = {
...
+  ).map((dependency) => {
+    const hostUrl = popularDependenciesWebsites[dependency];
+    const imgUrl = "https://favicon-fetcher.glitch.me/img/" + hostUrl;
+    return {
+      name: dependency,
+      imgUrl,
+    };
+  });
+       
+  const licenseLogos = {
```
judeallred: I'm not convinced that added these to the page adds value,  since so many projects just take on a license by default.  Maybe bounce this around with Tiff;  consider cutting them if there's not a clear argument of why they're nice to show

src/presenters/pages/project.jsx
```diff
@@ -81,6 +81,90 @@ ReadmeLoader.propTypes = {
...
+    <React.Fragment>
+      <ul className="logos">
+        {popularMatches.map(({name, imgUrl}) => 
+          <li key={name}>
+            <img 
+              alt={name} 
+              title={name} 
+              src={imgUrl} 
+              className="logo" 
+              onError={(e) => {e.target.src = 'https://cdn.glitch.com/4d5adefc-986d-4406-8b38-898e95610bc7%2Fnpm.png?1538079784799'}}
```
judeallred: This is a good idea for when a single one of them is failing;  when _all_ of them are failing (say, because favicon-fetcher is down), we'd render a bunch of tiles all with this same image.  It might be that it's safer to render blank?  I'd workshop this one with tiff and figure it out

src/presenters/pages/project.jsx
```diff
@@ -81,6 +81,90 @@ ReadmeLoader.propTypes = {
   domain: PropTypes.string.isRequired,
 };
 
+const InterestingPackageJsonThings = ({data}) => {
```
judeallred: Throughout this function, we have to assume that 'data' is an unstructured blob.  It might not even exist or it could be null -- what happens when there is no package.json file?  --,  so the pattern to adopt is that for any aspect of the data we use,  we have to first check that it's present.

For example, before we can call Object.keys on data.dependencies, we need to assert that data.dependencies exists in the first place and is the type we expect,  so in this case something like `data.dependencies && typeof data.dependencies === 'object'` .  And for this, we'd have to have already asserted that 'data' exists :p

src/presenters/pages/project.jsx
```diff
@@ -81,6 +81,90 @@ ReadmeLoader.propTypes = {
...
+      <div>
+        <img className="license-logo" alt={data.license} src={currentProjectLicenseLogo}/>
+      </div>
+    </React.Fragment>
+  );
+};
+
+const PackageJsonError = (error) => (
+  (error && error.response && !error.response.status === 404)
+    ? <React.Fragment>We couldn't load the package file. Try refreshing?</React.Fragment>
```
judeallred: I'm not sure it's worth bothering the user about this;  there's already an error if they don't get the readme file.

judeallred:
@pketh the design pass is the next step in this;  allyson has put together a mockup and she's now moving on to working with tiff on seeing if there's a delightful thing here.  You're welcome to weigh in -- you haven't missed the boat :-)   But go to the notion card, i think, rather than this PR. 

The goal is experiment with if exposing the package information on the project page gives a richer experience.  The hypothesis is yes -- at a glance, it's handy to know what a project was built with!  But that's not universally true for all packages,  it's more that there are some interesting ones that we can call out.   There's a lot of interesting data in the package.json file that we can expose, so we're dipping our toe into that.

pketh:
the design makes some assumptions about how important and recognizable this information is by it's prominence on the page. If you're in the know, it's indeed prob handy info to have - but not on the level as a critical project identifier like name, avatar pic etc. 

I'd argue that this is information, secondary to the readme or embed, and should be treated as such. 

Additionally, there's an assumption that people are gonna know the icons for their fav dependency, I don't think that lines up for most of humans. The current design, sends the signal that you Need to know these things. Especially for licenses, text might make more sense. (especially considering how visually ugly some of the icons are, the effect can look like the intel inside stickers on a cheap laptop. 

It's cool that there's a design step next, and the notion card is the right place for that stuff. Cards for new features should include a *very rough Why we're doing this , and what the goals are. Explicitly stating this ensures we're thoughtful about what we add, saves time, and helps make the final design stronger.

src/presenters/pages/project.jsx
whimsicallyson: https://www.notion.so/glitch/87bda09f85184c1d81e1659aa62e4392?v=9f7963becc9341ed9af502682c6e9eef&p=06d17e2d1fa34c7fbda0737a90235be3

ttseng:
Spoke to @whimsicallyson about this - will put together a design spec with some more info about the feature along with layout options.  
Will get around to doing this next week after I finish up work with collections.

PR #118, opened 2018-10-02
Superb guide: Handle 404s when getting users/teams, proxy the teams page

GregWeil:
The api is going to start giving 404s when teams don't exist rather than returning `null`.

I also did some cleanup around the proxy so we can use Pirijian's new teams page. Proxied sites can exist at their root url now, but they'll have to be careful not to use any base relative urls.

### DIFF
--- a/server/proxy.js
+++ b/server/proxy.js
@@ -13,21 +13,8 @@ module.exports = function(app) {
   
   function proxyGlitch(route, target, pathOnTarget="") {
     const routeWithLeadingSlash = urlJoin("/", route);
-    app.use(routeWithLeadingSlash, proxy(target, {
-      preserveHostHdr: false, // glitch routes based on this, so we have to reset it
-      https: true,
-      proxyReqPathResolver: (req) => {
-        const path = urlJoin("/", pathOnTarget, url.parse(req.url).path);
-        console.log("Proxied:", path);
-        return path;
-      }
-    }));
-    routes.push(routeWithLeadingSlash);
-  }
-
-  function proxyGhost(route, glitchTarget, pathOnTarget='') {
-    const routeWithLeadingSlash = urlJoin("/", route);
     const sandwichedRoute = urlJoin("/", route, "/");
+    
     // node matches /{route} and /{route}/;
     // we need to force /{route}/ so that relative links in Ghost work. 
     app.all(routeWithLeadingSlash, (req, res, next) => {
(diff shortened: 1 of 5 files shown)


judeallred:
Prior to this change, only Ghost blogs has the "url must end in a '/'" constraint.  This is because Ghost loves relative urls so we need that slash to guard us,  however pages that aren't in ghost, such as https://glitch.com/forplatforms and https://glitch.com/teams  were permitted without the trailing slash.

With this change, the trailing slash is now mandatory on all the proxied routes, although not on glitch.com sites themselves.

Not sure if intentional or regression?

judeallred:
appropos of ^^;   the redirects (to the ghost sites) should hall have trailing slashes on them otherwise we get the double redirect effect

GregWeil:
Fixed a regression with /faq. (Server redirecting broke some of the hash url redirects, such as glitch.com/faq#gfp )
Also improved console logging when the paths don't match (it now prints the requested url, not where it got proxied to)
Also fixed that double redirect issue.

The / is required for the new /teams route, but I guess I could make it specific to that as well.

src/presenters/pages/join-team.jsx
```diff
@@ -19,7 +19,9 @@ class JoinTeamPageBase extends React.Component {
     try {
       var {data: team} = await this.props.api.get(`/teams/byUrl/${this.props.teamUrl}`);
     } catch (error) {
-      Raven.captureException(error);
+      if (error && !(error.response && error.response.status === 404)) {
+        Raven.captureException(error);
```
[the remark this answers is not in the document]
GregWeil: It happens automatically for any uncaught errors.

server/proxy.js
```diff
@@ -39,27 +25,40 @@ module.exports = function(app) {
...
+      }
+    }));
+    
+    routes.push(routeWithLeadingSlash);
+  }
+
+  function proxyGhost(route, glitchTarget, pathOnTarget="") {
     // Proxy all the requests to /{route}/ over to glitchTarget
-    proxyGlitch(route, glitchTarget, pathOnTarget);
+    proxyGlitch(route, glitchTarget, urlJoin("/", pathOnTarget, route));
```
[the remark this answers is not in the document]
judeallred: url-join doesn't normalize;  it'll put a slash on the end if the final string in the list ends with a slash

server/redirects.js
```diff
@@ -1,7 +1,7 @@
 module.exports = function(app) {
-  redirect(app, '/partners*', '/teams');
-  redirect(app, '/foryourapi*', '/teams');
-  redirect(app, '/forteams*', '/teams');
+  redirect(app, '/partners*', '/teams/');
+  redirect(app, '/foryourapi*', '/teams/');
+  redirect(app, '/forteams*', '/teams/');
```
pketh: will `glitch.com/teams` work?

server/redirects.js
```diff
@@ -1,7 +1,7 @@
...
   redirect(app, '/website-starter-kit*', '/featured/website-starter-kit');
```
judeallred: these are all double-redirects too

server/redirects.js
GregWeil: It would but we redirect /teams to /teams/ and it's faster to just skip the second redirect.

pketh: the second redirect is conditional though, so if `/teams/` is used it'll just redirect once, 

but if `/teams` is used (as a fallback, or mistakenly), then the second redirect happens (and the speed penalty is paid only then)?

GregWeil:
@garethx All of the proxied sites should work as is. /forplatforms is still loaded from about.glitch.me, and /faq does some special redirecting based on hash urls (`/faq#gfp` is the only one I know)

PR #119, opened 2018-10-03
Nova-chance: Remove dev toggles for team creation and deletion

GregWeil:
Are there any blockers here?


pketh:
pinging this again, what are the remaining blockers?

judeallred:
Blocker is that Anil wants to review the feature before we expose it in the header

judeallred:
code wise, we're GTG.  We have some priority improvements but nothing that in my opinion blocks futher rollout

pketh:
Would you like me to own getting Anil’s stamp on it?

judeallred:
We're meeting tomorrow morning for a demo, so I don't know that there's anything to own.  This dragged out because he was out of town for so long.

PR #124, opened 2018-10-08
Superb guide: Cleaner analytics prompt when no projects

GregWeil:
image

### DIFF
--- a/src/curated/sample-analytics.js
+++ b/src/curated/sample-analytics.js
@@ -1,0 +1,9 @@
+// This is shown when a team has no projects
+// It should match the format of https://api.glitch.com/analytics/:id/team
+
+export default {
+  buckets: [],
+  referrers: [],
+  remixReferrers: [],
+  remixers: [],
+};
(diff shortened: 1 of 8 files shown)


pketh:
arg i discussed this w jude but the case wasn't updated. the empty analytics serves a purpose - as a peek into what you will get, and the action to add projects is supposed to feel like a direct manipulation action on top of those graphs to make them have data. 

As an alternative to this, try opacity: 0.5 the graphs instead of removing them entirely. how does that feel?

in any case, pure 'cleanness' isn't necessarily the goal here. rather it's convincingness, without reading the rest of the page you should know at a glance what this is/could be, and be able to guess what you'll get if you add a project

pketh:
can you do this by passing a 'disabled' state (based on #of projects) down from the team page context?

judeallred:
nope

GregWeil:
image

Pushed in the label enough to not overlap the graph. Fading the graph just made it harder to see, it still looked broken.

pketh:
Looks good, nice job

pketh:
I like tiff's suggestion, lets do that (using hardcoded dummy data) if possible

PR #125, opened 2018-10-09
Profuse tea: Proper handling of anon users in the create team pop

GregWeil:
If you are signed out, the create team pop instead shows a sign in prompt

image

After you sign in it'll open the create team pop as if you were signed in to begin with

The anon user pop also has a sign in button in the teams section when the dev toggle is enabled

image

### DIFF
--- a/src/presenters/pages/login.jsx
+++ b/src/presenters/pages/login.jsx
@@ -46,7 +46,7 @@ class LoginPage extends React.Component {
   
   render() {
     if (this.state.done) {
-      return <Redirect to="/"/>;
+      return <Redirect to={this.props.target || '/'}/>;
     } else if (this.state.error) {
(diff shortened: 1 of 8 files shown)


pketh:
design recommendations for each of the above screens:

1. the title for a pop up generally only describes the pop-over, in order to help anchor the user in the navigation. eg: 

screen shot 2018-10-10 at 10 35 16 am

In this case the pop title should be 'sign in', and the "you'll need to sign up to create a team" conditional text should be an .action-description above the actual sign in buttons

2. The sign in text should be independent from the button and lead into the sign in button. For both consistency with places in the editor, but also so that you can go top-down without having to go back up. here are some examples of how we do this in the editor that would apply here

screen shot 2018-10-10 at 10 32 24 am
screen shot 2018-10-10 at 10 32 38 am

The last one is a good fit:
-----------
 *sign in* to create teams
[create team].disabled
----------

PR #126, opened 2018-10-09
Longing elf: Collections

ttseng:
http://longing-elf.glitch.me/

This is the in-progress implementation of Collections.

I'm still hooking up the API to add and remove projects from collections.

Here's a quick status on the various pages:
User Profile Page
- [x] Deleting collection
- [x] Update collection list after delete (needs to be done in entity-page-collections.jsx)
- [x] Create new collection with redirect to collection page
- [x] Add project to collection via project options pop

Collection Page
- [x] Edit collection name
- [x] Link collection name + handle
- [x] Edit collection description
- [x] Edit collection color
- [x] Edit collection avatar from preset list
- [x] Add project to collection
- [x] Remove project from collection
- [x] Authorization 

Project Page
- [x] Add project to collection

I also haven't started fixing all warnings yet.  

Will update the PR when additional features are complete.

I don't need a review ATM - just sharing here for documentation purposes.


pketh:
i've got some high and low level design feedback (ignoring the in-progress bugs):

high level:

collections page
homepage
screen shot 2018-10-10 at 10 57 56 am
- the collection view, homepage view, and user page view of collections all visually feel like distinct things. They share color, but besides that, the layouts (e.g. how things like metadata are displayed differently in each context) of each type doesn't evoke the other forms strongly enough. This make it feel like there's more complexity than there actually is. Ideally these should all feel like different sizes of the same thing (like a tshirt that comes in S, M, L)
- in terms of the one of the primary/overall goals of user-created collections adding a more organic community feel to the site, the use of presets (over random generation and adjustments for example) does the opposite. At a glance, presets make all collections feel the same.
- the randomly generated names for collections are the same as projects. Just like how team names use slightly distinct randomization, I think collections could benefit from this as well. 
- in terms of overall user page placement, the collections are up first, and also because of their color and large size, dominate over the projects. The downside of this is that the pages now prioritizing curation over creation. This feels like the opposite order of what we want to encourage users to value (creation, pride of creation > curation, pride of curation).

details:

should not be presets
color is out of context should not be presets
screen shot 2018-10-10 at 10 57 56 am
- ^ re presets comment above, these illustrations aren't meant to be used outside of their current collection context or outside of their specific collection color. These presets allow for unattractive combinations
- instead of being a blank box, the collection color should display the active hex code, and allowing live editing. 
- there also doesn't appear to be an active state on the color. You can intuit what color is selected from the surrounding context, but in terms of feel/direct manipulation, having both would feel better

screen shot 2018-10-10 at 10 42 30 am
- typically, we use menlo for text that is directly user editable (ie text in the editor, or your user name). Because this text is non editable it shouldn't be used here. Additionally, there may be a more visually intuitive way to communicate emptyness (or size in general)

screen shot 2018-10-10 at 10 59 15 am
- avatar sizes, and positioning are inconsistent

I can help you out with these if you need it, just lmk

pketh:
additional quirks i forgot to mention
- because of how 'add X' works in other places (like adding a project to a team), it feels awkward and juddering when the whole page changes in response to clicking 'add collection'. Based on ^, my expectation is that some kind of entity gets added to the page
- using it on the user page, it seems like functionality wise, pinned projects and collections share core functionality but work completely differently. Having to learn two separate flows for adding things for example feels awk as well

ttseng:
@pketh please enumerate your feedback rather than use bullets as it makes it difficult to respond to your comments otherwise.

I'll attempt to enumerate your points in my responses.

1. Re: consistency of collection card on homepage & user page, and collection page layout - will consider whether I can modify the card on the homepage to look more like the newest collection page layout, and to use a similar grid vs list structure for collection card on user page.
2. Auto-generated names for collections are consistent with how team names are created.  If we want to consider making auto-generated names draw from a different list of names for each type of object, we should consider changing team names too.
3. Can consider different auto-generated background color in next version of collections.  I don't consider it a blocker for V1.  Randomly generated background colors from our palette makes it at least feel consistent with the rest of the site.  We don't have completely personalized backgrounds for everything - for example, user profiles all have the same default pixel grid background.  I would argue we should change that too if we want everything to feel customized by default.
4. Collections placed at top of user page helps promote the new feature, at least at initial introduction.  Soon after we will enable user to reorganize sections on their profile page and also set a "see more" cut off so that users don't have to scroll far down the page to see different sections if the user has a disproportionate # of collections or projects on their profile page.
5. Agree about adding active state to collection colors pop-over - that was part of the initial design but just hasn't been implemented yet.  Will be there for V1.
6. Changed placeholder font from Menlo to Benton-Sans.  Note that Menlo was initially selected to match the help text shown in the editor when it's loading, which is not editable. 
7. Fixed default avatar size.
8. I chose to redirect the user to the collection page on clicking "Create Collection" rather than create an empty state on their profile page, because it reduces the # of clicks to adding a project to collection from 3 clicks (1. click create collection, 2. click on empty collection created on profile page to be redirected to collection page, 3. click on add project button on collection page) to 2 (1. click create collection and auto-redirect to empty collection page, 2. click add project button on collection page). I still think this is preferable.
9. Users can add projects from their profile page to a collection through the project dropdown now, so pinning and adding project to the collection workflow is more similar.

ttseng:
An update from the checklist above:

User Profile Page
- [x] Deleting collection
- [ ] Update collection list after delete (needs to be done in entity-page-collections.jsx)
- [x] Create new collection with redirect to collection page
- [x ] Add project to collection via project options pop

Collection Page
- [x] Edit collection name
- [ ] Link collection name + handle
- [x] Edit collection description
- [x] Edit collection color
- [x] Edit collection avatar from preset list
- [x] Add project to collection
- [x] Remove project from collection
- [x] Authorization 

Project Page
- [x] Add project to collection

Remaining tasks are being documented here: https://www.notion.so/glitch/54e8664170cb4ef2849615bcc23b5f90?v=65bc39c2288248bd8ab1a2ed7096172f

pketh:
2. I just created a collection on longing-elf and got back `fancy-knave`. This is using the name generation path as projects. Teams have a distinct name generation path where you'll see that the second word is a synonym for team (eg 'collective'). Based on this, collection names are currently consistent with how project names work. But, just like how teams and projects are different, collections may benefit from the added context or whimsy possibilities that collection specific randomization allows

3. I agree we should randomly generate default user backgrounds and that's something we should revisit as well. Consistency of user generated content with content that we (glitch) creates is an anti goal of community. Your content should feel like your content, that's what helps you feel ownership of it. 

Also, the current scheme of random color presets + canned illustrations that were not designed for other background colors allows for visually unattractive combinations.

4. new feature promotion shouldn't trump the core job of a page and the primary reason people are using it. (We have lots of other ways to promote new features, marketing, new-stuff-dog, homepage featured collections) Our design decisions should be aligned with users first. I think the 'cut off' idea on projects makes sense. Especially after I add a collection, the physical and attention space they take gets in the way of the primary use case of going to my own user page to get back into my projects, or a visiting another user's page to see what projects i've created.

ttseng:
2. spoke with @whimsicallyson who is going to add some custom words for collections to friendly-words.
3. I can see if I can change all the existing "canned" art to at least line-work with no fills, so the color combinations aren't as jarring.  or, as an alternative, we can simply have some stock artwork and not allow users to change the avatar at all, for example, the Glitch Fish in a picture frame or something like that.  A last alternative is that we can omit avatars altogether for this first version.
4. I think some people will enjoy curating collections rather than creating their own content (for example, people who create playlists on YouTube who never update their own videos).  So I wouldn't necessarily think that getting to your own projects is the #1 priority for your profile page.  We can keep collections on top at launch and see what patterns of use emerge as we decide how to let users prioritize different types of content on their profile.

pketh:
re 3. would collections work without avatars at all (at least in this version)? If the avatars are fixed then they don't really convey anything right?

pketh:
re 4. for sure some people will be primarily curators, but just like on spotify/youtube, the vast majority of people will be there to consume/create (the latter is where the most value is for us and them). Similarly, if I'm visiting user X's profile-page/artist-page/channel, I might be curious about what they've curated/liked - but I'm way more interested in the content that they've made

ttseng:
3. I'm leaning towards black and white avatars, with some standard defaults (game, laptop, turntable, art) - the rest are kind of random and should probably be omitted. they are still useful signifiers of content, and help us visualize what the thumbnails will look like when users are able to upload they own.
4. it doesn't make sense to put collections under recent projects at this point, as we have no way to hide overflow of projects and some users with lots of projects would never see collections if it's underneath their recent projects.  the only other place where collections might make sense is underneath pinned projects.  the reason I would prefer collections above pinned projects is because there is a more logical grouping of pinned projects > recent projects than pinned projects > collections > recent projects

pketh:
to help convey my visual/size comments from above: 
screen shot 2018-10-16 at 4 12 52 pm

jennschiffer:
> to help convey my visual/size comments from above:
> screen shot 2018-10-16 at 4 12 52 pm
> 
> (i pushed up the sketch file to the design repo, it's in `/archived-and-expeiments/collections-sample.sketch`)

one concern i have with the horizontal scrolling in the above visual is that it makes the keyboard and non-traditional input peripheral experience more difficult to manage - especially if they have the focus on the projects and they have to tab through a bunch of projects. this is important to keep in mind if this become the design that gets implemented.

i also think the implication of more projects with a horizontal scroll looks like a css bug. scroll bars look different in literally every browser too, and i think it would be hard to manage the look and feel because of that

pketh:
re jenn's feedback , these are pretty easy things to address

screen shot 2018-10-17 at 11 53 47 am

ttseng:
Re: cards on profile page, I move towards this design (which actually ends up being pretty similar to @pketh's second suggestion above - I hadn't gotten around to reviewing the comments before I implemented it)

image

I prefer the vertical arrangement over the horizontal because it more closely matches the style of the collection page on mobile (all projects listed in a single column).

I also intentionally left the project names un-bolded, because I'm trying to avoid having the listed projects look anything like buttons.  This is also why I left View X projects -> as text, rather than turn it into a button.  Containers that are links should never have more than one item that appears as a button - otherwise, they compete for visual hierarchy and make it potentially confusing to know where to click.

Re: avatars - I tried changing the avatars to greyscale, but it didn't feel right because the rest of the app is colorful, and having black-and-white avatars seemed jarring with the rest of the site.  
But now I'm concerned that having the default avatars available for any user collection will dilute the value of / lead to confusion about the Glitch-curated categories, so I no longer think we should use them.  Those avatars should be reserved only for Glitch-curated stuff, I think.

@pketh do you want to think about what auto-generated collection avatars might look like?

pketh:
avatars should be more like cover art than icons, they evoke a vibe more than a function. Also it's good for randomly generated avatars to be visually distinct between entity types (ie projects look different from collections). As for examples of how that could play out, in the original spec I threw out 
screenshot2018-08-22at10 51 22am
based on drawing a random set of circles of random sizes and colors. 

But you could also go in a more constrained direction with randomized prebaked shapes, for something that's got a sorta spotify playlist vibe by creating canvas shapes and generating parameters like placement, orientation, color and size. (for making shapes in a gui and getting back canvas-y code, https://www.paintcodeapp.com is helpful)

for example, 
raw shapes:
screen shot 2018-10-18 at 11 33 32 am

can recombined into an avatar:
screen shot 2018-10-18 at 11 38 16 am

the way i built this up 
1.  have a background layer for big shapes and keep adding until x% of the canvas is covered
2. have a mid layer for medium shapes until y%
3. etc for as as many layers needed

it'll be playing around w what shapes work best in a random context (hard/non-organic lines might not work well), and the various percents. With `randomcolor` you could also do stuff like ensure background or foreground random colors are light, normal or dark hues

sketch file:
[collection-avatar.zip](https://github.com/FogCreek/Glitch-Community/files/2492570/collection-avatar.zip)

There have also been a couple glitch projects that randomly generate art, some of those might be good inspiration. codepen also has lots of good stuff

I think the background color being colored is a big part of what makes them feel so much more 'weighty', shouty and attention grabbing from the other items on a user page. On the homepage, doing it that way originally made sense because in that context they should be shouty because we had nothing else. But with the extra content marketing stuff, and the quieter visual context a colored background isn't necessary in both places. Also only using color for the projects, means the collection avatar will always be on a neutral background and the collection page itself will be able to better draw your eye to the colored content (the projects)

The other thing that'll make this fit in better is to trim out as much vertical space as you can, imply size/depth with things like shadow, and using 6px padding instead of the standard 12px in places where that makes sense to visually communicate the idea of 'little things inside of a big thing'.

pketh:
re random avatars: here are some other things to start getting ideas from

https://www.are.na/block/1634226
https://open.spotify.com/album/72rhxNzk3Lm3I6zT1YqYQx
https://www.are.na/block/2480352
https://www.are.na/block/2490326
https://www.are.na/block/791296
https://www.are.na/block/2711184
https://www.are.na/block/2210050

ttseng:
thanks for the links. I'm not going to dive into this right away - there
are some more fundamental parts that need to be fully implemented first.
 afterward, I'll give this some more thought.

tiff

ttseng:
@judeallred @GregWeil @pketh 
I just pushed the latest code and believe this branch is now ready for code review.

pketh:
@ttseng responded to design comments on https://www.notion.so/glitch/collections-visual-feedback-oct-31-a746eef85afa427bb4e58e784257bd2c

ttseng:
closing this PR and opening new one specifically for code review.  design feedback will be resolved outside of Github.

PR #129, opened 2018-10-16
Superb guide: Show zine posts on the homepage

GregWeil:
image

Also redirects all /featured/* urls to /culture/*.

Zines are loaded on the server every 15 minutes and sent down with the html, both to avoid cors and to reduce the load on the ghost api.

I also did some refactoring around the server side data loading to make the error handling a bit more specific, as well as in response rendering so the manifests are read asynchronously.


judeallred:
during page load, there's a weird transition point as the images come in:
image

judeallred:
styles are busted in Edge :/ 
image

judeallred:
i find the faded to white effect on the website starter kit to be pretty hard on my eyes -- my personal pref would be to up the contrast there or avoid a fade of grey to white: 
image

judeallred:
when i navigate to any of the zines,  the image loads large then flashes out of view -- @pketh ?

server/api.js
```diff
@@ -35,24 +33,54 @@ const api = axios.create({
...
+    if (error.response && error.response.status === 404) {
+      return null;
+    }
+    throw error;
+  }
+}
+
+async function getCultureZinePosts() {
+  console.log('Fetching culture zine posts');
+  const client = 'client_id=ghost-frontend&client_secret=c9a97f14ced8';
```
judeallred: how does this secret work?  Do we need to guard in it env?

src/presenters/featured.jsx
```diff
@@ -1,14 +1,43 @@
...
+class ZineItems extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      posts: window.ZINE_POSTS.slice(0, 4),
+      masks: sampleSize([1, 2, 3, 4, 5], 4),
+    };
+  }
+  render() {
+    return (
```
judeallred: what's the behavior when the culture-zine app is down (so our cache is empty) ?

server/api.js
GregWeil: https://api.ghost.org/docs/client-authentication

should be public safe, it can only read public data

judeallred:
overall lgtm but I think we should fix up those little bugs.  (no blockers.) :-)

src/presenters/featured.jsx
GregWeil: You get an empty `<ul>` which doesn't affect the page

GregWeil:
https://developer.microsoft.com/en-us/microsoft-edge/platform/status/masks/
Sounds like masks should be supported in Edge 17763 but I'm still running 17134. Maybe it just hasn't been released yet?

pketh:
my feedback:

the hover state should just be a black underline on the post name, and no underline on the category name

re jude's feedback:

re busted styles on edge, hmm https://caniuse.com/#feat=css-masks looks like mask support is coming in the next version (18). In the meantime, adding border-radius: 5px to the images would make it less broken for those users. Also a fixed height (background-size: contain).

re faded to white, the issue there is with the content. We'll have more design guidelines for featured post images to prevent this issue on future posts

re loading the zine,
what happens is that a default mask is applied, then some client side js runs to determine what random mask should show up . this would change the mask. To deal with this:
- I can either have the mask at the top always be fixed though (css has no dynamic random number generation that i'm aware of)
- Or I can not have a mask on load, and only have a mask after the client side js runs (so it'd start square, then be squiggled)

pketh:
how does the hover and active states feel to you generally, i'm not 100% feeling them yet, any suggestions/ideas?

GregWeil:
Updated with rounded corners for edge, fixed size images (for the underline issue), and title only underline.

I'll update this after lunch with node 10 `fs.promises.readFile` instead of `util.promisify`.

pketh:
can you send me an invite link, just in case i need to do little nip tuck things to the design soon too?

GregWeil:
You can join in with the performance boost team.

src/presenters/featured.jsx
```diff
@@ -1,14 +1,43 @@
...
-const Witch = () => (
-  <img className="witch" src={imgWitch} width="110px" height="82px" alt=""/>
-);
+
+class ZineItems extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      posts: window.ZINE_POSTS.slice(0, 4),
+      masks: sampleSize([1, 2, 3, 4, 5], 4),
```
pketh: this is a clever way to do set random numbers, i'm gonna have to remember that :)

styles/includes/featured.styl
```diff
@@ -9,43 +9,48 @@
...
+  .community-pick-embed-container
+    position: relative
+    .witch
+      position:absolute
+      z-index: -1
+      right: 0px
+      top: -72px
+      
+    .glitch-embed-wrap
+      //max-height: 500px !important //override the Embed's configurable height.
```
pketh: is this supposed to be commented out?

pketh:
made some minor proportions tweaks. Visually I'm now treating each each type of featured content as it's own section (where before all curated content was squished in to the same section). Conceptually, a viewer shouldn't have to think about it too much, these things should just come off as cool-content to a viewer.

PR #128, opened 2018-10-16
Ordinary emu -- touch up 'show more' button styling on readme files

judeallred:
https://www.notion.so/glitch/87bda09f85184c1d81e1659aa62e4392?v=9f7963becc9341ed9af502682c6e9eef&p=44f17f21280442dca6c6a39949f4448b

before:
image

before (on hover):
image

after:
image

after (on hover):
image

PR #131, opened 2018-10-18
Coffee decimal -- in project list sections, only display the first twelve projects

judeallred:
https://www.notion.so/glitch/87bda09f85184c1d81e1659aa62e4392?v=9f7963becc9341ed9af502682c6e9eef&p=bf4bb5850c5241b8b886aaf8f197465f

On the user and team pages, the 'pinned projects' and 'recent projects' sections now stop displaying after 12 tiles, and then give a 'show {N} more' button, which will then display the rest.

The button uses the same styles as the existing 'show deleted' button in the deleted projects section. 

For users with lots of projects, this is a big improvement to page load time. 

good spots to view:

https://coffee-decimal.glitch.me/@Performance-Boost
image

https://coffee-decimal.glitch.me/@judeallred
image


judeallred:
@ttseng please review for UX feedback when you get a chance :-)

pketh:
visually, this might benefit from the same fade treatment as the project page readme.

When you hit command/ctrl+f to search for a particular project name, it'd be handy if this auto expanded so I could do my search against the entire project list. (same thing for the project page I guess too)

pketh:
I don't know whether the count is necessary on the button, it feels like more work to think about than simply 'show all projects'. Especially if it's a large number, it feels like work/anxiety-causing to click because now I'm thinking about the 55 things i'll have to parse.

ttseng:
cool, this is really helpful to have.

I agree with having a mask that covers up part of the next row of projects, to visually hint that the project overflow is being hidden.

I like having the number count - it lets you know just how many projects to expect.

I don't think that having it auto-expand on search is necessary at this point, especially since I heard @jennschiffer is working on a search field for the user profile page that will automatically filter out search results.

jennschiffer:
i'm only just about finished on that prototype, so i don't want decisions made here based on that feature since there hasn't been any discussion on if we even will add it.

that being said, cmd-f searching implies that you're searching what the browser is already showing, so i don't think it should add more content before that search.

jennschiffer:
that would say to me that this means the "show more" isn't obvious enough so you know going to the page that not all of the projects are listed. 

I'm thinking of other sites like github that have "show more" and don't expand on browser search keywords. maybe we should shorten the number of projects we show initially rather than change the expected behavior.

pketh:
in the above scenario i haven't even gotten to 'show more' so it's not a matter of the obviousness of the button. It's one of those page load and immediately cmd-f cuz I know what I want. In the gh case, I suspect they don't show all for performance reasons, we've got around much of that tho. 

What's the case against expanding when you hit cmd-f?

jennschiffer:
we'd be having a known shortcut do something that isn't typical of what it does outside of our site, and introducing the complexity of managing cross-os shortcuts (cmd-f, ctrl-f, windows+f) - although i recognize that complexity may be more trivial if we manage it elsewhere on the site already. i'm also wondering if hitting cmd-f again would then bring the project list back to 12?

pketh:
We wouldn't be overriding the browser default, the normal search would still function as normal. we're just doing an extra thing once triggered. 

My assumption would be that this is a one way thing - just like how in the project readme, we don't let you recollapse things after you've revealed them (that'd just be finnicky). 

Also we already handle multi-platform keyboard shortcuts in the editor, that part is pretty straightforward

pketh:
on the project page the fade out effect is incorrectly implemented
screen shot 2018-10-25 at 10 28 07 am

to fade from transparent at the top to white at the bottom you'll want to update `.expander-mask` to use 

`background: linear-gradient(-180deg, rgba(255,255,255,0.00) 0%, #FFFFFF 100%);`

additionally, instead of using a bespoke button type (`expander-button`), you can remove that and simply use just `.button-small`

screen shot 2018-10-25 at 10 32 04 am

judeallred:
recapping above--
  i'll add a mask (/see if i can reuse the other one) to fade the bottom projects,  and look at the button styling.

seems like we have two inconsistent button styles atm; the style we use for showing deleted projects (matches this style) and the style we use for the readme show more.

judeallred:
a quirk about using the faded expander for the bottom of the projects list is that its click area is going to overlap the click area of the project tyles

judeallred:
w/r/t the gradient, @pketh , what browser/os you in?  it looks way better on my end:
 
image

(windows / all browsers)

w/r/t the button-small -- it's missing its shadow effect in this context,  so there's a little more to it than that to get it working,  though i agree we want to be more standard with it.  i'll play around in this pr.

judeallred:
patched up the show more button for readme files per ^^ and https://www.notion.so/glitch/bug-style-touchups-on-show-more-section-on-project-page-44f17f21280442dca6c6a39949f4448b and https://github.com/FogCreek/Glitch-Community/pull/128 

not hovering:
image

hovering:
image


not changing the gradient since i don't have a repro of what pirijan is commenting on; will update if that evolves

judeallred:
ok,  back to the actual project list....

pketh:
i'm in safari mac, this is also an issue on ios

pketh:
i see `button button-small` used for both projects and deleted projects at https://glitch.com/@pirijan

judeallred:
Doing the fade is kind of tricky, since it needs to work both with the 3-pane view and the 1-pane view.

I might also be tempted to do something like this -- add a placeholder project tile for the button:
image

This has the advantage of resizing nicely as the page layout changes.  The one pictured here doesn't feel quite right yet, but it conveys the idea.

pketh:
by 3-pane-view and 1-pane-view are you talking about view port sizes?

pketh:
if so, does the fade work less badly here than the readme on the project page?

judeallred:
I'm going to go ahead and get this version out there (since the UI is consistent with the other 'show' UI) and i've made this notion card to discuss UI design for the show more display:
  https://www.notion.so/glitch/87bda09f85184c1d81e1659aa62e4392?v=9f7963becc9341ed9af502682c6e9eef&p=56656aed574a4909baaabf007aa2488f

Send further design discussion to comments on that card, please :-)

src/presenters/deleted-projects.jsx
```diff
@@ -78,7 +86,12 @@ export default class DeletedProjects extends React.Component {
     } else if (!this.props.deletedProjects.length) {
       return 'nothing found';
     }
-    return <DeletedProjectsList {...this.props}/>;
+    return (
+      <React.Fragment>
+        <DeletedProjectsList {...this.props}/>
+        <button className="button button-tertiary" onClick={this.clickHide}>Hide Deleted Projects</button>
```
judeallred: I added a hide button for the deleted projects;  it was rather conspicuously missing.   We may want a similar one for the 'show projects' button too, but I'll leave that to the forked notion card.

judeallred:
@GregWeil code review please :-)

judeallred:
@pketh  ok i'll move the gradient thing out to https://www.notion.so/glitch/87bda09f85184c1d81e1659aa62e4392?v=9f7963becc9341ed9af502682c6e9eef&p=75b11c58c0f44e6eb63fac6c44afdbbe  --  i can't reproduce it, so i can't really fix it

judeallred:
got a repro, will fix it here....

judeallred:
gradient bug fixed :-)   It must've been like that for ages.  Fortunately it looked ok on all non-ios devices.

PR #132, opened 2018-10-18
Profuse tea: A couple small bug fixes

GregWeil:
1. Some just barely supported browsers don't support the `scrollTo` object syntax
2. Hide the remove from team button if you're the only member of the team
3. The analytics project switcher would show user names when you hovered over projects


GregWeil:
Fixed reflow when the delete team pop is opened

PR #134, opened 2018-10-25
Quixotic rocket: Update react to 16.6.0, replace uglify with terser

GregWeil:
https://github.com/webpack-contrib/uglifyjs-webpack-plugin/releases#2.0.0
Adds `React.memo()` which is functional `PureComponent`
There's also a way to get a single context value as a class property, which is nice but seems limiting

https://github.com/webpack-contrib/uglifyjs-webpack-plugin/releases#2.0.0
Uglify stopped supporting es6, which meant we couldn't use the latest version
They recommended terser though, so switch to the latest version of that


judeallred:
Sweet, thanks. Just making sure -- you verified that the bundle looks ok in 'production' builds as well as dev builds?

PR #135, opened 2018-10-26
Superb guide: Don't auto expand readmes

GregWeil:
https://superb-guide.glitch.me/~three-js

I got inspired for this based on the only show x projects pr. Readmes used to check their height on load and if they were short enough they'd snap to the expanded state.

That had the side effect of sometimes expanding readmes that were too large, because they images that didn't load until after the height check. The new solution is to track the scrollHeight and show the fade/button if the readme is actually being cut off.

There is no way to actually listen for scrollHeight changes, so instead I'm just watching for onLoad events and updating whenever one comes in. (capturing because onLoad doesn't bubble)

### DIFF
--- a/src/presenters/includes/expander.jsx
+++ b/src/presenters/includes/expander.jsx
@@ -5,6 +5,7 @@ export default class Expander extends React.Component {
   constructor(props) {
     super(props);
     this.state = {
+      expanding: false,
       expanded: false,
       scrollHeight: Infinity,
     };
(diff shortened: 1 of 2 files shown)


GregWeil:
Some updates:
- Added a listener for window resizing which was completely missing
- Brought back some functionality lost in this pr. If the box is just a few pixels over the limit it lowers the max height so there's a more visible change on expand

PR #137, opened 2018-10-26
Leeward aftermath -- resilience against if Raven is blocked from loading

judeallred:
https://www.notion.so/glitch/87bda09f85184c1d81e1659aa62e4392?v=9f7963becc9341ed9af502682c6e9eef&p=45c8ae5984ef44cdbded976b46578cad

This PR allows the site to still run if Raven fails to load.

Rather than requiring us to check window.Raven before using it each time, though, I'd rather build a wrapper utility so that it falls back to no-ops.  I'll play with that a bit.


judeallred:
I feel like a stub JS library for raven is something that should already exist and be officially supported (just like google analytics bootstraps itself with a function stub).  I'll contact sentry and see what they have...

in the meantime I'll stick with this approach.

judeallred:
aha!  they deprecated the library we're using -- https://www.npmjs.com/package/@sentry/browser is the new onw

judeallred:
Updated the client to use the new Sentry SDK: https://docs.sentry.io/quickstart?platform=browser

judeallred:
we should also update the server to use the new sdk, but that can be a separate PR -- https://www.notion.so/glitch/87bda09f85184c1d81e1659aa62e4392?v=9f7963becc9341ed9af502682c6e9eef&p=85a095a309ae449abb1b8b72fa21f715

judeallred:
One quirk of this implementation is that it doesn't guarantee that Sentry is initialized before the rest of the codebase loads, which causes any Sentry calls in the top level of files to silently fail.  I think I can fix this by putting a little wrapper around Sentry, I'll give that a shot.

judeallred:
wrapping it up fixed that work.
@GregWeil Ready for your review :-)

src/presenters/current-user.jsx
```diff
@@ -28,19 +29,17 @@ function identifyUser(user) {
...
-      if (user) {
-        Raven.setUserContext({
+    if (user) {
+      configureScope((scope) => {
+        scope.setUser({
           id: user.id,
           login: user.login,
         });
-      } else {
-        Raven.setUserContext();
```
GregWeil: This lost the ability to clear the user on sign out. Was that intentional?

GregWeil:
Does this still catch all of the bootstrap errors? It looks like it won't get parser errors at the least, but would probably get errors in client.jsx 's tests before setting window.bootstrap

src/presenters/current-user.jsx
judeallred: ah, it was not!  I wasn't clear on the intent of that, thanks! fixin....

judeallred:
I don't think it will, since we're not loading it from a CDN anymore. Any suggestions on ways to continue capturing those without CDNning-it ?

src/presenters/current-user.jsx
judeallred: fixed

GregWeil:
I don't think it's possible as long as it's being bundled with everything else. We could probably add an extra entry point for it, but I we'd probably have to go back to a `window.Sentry` type usage

judeallred:
Agreed.  I think this technique is our better bet, if it's ok by you

GregWeil:
That's fine by me, maybe we should move the sentry init into a func that gets called by bootstrap so there's a clear start to error tracking, without implying that top level stuff gets checked

judeallred:
good idea, will fix

judeallred:
Mmm, on second thought I don't think we can do that without losing our ability to capture errors at the top scope.  Maybe I'm missing something, but it seems like if we do the sentry.init() anywhere except for in the top level of sentry.io, then it means that calls to sentry (sendmessage, etc.) will fail depending on the load order.

src/client.jsx
```diff
@@ -41,15 +17,21 @@ const func = (f, ...args) => f(...args); // Can we define arrow functions?
...
     window.location.replace(EDITOR_URL + window.location.hash);
-  } else {
-    const dom = document.createElement('div');
-    document.body.appendChild(dom);
-    render(<App/>, dom);
+    return;
   }
+  
+  configureScope((scope) => {
+    scope.setTag("bootstrap", "true");
```
GregWeil: This looks left over, by the time it runs it's putting the flag on actual bugs rather than syntax/compat errors.

GregWeil:
That makes sense. We can't guarantee that  errors are caught at the top level, but if you call a sentry func it will work.

PR #138, opened 2018-10-29
Profuse tea: stricter email identification, more 404 handling, error boundaries

GregWeil:
- When adding a user to a team, don't consider the string to be an email domain unless it contains a . (`@greg` is not a domain, but `@gre.g` is)
- The create team pop didn't handle 404 errors from the api, now it does
- Added an `ErrorBoundary` component, which catches errors during rendering (not in event handlers or async stuff, only during rendering) and replaces its children with the standard 'something went wrong' text. It's used on the questions list (since it has had issue with bad data in the past) the page within `Layout`, and around the entire react tree. This should be enough such that users are never left with a completely blank page, and in most cases will still have a functional header

### DIFF
--- a/src/client.jsx
+++ b/src/client.jsx
@@ -5,6 +5,7 @@ import React from 'react';
 import {render} from 'react-dom';
 import {BrowserRouter} from 'react-router-dom';
 
+import ErrorBoundary from './presenters/includes/error-boundary.jsx';
 import {CurrentUserProvider} from './presenters/current-user.jsx';
 import {UserPrefsProvider} from './presenters/includes/user-prefs.jsx';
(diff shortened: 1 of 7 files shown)


GregWeil:
The easiest way to test without making a remix is probably to use the react debugger to inject bad state/props

judeallred:
What's your assessment of ErrorBoundaries?  Were you able to break them with asynchrony, or do they seem fairly reliable?

GregWeil:
Error boundaries don't catch async errors at all, but async errors also don't cause the page to white out (that only happens when there's an error in a render function or a constructor) I wasn't able to get the page to white out like this, though

judeallred:
nice

PR #141, opened 2018-10-31
Quixotic rocket: Allow project page embed to request permissions

GregWeil:
Allow project page embeds to request permissions. This isn't enough on its own, because the embed has an iframe in it to load the preview, which will also need the same permission set.

https://www.notion.so/glitch/Embeds-iframe-element-needs-allow-attribute-so-getusermedia-geolocation-and-other-web-apis-in-apps-work-130ffb626a2d4c50b8987f444be1cbc4

### DIFF
--- a/src/presenters/pages/project.jsx
+++ b/src/presenters/pages/project.jsx
@@ -60,7 +60,7 @@ PrivateToggle.propTypes = {
 
 const Embed = ({domain}) => (
   <div className="glitch-embed-wrap">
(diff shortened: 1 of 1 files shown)


judeallred:
curiously I don't see the change being reflected in the dom:
image

GregWeil:
This is only for project pages. The homepage embed is a direct copy from the editor, so it'll change when it gets swapped out.

PR #142, opened 2018-11-01
Lovely edger: Placeholder avatars

whimsicallyson:
Placeholder avatars after you've invited someone to a team (https://www.notion.so/glitch/Show-placeholder-avatars-for-people-after-inviting-them-88fdccd9e24e4172ae9d261e83e6dc2c)

### DIFF
--- a/src/presenters/includes/team-users.jsx
+++ b/src/presenters/includes/team-users.jsx
@@ -122,11 +122,13 @@ export class AddTeamUser extends React.Component {
         {({visible, togglePopover}) => (
           <span className="add-user-container">
             <button onClick={togglePopover} className="button button-small button-tertiary add-user">Add</button>
-            <UsersList users={this.state.alreadyInvited}/>
             {!!this.state.invitee &&
+              <React.Fragment>
               <div className="notification notifySuccess inline-notification" onAnimationEnd={this.removeNotifyInvited}>
                 Invited {this.state.invitee}
               </div>
+              <UsersList users={this.state.alreadyInvited}/>
+              </React.Fragment>
             }
             {visible && 
               <AddTeamUserPop 
--- a/src/presenters/includes/team-users.jsx
+++ b/src/presenters/includes/team-users.jsx
@@ -1,7 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
-import {getDisplayName, getAvatarThumbnailUrl} from '../../models/user';
+import {getDisplayName} from '../../models/user';
 import {WhitelistedDomainIcon} from './team-elements.jsx';
 import AddTeamUserPop from '../pop-overs/add-team-user-pop.jsx';
 import PopoverContainer from '../pop-overs/popover-container.jsx';
@@ -94,7 +94,6 @@ export class AddTeamUser extends React.Component {
   
   async inviteUser(togglePopover, user) {
     togglePopover();
-    console.log(getAvatarThumbnailUrl(user));
     this.setState({
       invitee: getDisplayName(user),
       alreadyInvited: [...this.state.alreadyInvited, user],
--- a/src/presenters/includes/team-users.jsx
+++ b/src/presenters/includes/team-users.jsx
@@ -96,7 +96,7 @@ export class AddTeamUser extends React.Component {
     togglePopover();
     this.setState((state) => ({
       invitee: getDisplayName(user),
-      alreadyInvited: [...this.state.alreadyInvited, user],
+      alreadyInvited: [...state, user],
     }));
     await this.props.inviteUser(user);
   }
--- a/src/presenters/includes/team-users.jsx
+++ b/src/presenters/includes/team-users.jsx
@@ -103,9 +103,10 @@ export class AddTeamUser extends React.Component {
   
   async inviteEmail(togglePopover, email) {
     togglePopover();
-    this.setState({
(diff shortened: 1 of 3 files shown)


src/presenters/includes/team-users.jsx
```diff
@@ -94,6 +96,7 @@ export class AddTeamUser extends React.Component {
     togglePopover();
     this.setState({
       invitee: getDisplayName(user),
+      alreadyInvited: [...this.state.alreadyInvited, user],
     });
```
judeallred: When you need to reference the current state to set the new state, we get to use the alternate setState syntax so that it can handle asynchrony -- see https://reactjs.org/docs/state-and-lifecycle.html#state-updates-may-be-asynchronous

src/presenters/includes/team-users.jsx
```diff
@@ -124,6 +127,9 @@ export class AddTeamUser extends React.Component {
                 Invited {this.state.invitee}
               </div>
             }
+            {!!this.state.alreadyInvited &&
```
judeallred: Empty arrays are truthy-- need to use this.state.alreadyInvited.length here if you want to prevent rendering the UsersList when the array is empty

whimsicallyson:
> can you add a screenshot here or in the notion card so i can get a quick glance at what this looks like?

screen shot 2018-11-02 at 9 43 50 am

pketh:
In other systems with ‘shadow’ users like trello, there isn’t a strong top level distinction in position from users invited and users who’ve joined. The reason there’s no distinction is because if I’m just using my team, i can usually consider invited users already as members of the team. 

If you want to convey that user hasn’t accepted the invite yet instead of repositioning, it might be more low key to either day that in the person-pop, or display them as transparent (to hint at the ghosty ness of them)

pketh:
If you’d like I can move the above comment to the notion card instead, just lmk

src/presenters/includes/team-users.jsx
whimsicallyson: UsersList is just an empty div, so its not problematic to have it rendered, but I agree its cleaner to not.

whimsicallyson:
@pketh I ghosted the avatars to 50% opacity, I think that helps

src/presenters/includes/team-users.jsx
whimsicallyson: I think I updated this properly?

src/presenters/includes/team-users.jsx
```diff
@@ -92,9 +94,10 @@ export class AddTeamUser extends React.Component {
   
   async inviteUser(togglePopover, user) {
     togglePopover();
-    this.setState({
+    this.setState((state) => ({
       invitee: getDisplayName(user),
-    });
+      alreadyInvited: [...this.state.alreadyInvited, user],
```
judeallred: almost; you need to use the `state` object that comes in from the lambda rather than the `this.state` object.

src/presenters/includes/team-users.jsx
judeallred: almost :-)  see comment on latest code

src/presenters/includes/team-users.jsx
whimsicallyson: done!

pketh:
Awesometown, can you upload a new screen w the opacity?

src/presenters/includes/team-users.jsx
whimsicallyson: whoops, *now* its fixed.

whimsicallyson:
screen shot 2018-11-02 at 1 41 15 pm
@pketh

pketh:
thanks @whimsicallyson , looks great. I think because the ghost users now appear ghostly, they can appear in the normal list of users to the left of the button. 

btw what does it look like if you add a user with an unknown email? do you get the colored anon face?

whimsicallyson:
added support for showing added-user-by-email (whenever that ships) and removed unnecessary padding now that the avatars are to the left of the button.

Jenn says this is fine to ship without persistence as long as that part is going to be worked on soon, so @judeallred if you think we're good to go on all sides, let me know and we can merge/release!

judeallred:
Sounds good,  could you please extract her Jenn's persistence request out of this PR and into a notion card?  

functionally lgtm now, taking a glance at the latest code...

src/presenters/includes/team-users.jsx
```diff
@@ -92,17 +94,19 @@ export class AddTeamUser extends React.Component {
...
     await this.props.inviteUser(user);
   }
   
   async inviteEmail(togglePopover, email) {
     togglePopover();
-    this.setState({
+    this.setState((state) => ({
       invitee: email,
-    });
+      alreadyInvited: [...state.alreadyInvited, email],
```
judeallred: this doesn't seem right to me -- it looks like we're appending email addressses to a list of users.  I'd expect this to fail down the line when the <UsersList> receives a string email address instead of a user object.

judeallred: i think it'll be hard to test this until invite-by-email (for non preexisting users) exists, so I'd suggest ignoring this part of the feature.  Speculative code is hard to write; we can add it once email invites are implemented

whimsicallyson: ok, took it out 👍

PR #144, opened 2018-11-05
Longing elf - Collections Code Review

ttseng:
Code review **only** for collections feature on [https://longing-elf.glitch.me/](https://longing-elf.glitch.me/)


src/curated/categories.js
```diff
@@ -92,4 +92,16 @@ export default [
     "description": "The sounds of music created, connected, visualized and remixed.",
     "backgroundColor": "rgba(162, 193, 255, 0.5)",
   },
-];
\ No newline at end of file
+];
+
+{/*
```
judeallred: best to clean this out

src/presenters/collection-editor.jsx
```diff
@@ -0,0 +1,133 @@
...
+  
+  updateAvatar(avatarUrl){
+    this.setState({
+      avatar: avatarUrl
+    });
+    this.updateFields({avatarUrl});
+  }
+
+  async updateFields(changes) {
+    // console.log('update collection fields');
```
judeallred: stray comment

src/presenters/includes/collection-avatar.jsx
```diff
@@ -0,0 +1,33 @@
...
+  componentDidMount() {
+    this.syncColor();
+  }
+  
+  componentDidUpdate() {
+    this.syncColor();
+  }
+  
+  render() {
+    return <span ref={this.ref}><SVGInline svg={defaultAvatarSVG}/></span>;
```
judeallred: this is neat, good job!

src/presenters/includes/collection-result-item.jsx
```diff
@@ -0,0 +1,89 @@
...
+const addProject = (addProjectToCollection, project, collection, collectionPath, notification, togglePopover) => {
+
+  // add project to collection via api
+  addProjectToCollection(project, collection);
+  
+  // toggle popover
+  togglePopover();  
+  
+  // show notification
+  // TO DO - only show this if add project to collection completes successfully
```
judeallred: TODO's should be extracted to notion cards, then we can delete the comments.  (once we're not actively in the code, it's easy to lose track of todo's :-) )

src/presenters/includes/collection-result-item.jsx
```diff
@@ -0,0 +1,89 @@
...
+  const content = <AddProjectMessage projectName={project.domain} collectionName={collection.name} url={collectionPath}/>;
+  notification(content, "notifySuccess");
+};
+
+// NOTE: CAN WE PASS CURRENT USER INSTEAD OF DOING AN API REQUEST TO GET THE USER LOGIN HERE?
+async function getCollectionUrl(api, userId, collectionUrl){
```
judeallred: yup, and you can get the user from the user context

src/presenters/includes/collection-result-item.jsx
```diff
@@ -0,0 +1,89 @@
...
+  const {data} = await api.get(`users/${userId}`);
+  const username = data.login;
+  let path = `/@${username}/${collectionUrl}`;
+
+  return path;
+}
+
+const CollectionResultItem = ({addProjectToCollection, api, project, collection, isActive, togglePopover}) => {
+  var resultClass = "button-unstyled result result-collection";
```
judeallred: use 'let' please instead of 'var'

src/presenters/includes/edit-collection-name-and-url.jsx
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
```
judeallred: [note to self; I left off here. Will resume from this point in the PR as I continue reviewing tomorrow :-) ]

src/models/collection.js
```diff
@@ -0,0 +1,72 @@
...
+
+export const avatars = {
+  computer: "https://cdn.hyperdev.com/us-east-1%3Acba180f4-ee65-4dfc-8dd5-f143280d3c10%2Fcomputer.svg",
+  tetris: "https://cdn.hyperdev.com/6ce807b5-7214-49d7-aadd-f11803bc35fd%2Ftetris.svg",
+  robot: "https://cdn.hyperdev.com/6ce807b5-7214-49d7-aadd-f11803bc35fd%2Frobot.svg",
+  hardware: "https://cdn.gomix.com/6ce807b5-7214-49d7-aadd-f11803bc35fd%2Fhardware.svg",
+  art: "https://cdn.glitch.com/2bdfb3f8-05ef-4035-a06e-2043962a3a13%2Fart.svg?1499357014248",
+  music: "https://cdn.glitch.com/2bdfb3f8-05ef-4035-a06e-2043962a3a13%2Fmusic.svg?1502555440002",
+};
+
```
GregWeil: I think everything above this is dead code and can be removed.

src/presenters/collections-list.jsx
```diff
@@ -0,0 +1,196 @@
...
+      this.setState({newCollectionUrl: newCollectionUrl});
+      this.setState({shouldRedirect: true});
+      return true;
+    }
+    return false;
+  }
+  
+  async generateNames(userLogin) {
+    let collectionSynonyms = ["mix","bricolage","playlist","assortment","potpourri","melange","album","collection","variety","compilation"];
+    let predicate = userLogin;
```
GregWeil: Since collections are already scoped to the user, why use their login instead of a fixed predicate?

src/presenters/collections.jsx
```diff
@@ -0,0 +1,23 @@
...
+            </div>
+            <div className="category-box-label centered" style={{backgroundColor: color}}>{name}</div>
+          </li>
+        </a>
+      ))}
+    </ul>
+  </section>
+);
+
+export default Categories;
```
GregWeil: This should get renamed back to categories.jsx as long as it's rendering categories and not collections

src/presenters/includes/add-collection-avatar.jsx
```diff
@@ -0,0 +1,27 @@
...
+      )}
+    </PopoverContainer>
+  );
+};
+
+AddCollectionAvatar.propTypes = {
+  api: PropTypes.func.isRequired
+};
+
+export default AddCollectionAvatar;
```
GregWeil: This isn't coming back, right? If not it should get removed

src/presenters/includes/collection-result-item.jsx
ttseng: oh, so actually there is probably an issue here worth resolving, which is that I don't handle if addProjectToCollection throws an error - I always add the success notification after the API request is made.  Do you know if there's an example of error notifications I can look at for reference?

src/presenters/includes/editable-wrapping-field.jsx
```diff
@@ -0,0 +1,73 @@
...
+    super(props);
+    this.state = { id: uniqueId("editable-field-") };
+    this.textInput = React.createRef();
+    this.onChange = this.onChange.bind(this);
+  }
+  
+  componentDidMount() {
+    if (this.props.autoFocus) {
+      this.textInput.current.select();
+    }
```
judeallred: autofocus on mount should be a default behavior in react;  if that works then we should be able to do away with this ref... i'll play with it for a sec...

judeallred: yup, all still works without that code.  I edited longing-elf and pulled it out.

judeallred: lmk if there's something i'm missing here :)

src/presenters/includes/project-result-item.jsx
```diff
@@ -4,20 +4,31 @@ import PropTypes from 'prop-types';
 import {getAvatarUrl} from  '../../models/project';
 import {StaticUsersList} from '../users-list.jsx';
 
-const ProjectResultItem = ({id, domain, description, users, action, isActive}) => {
-  var resultClass = "button-unstyled result ";
+const ProjectResultItem = ({id, domain, description, users, action, isActive, isPrivate}) => {
+  var resultClass = "button-unstyled result result-project";
```
judeallred: mostly personal preference, but I tend to favor composing variables once instead of testing and modifying them.  i wrote to look like this:
```
  const activeClass = isActive ? "active" : "";
  const privateClass = isPrivate ? "private" : "";
  const resultClass = `button-unstyled result result-project ${activeClass} ${privateClass}`;
```

src/presenters/overlays/overlay-select-collection.jsx
```diff
@@ -0,0 +1,94 @@
...
+    </PopoverContainer>
+  </>
+);
+OverlaySelectCollection.propTypes = {
+  children: PropTypes.node.isRequired,
+  domain: PropTypes.string.isRequired,
+};
+
+
+export default OverlaySelectCollection;
```
GregWeil: This isn't used

src/presenters/overlays/overlay-select-collection.jsx
```diff
@@ -0,0 +1,94 @@
...
+import PropTypes from 'prop-types';
+
+import PopoverContainer from '../pop-overs/popover-container.jsx';
+import categories from '../../curated/categories.js';
+
+import CollectionResultItem from '../includes/collection-result-item.jsx';
+
+const OverlaySelectCollection = ({children, domain}) => (
+  <>
+    <PopoverContainer>
```
judeallred: we only need the fragments when there isn't a different top-level wrapper.  In this case, we have PopoverContainer, so th e fragment can be removed

judeallred: [fixed]

src/presenters/overlays/overlay-select-collection.jsx
```diff
@@ -0,0 +1,94 @@
...
+          <dialog className="overlay overlay-collection">
+            <section className="pop-over-actions">
+              <section className="title">
+                <span>Add {domain} to collection</span>
+              </section>            
+              <section className="pop-over-actions results-list">
+                <ul className="results">
+                  <li>
+                    <CollectionResultItem
+                      projectName={domain}
```
judeallred: this list should be able to simplify to:
```
{ categories.slice(0,4).map( ({name, description, id, avatarUrl, url}) => (
        <li>
                  <CollectionResultItem
                    projectName={domain}
                    collectionName={name}
                    description={description}
                    id={id}
                    avatarUrl={avatarUrl}
                    url={url}
                    isActive={false}
                    togglePopover={togglePopover}
                  />
        </li>
))}
```

src/presenters/collections-list.jsx
ttseng: I guess this does lead to some redundancies in the URL, but it also may make the collection easier to associate with a user when it does appear elsewhere on the site (for example, search results, when we add collections there).  I don't have a strong preference here, though.  do you have a specific predicate in mind?

src/presenters/includes/project-result-item.jsx
ttseng: cool that's cleaner

src/presenters/overlays/overlay-select-collection.jsx
ttseng: yeah, this was an earlier prototype that used an overlay rather than a nested pop-over.  will remove.

src/presenters/pages/collection.jsx
```diff
@@ -0,0 +1,265 @@
...
+    if(loop.stop){return;}
+    if(el.url === collectionName){
+      collectionMatch = data[i];
+      loop.stop = true;
+    }
+  });
+  if(!collectionMatch){
+    return null;
+  }
+  return collectionMatch.id;
```
GregWeil: This can be simplified into `const collectionMatch = data.find(c => c.url === collectionName);

GregWeil: Also, downloading the full set of collections (and projects in them) for the user will scale pretty badly. I'm fine with it being like this for now, but we should keep it in mind to change soon (which will need api side support)

src/presenters/pages/team.jsx
```diff
@@ -136,13 +142,31 @@ class TeamPage extends React.Component {
...
+            addProjectToCollection: this.addProjectToCollection,
+            removeProjectFromTeam: this.props.removeProject,
+            joinTeamProject: this.props.joinTeamProject,
+            leaveTeamProject: this.props.leaveTeamProject,
+          }}
+          api={this.props.api}
+        />
+
+        <EntityPageRecentProjects
           projects={this.props.team.projects}
```
GregWeil: Having two components that are always used together and always take the same props is pretty redundant. Not a blocker, but I wonder what we can do about this in the future.

src/presenters/pages/team.jsx
```diff
@@ -136,13 +142,31 @@ class TeamPage extends React.Component {
...
           pins={this.props.team.teamPins}
           isAuthorized={this.props.currentUserIsOnTeam}
           addPin={this.props.addPin}
           removePin={this.props.removePin}
+          addProjectToCollection={this.addProjectToCollection}
           projectOptions={{
+            addProjectToCollection: this.addProjectToCollection,
```
GregWeil: Both of these have addProjectToCollection twice, and both addPin and removePin. Can these redundant props get cleaned up for readability?

src/presenters/pop-overs/add-collection-project-pop.jsx
```diff
@@ -0,0 +1,258 @@
...
+    if(searchByUrl){  
+      // get the single result that matches the URL exactly - check with https://community.glitch.me/
+      nonCollectionResults = results.filter(result => result.domain == query);
+      
+      // check if the project is already in the collection
+      if(nonCollectionResults.length > 0 && collectionProjectIds.includes(nonCollectionResults[0].id)){
+        nonCollectionResults = [];
+        this.setState({projectName: query});
+      }  
+    } else {
```
GregWeil: This fails for community, because there are too many irrelevant results that it doesn't get found at all. If url searches will only show the single exact match it should probably not use the search api at all, and go right to `/projects/:domain`

src/presenters/pop-overs/project-options-pop.jsx
```diff
@@ -49,13 +50,29 @@ const ProjectOptionsPop = ({...props}) => {
...
+            </section>
+          )}
+          
+          <section className="pop-over-actions">
+            {!!props.addProjectToCollection && <PopoverButton onClick={addToCollectionPopover} {...props} text="Add to Collection " emoji="framed_picture"/>}
+          </section>
+        </>
+        : <section className="pop-over-actions">
+          {!!props.addProjectToCollection && <PopoverButton onClick={addToCollectionPopover} {...props} text="Add to Collection " emoji="framed_picture"/>}
+        </section>
```
GregWeil: The addProjectToCollection button exists regardless of the conditional, so it can live outside of it rather than duplicated.

Also, requiring that the user is on the project to pin and unpin it means you'll have to join team projects before you can pin them

GregWeil:
Signing out on a team page causes a `currentUser is null` error in the console and triggers the error boundary

src/presenters/pages/collection.jsx
ttseng: thanks for the suggestion!  i updated the method.

src/presenters/pop-overs/add-collection-project-pop.jsx
```diff
@@ -0,0 +1,258 @@
...
+    const collectionProjectIds = this.props.collection.projects.map((project) => project.id);
+    
+    if(isUrl(query)){
+      searchByUrl = true;
+      // check if the query is a URL or a name of a project
+      // Project URL pattern: https://add-to-alexa.glitch.me/, https://glitch.com/~add-to-alexa
+      let queryUrl = new URL(query);
+      if(queryUrl.href.includes("me") && !queryUrl.href.includes("~")){
+        // https://add-to-alexa.glitch.me/
+        query = queryUrl.href.substring(queryUrl.href.indexOf("//")+"//".length, queryUrl.href.indexOf("."));
```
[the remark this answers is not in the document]
ttseng: thanks for the suggestion!

src/presenters/pop-overs/add-collection-project-pop.jsx
ttseng: I think we had determined there was an issue with "community" more generally, as it doesn't appear in search results at all. https://longing-elf.glitch.me/search?q=community
@judeallred @ischwarz 

I can switch it to using the project endpoint rather than the search endpoint for URL search entities

ttseng: I switched the endpoint to `projects/domain` as Greg suggested when the user enters a URL into the search input field.

src/presenters/pop-overs/project-options-pop.jsx
ttseng: I'm not sure I follow the second point - if the user is not on the project, they won't be able to pin or unpin it right?

GregWeil: You can pin projects you aren't a member of on team pages

ttseng: oh I see now.  I guess it's not clear to me what any of the permissions are on teams.  can any user on the team pin a project, or only admins, for example?

src/presenters/pop-overs/add-collection-project-pop.jsx
```diff
@@ -0,0 +1,268 @@
...
+        
+          {!!this.state.maybeResults && 
+            <ProjectsLoader api={this.props.api} projects={this.state.maybeResults}>
+              {() => <ProjectSearchResults
+                projects={this.state.maybeResults}
+                onClick={this.onClick}
+                collection={this.props.collection}
+                projectName={this.state.projectName}
+                excludedProjectsCount={this.state.excludedProjectsCount}
+              />}
```
GregWeil: This isn't actually using the projects from the `ProjectsLoader`. It would get them as the first param to the child function on line 245

src/presenters/pop-overs/add-project-to-collection-pop.jsx
```diff
@@ -0,0 +1,173 @@
...
+      await this.props.api.patch(`collections/${newCollection.id}/add/${this.props.project.id}`);         
+      
+      // redirect to that collection
+      const newCollectionUrl = getLink(this.props.currentUser.login, newCollection.url);
+      this.setState({newCollectionUrl});
+    }catch(error){
+      if (error && error.response && error.response.data && error.response.data.message) {
+        this.setState({error: error.response.data.message});
+      } else {
+        window.Raven.captureException(error);
```
GregWeil: A recent pr removed `Raven`. Every reference to it should import what it wants: `import {captureException} from '../../utils/sentry';`

GregWeil: (You'll also have to merge in master before that import will work)

src/presenters/pop-overs/project-options-pop.jsx
judeallred: I believe that all users on team should have the same permissions on all the team projects,  including adding/removing themselves to them, adding/removing them from the team, etc.  The team Admins get some bonus powers in configuring the team itself -- deleting the team, changing its avatar, etc.    I'd expect this to carry through to pins (and eventually team collections) as well

ttseng: got it - I fixed it so that any user that's authorized can pin and unpin projects within the team.

src/presenters/pop-overs/add-collection-project-pop.jsx
ttseng: thanks!

src/presenters/collections-list.jsx
ttseng: I chose "radical" ((:

src/presenters/includes/collection-result-item.jsx
ttseng: I wrapped it all in a try statement and show a different error message if adding the project fails
```
  try{
    // add project to collection via api
    addProjectToCollection(project, collection);

    // toggle popover
    togglePopover();  

    // show notification
    const content = <AddProjectMessage projectName={project.domain} collectionName={collection.name} url={collectionPath}/>;
    notification(content, "notifySuccess");
  }catch(error){
    const content = <p>Something went wrong. Try refreshing and adding the project again.</p>;
    notification(content, "notifyError");
  }
```

src/presenters/pop-overs/add-project-to-collection-pop.jsx
ttseng: fixed

ttseng:
merged with master, and also had to fix a styling bug with the invalid hex error message with the custom color picker

PR #145, opened 2018-11-05
Superb guide: Webpack config cleanup + some slight perf improvements + limit webpack memory

GregWeil:
Will create inline comments to explain changes...


package.json
```diff
@@ -4,7 +4,7 @@
   "description": "Find cool projects and people using Glitch",
   "scripts": {
     "prestart": "bash sh/watchers.sh &",
-    "start": "nodemon ./server/server.js --watch server --watch views -L"
+    "start": "nodemon --watch server --watch views ./server/server.js"
```
GregWeil: From the nodemon docs seemed ambiguous about whether params after the script were sent to the script or nodemon, so I just moved them before it. `-L` uses legacy watchers, which are less efficient and don't seem to be necessary

webpack.config.js
```diff
@@ -6,127 +6,121 @@ const TerserPlugin = require('terser-webpack-plugin');
...
+    new ManifestPlugin({
+      fileName: "styles.json",
+      filter: ({isInitial, name}) => isInitial && name.endsWith('.css'),
+    }),
+    new MiniCssExtractPlugin({
+      filename: "[name].css?[contenthash]"
+    }),
+  ],
+  watchOptions: {
+    ignored: /node_modules/,
```
GregWeil: Don't watch the node_modules folder, ultimately didn't have a noticeable impact on memory usage

webpack.config.js
```diff
@@ -6,127 +6,121 @@ const TerserPlugin = require('terser-webpack-plugin');
...
-              loader: 'stylus-loader',
-              options: {
-                compress: true, // Compress CSS as part of the stylus build
-                use: [AutoprefixerStylus()],
-              },
+          },
+          {
+            loader: 'stylus-loader',
+            options: {
+              compress: true, // Compress CSS as part of the stylus build
```
judeallred: i wonder if we should do something like "compress = mode==='development'"?

webpack.config.js
```diff
@@ -6,127 +6,121 @@ const TerserPlugin = require('terser-webpack-plugin');
...
-                sourceMap: true,
-              },
+      },
+    },
+    minimizer: [
+      new TerserPlugin({terserOptions: {safari10: true}}),
+    ],
+  },
+  resolve: {
+    extensions: ['.js', '.jsx'],
```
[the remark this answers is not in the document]
judeallred: Wondering if we should do a seperate PR to clear out all the .jsx entries....

PR #146, opened 2018-11-06
Spiky juice -- bugfix on the 'page not found' starfield

judeallred:
minor bugfix for the stars page

old behavior, after a few minutes:
image

fixed behaviorbehavior:
image


GregWeil:
The additional vs remaining stars change had me really confused, but I think the real fix is the no change check. We rely on the browser to clear the canvas on size change and otherwise will just keep redrawing the initial stars to the canvas. It didn't get to your screenshot's level, but I could imagine it getting there if you were logging in/out in another window, which would double the number of stars each time.

judeallred:
yup that's right!  sorry i didn't call that out;  I made that first change as an attempted fix and then decided to keep it as a refactor

PR #152, opened 2018-11-08
🏬⭐️ raincoder-community — make `/careers` work

cori:
./server/proxy.js:242011/71

fixes https://fogcreek.slack.com/archives/C3U9U59NE/p1541710318041500 (mostly just stretching my legs). https://raincoder-community.glitch.me/careers/ in action.

### DIFF
--- a/server/proxy.js
+++ b/server/proxy.js
@@ -50,7 +50,6 @@ module.exports = function(app) {
   proxyGhost('culture', 'culture-zine.glitch.me');
   proxyGhost('about', 'about-glitch.glitch.me');
   proxyGhost('legal', 'about-glitch.glitch.me', '/about');
-  proxyGhost('careers', 'about-glitch.glitch.me', '/about');
   
   // Pages hosted by 'about.glitch.me':
   [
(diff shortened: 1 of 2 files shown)


judeallred:
Sadly, it's bad mojo to have duplicate pages at https://raincoder-community.glitch.me/about/careers/ and https://raincoder-community.glitch.me/careers/ , would be better to redirect from one two the other if we need to support the pair.

because /about/careers/ is the cannonical url, we get some weirdness like this link will always point to the /about/careers/ location:
image


Before we take this approach, can we verify that we have to keep the canonical spot at /about/careers?  I believe Gareth made that original routing choice. If we could move it to /careers/ (and add a redirect from about/careers -> careers) that would get us to a better place in the long run.

judeallred:
(unrelated to these changes;   Hi cori!  great to see your first PR!
  If it's helpful to you, here's the flow that we like to use: https://glitch.com/edit/#!/community?path=CONTRIBUTING.md:1:0

..but it's also perfectly to just send in PR's as they feel appropriate to you, like this one.  It probably deserved a better title that "🏬⭐️ Checkpoint #152" (we usually do a title like {remix-name} -- {feature description} ),  but other than that this is plenty clear :-)

)

cori:
Updated to use a redirect instead of a proxy to maintain `/about/careers` as canonical.

judeallred:
I think you did a great job here cori! I was calling out the doc because I didn't want to assume that you found it;  since you did, awesome

cori:
ah perfect, thanks! It seemed like maybe I either went awry somewhere or the doc was out-of-date, and I wanted to make sure I understood it and was following the preferred process; since I've already pointed someone at the process I'd like to make sure I understand it.

PR #154, opened 2018-11-09
Mercury saw -- fix null-user crashers on project, search, and category pages

judeallred:
more quirks of the 'null' case of our currentUser consumer


GregWeil:
This'll fix crashes, but it also means that signing out will kick the project and category page back to the loading spinner, and the search page whites out for a moment before going to the spinner.

judeallred:
correct,  which I think is a good thing on short notice, but needs more work

PR #155, opened 2018-11-11
Maple harmonica - switch between API backends based on hostname

etamponi:
This PR allows the Platform team (and anyone else in the company, actually!) to let the community site or any of its remixes use the _development_ API that they are running in their devbox. Of course this only applies to them, the real community site at glitch.com is not affected at all :)

The goal of the PR is to make it easier to collaborate between the Platform team and the Community team: API changes can be tested directly against the community site _before being pushed to production_, and bugfixes can be done _in real time_.

As described in [this notion case](https://www.notion.so/glitch/87bda09f85184c1d81e1659aa62e4392?v=9f7963becc9341ed9af502682c6e9eef&p=92d2d65272d74da7bc59d3397af57275), API is switched on the frontend based on the hostname on which the community site is served.

This PR also add the following fixes (just to show myself and everyone that this is _actually_ super helpful!):

- The New Project popup works on all environments
- The "Show Project" button works on all environments
- The embed in the project page works on all environments
- The Analytics page _finally_ uses the right value for visits!

Additionally, I've been using this remix of the community site against my development API changes that are in [this other PR](https://github.com/FogCreek/Glitch/pull/826), and it was great, I was able to fix several bugs that couldn't be found by unit tests nor manual API calls!

### DIFF
--- a/server/routes.js
+++ b/server/routes.js
@@ -102,7 +102,6 @@ module.exports = function(external) {
   });
 
   app.get('*', async (req, res) => {
-    console.log(req.headers);
     await render(res,
       "Glitch",
       "The friendly community where everyone can discover & create the best stuff on the web");
--- a/server/routes.js
+++ b/server/routes.js
@@ -69,11 +69,12 @@ module.exports = function(external) {
       ZINE_POSTS: JSON.stringify(zine),
       PROJECT_DOMAIN: process.env.PROJECT_DOMAIN,
       ENVIRONMENT: process.env.NODE_ENV || "dev",
-      ...constants,
+      CONSTANTS: constants,
     });
   }
 
-  const {CDN_URL} = constants;
+  const env = process.env.RUNNING_ON === "staging" ? "staging" : "production";
+  const {CDN_URL} = constants[env];
 
   app.get('/~:domain', async (req, res) => {
     const {domain} = req.params;
--- a/server/routes.js
+++ b/server/routes.js
@@ -102,6 +102,7 @@ module.exports = function(external) {
   });
 
   app.get('*', async (req, res) => {
+    console.log(req.headers);
     await render(res,
       "Glitch",
       "The friendly community where everyone can discover & create the best stuff on the web");
--- a/server/routes.js
+++ b/server/routes.js
@@ -73,8 +73,7 @@ module.exports = function(external) {
     });
   }
 
-  const env = process.env.RUNNING_ON === "staging" ? "staging" : "production";
-  const {CDN_URL} = constants[env];
+  const {CDN_URL} = constants.current;
 
   app.get('/~:domain', async (req, res) => {
     const {domain} = req.params;
--- a/sh/webpack.sh
+++ b/sh/webpack.sh
@@ -2,12 +2,14 @@
 
 # Webpack will occassionally crash, so run it in a loop so that it'll come back to life on failure.
 
+export TMP=/app/.tmp
+
 eslint --config server/.eslintrc.server.js webpack.config.js
 
 echo "Starting webpack watcher"
 
 while true; do
-  env NODE_OPTIONS="--max-old-space-size=256" webpack --watch --info-verbosity verbose
+  webpack --watch --info-verbosity verbose
   
   echo "Webpack crashed; Restarting webpack --watch"
 done
--- a/sh/webpack.sh
+++ b/sh/webpack.sh
@@ -2,8 +2,6 @@
 
 # Webpack will occassionally crash, so run it in a loop so that it'll come back to life on failure.
 
-export TMP=/app/.tmp
-
 eslint --config server/.eslintrc.server.js webpack.config.js
 
 echo "Starting webpack watcher"
--- a/sh/webpack.sh
+++ b/sh/webpack.sh
@@ -9,7 +9,7 @@ eslint --config server/.eslintrc.server.js webpack.config.js
 echo "Starting webpack watcher"
 
 while true; do
-  webpack --watch --info-verbosity verbose
+  NODE_OPTIONS="--max-old-space-size=256" webpack --watch --info-verbosity verbose
   
   echo "Webpack crashed; Restarting webpack --watch"
 done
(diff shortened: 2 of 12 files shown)


src/presenters/includes/team-analytics-project-details.jsx
```diff
@@ -47,7 +47,7 @@ const ProjectDetails = ({projectDetails}) => {
             <td>{moment(projectDetails.createdAt).fromNow()}</td>
           </tr>
           <tr>
-            <td className="label">Last code view</td>
```
judeallred: do you have a case or a notion card around this bugfix?

src/presenters/pop-overs/new-project-pop.jsx
```diff
@@ -14,7 +14,7 @@ const NewProjectPop = ({projects}) => (
       <div className="results">
         {projects.length ? projects.map((project) => (
           <Link key={project.id} to={getRemixUrl(project.domain)}>
-            <ProjectResultItem {...project} users={[]} action={()=>{
+            <ProjectResultItem {...project} cdnUrl="https://cdn.glitch.com" users={[]} action={()=>{
```
judeallred: for the cases where we want to pick a specific env, i think we should specify the env rather than repeating the hardcoded value.

e.g something like 'env.production.cdnUrl'.

i'll come back to this in later comments....

src/presenters/includes/team-analytics-activity.jsx
```diff
@@ -21,7 +21,7 @@ const createHistogram = (bins) => {
         timestamp = data['@timestamp'];
       }
       totalRemixes += data.analytics.remixes;
-      totalAppViews += data.analytics.requests;
+      totalAppViews += data.analytics.visits;
```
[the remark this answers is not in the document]
etamponi: yes

src/presenters/includes/team-analytics-project-details.jsx
etamponi: I don't have a case or a notion card... I just found it and fixed it... since it's a pretty severe bug.

src/models/project.js
```diff
@@ -14,27 +14,31 @@ export default function Project({teams, users, ...project}) {
   };
 }
 
-export function getAvatarUrl(id) {
-  return `${CDN_URL}/project-avatar/${id}.png`;
+export function getAvatarUrl(id, cdnUrl) {
```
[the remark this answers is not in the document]
etamponi: oh... right! :D

sh/webpack.sh
```diff
@@ -7,7 +7,7 @@ eslint --config server/.eslintrc.server.js webpack.config.js
 echo "Starting webpack watcher"
 
 while true; do
-  env NODE_OPTIONS="--max-old-space-size=256" webpack --watch --info-verbosity verbose
+  NODE_OPTIONS="--max-old-space-size=256" webpack --watch --info-verbosity verbose
```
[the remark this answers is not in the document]
etamponi: I think they're the same, but I can revert back

views/index.ejs
```diff
@@ -37,17 +37,27 @@
   </head>
   <body> 
     <script>
-      var API_URL = '<%= API_URL %>';
-      var APP_URL = '<%= APP_URL %>';
-      var BUILD_COMPLETE = <%- BUILD_COMPLETE %>;
-      var CDN_URL = '<%= CDN_URL %>';
-      var EDITOR_URL = '<%= EDITOR_URL %>';
+      var CONSTANTS = <%- JSON.stringify(CONSTANTS) %>;
```
judeallred: This works for a first pass.

I think what I'd prefer to do is to refactor away most of these global constants and keep the role of the index.ejs file limited to populating the global constants blob.  I'd be tempted to call it 'ENV'.  In client side code, we could add a class that has the global dependency on env, then all of our code that makes use of it could import the specific variables they need from the package.  

I'm imagining using the variables like this:

//pulling in values for your current env
import {API_URL, CDN_URL} from '../utils/env';

// pulling values from an env that's not your own
import {API_URL: production.API_URL} from "../utils.env/"

etc.   Might not look exactly like that,  but the core idea is to get to use imports instead of globals and to do the env detecting in a utility file rather than in index.ejs.

etamponi: I agree 👍

judeallred:
@GregWeil  looping in your architecture thoughts on this whole pr

sh/webpack.sh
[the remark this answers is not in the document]
judeallred: @GregWeil added it recently, he might have context on if they are or are not equivalent.  I'm not sure, personally

src/presenters/includes/team-analytics-project-details.jsx
judeallred: https://www.notion.so/glitch/87bda09f85184c1d81e1659aa62e4392?v=9f7963becc9341ed9af502682c6e9eef&p=f6604cfe72ea4fdea9ad8d3a8c917fbc  :)

etamponi:
`env` should be redundant in this case, not using it also sets the variable
only for the command.

Il giorno lun 12 nov 2018, 21:24 GregWeil <notifications@github.com> ha
scritto:

judeallred:
cool;  I'll handle the merge and deploy of this tomorrow morning :-)

server/routes.js
```diff
@@ -69,11 +69,12 @@ module.exports = function(external) {
       ZINE_POSTS: JSON.stringify(zine),
       PROJECT_DOMAIN: process.env.PROJECT_DOMAIN,
       ENVIRONMENT: process.env.NODE_ENV || "dev",
-      ...constants,
+      CONSTANTS: constants,
     });
   }
 
-  const {CDN_URL} = constants;
+  const env = process.env.RUNNING_ON === "staging" ? "staging" : "production";
```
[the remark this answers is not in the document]
etamponi: fixed :)

judeallred:
thanks for the updates;
code lgtm,  doing a quick test then can merge...

judeallred:
greg is out today, i'll go ahead with this in his absense

PR #156, opened 2018-11-12
Profuse tea: Remove the leftover model code

GregWeil:
With this change, projects/users/teams are no longer guaranteed to have their arrays set, so project.users could be one of `undefined` `[]` or a populated array

I also stuck in a few extra error boundaries and tiny fixes


src/presenters/includes/team-elements.jsx
```diff
@@ -24,7 +24,7 @@ export const VerifiedBadge = () => {
   const tooltip = 'Verified to be supportive, helpful people';
   return (
     <span data-tooltip={tooltip}>
-      <img className="verified" src={image} alt={tooltip}/>
+      <img className="verified" src={image} alt="✓"/>
```
GregWeil: Firefox is quick to show alt text, so team pages show as 'Name Verified to be...' which wraps and causes a big reflow when the image comes in.

src/presenters/current-user.jsx
```diff
@@ -177,7 +176,7 @@ class CurrentUserManager extends React.Component {
     const currentUser = cachedUser || sharedUser;
     return children({
       api: this.api(),
-      currentUser: currentUser ? UserModel(currentUser).asProps() : null,
+      currentUser: currentUser ? currentUser : null,
```
judeallred: this line may be unnecessary, might just want `currentUser,` instead?  Otherwise I'd expect to see `currentUser || null` instead of the ternary

src/presenters/pages/team.jsx
```diff
@@ -186,7 +190,7 @@ class TeamPage extends React.Component {
           </aside>
         }
 
-        { this.props.currentUserIsOnTeam &&
+        { this.props.currentUserIsOnTeam && <ErrorBoundary>
```
judeallred: good idea.

src/presenters/current-user.jsx
GregWeil: Updated with a cleanUser function that does pretty much the same thing as the model. A bunch of places were breaking on logout because the create anon and the login endpoints don't include projects with the user.

I think the eventual solution (Including for the null currentUser) is to have all of the standard user fields stored somewhere in the js for an anon user. Then do `{...genericAnon, ...user}` instead of `user ? cleanUser(user) : null`. We'll need some extra checks in a few places to avoid creating broken links (ex no logout button or profile links) but everything else should work just fine if it gets a user with id -1 and all defaults

PR #157, opened 2018-11-12
Jelly alley: don't show Collections section if there are no collections

whimsicallyson:
https://jelly-alley.glitch.me/

https://www.notion.so/glitch/collections-hide-collections-section-when-viewing-a-user-who-has-no-collections-0df0bfdf1bfe482c9e59b24f2df7e70a

Screenshot of nothing when looking at the page of a user with no collections. On your own page, even without collections, the section is there so you can begin adding them.
screen shot 2018-11-12 at 3 47 44 pm


src/presenters/collections-list.jsx
```diff
@@ -37,15 +37,18 @@ class CollectionsList extends React.Component {
...
-            {!collections.length && <CreateFirstCollection {...{api, urrentUser: maybeCurrentUser}}/>}
+            {!collections.length && <CreateFirstCollection {...{api, currentUser: maybeCurrentUser}}/>}
           </>
         }
-        
-        <CollectionsUL {...{collections, api, isAuthorized, deleteCollection, userLogin}}/>
+        {!!collections.length && 
+          <CollectionsUL {...{collections, api, isAuthorized, deleteCollection, userLogin}}/>
+        }
       </article>
```
judeallred: I think this is a good circumstance to break this case out from the rest.  Something like :
```
return {
  const hasCollections = !!collections.length;

  if(!hasCollections && !isAuthorized) {
     return null;
  }

   return (
     <article> 
        ...
         {hasCollections && ...}
       ...
    );
}
```

I feel like that helps us unpack the logic a little,  otherwise it feels like each line of this file is handling yet a different view case and it's hard to figure out.

whimsicallyson: updated!

src/presenters/collections-list.jsx
```diff
@@ -34,17 +34,21 @@ class CollectionsList extends React.Component {
...
+    }
     return (
       <article className="collections">
         <h2>{title}</h2>
-        {isAuthorized && !!maybeCurrentUser &&
+        {canMakeCollections &&
           <>
             <CreateCollectionButton {...{api, currentUser: maybeCurrentUser}}/>
-            {!collections.length && <CreateFirstCollection {...{api, urrentUser: maybeCurrentUser}}/>}
+            {!collections.length && <CreateFirstCollection {...{api, currentUser: maybeCurrentUser}}/>}
```
judeallred: I'd probably use`!hasCollections` here, since we have it as a utility now,  otherwise lgtm

whimsicallyson: done

PR #161, opened 2018-11-13
Avatar shapes

ttseng:
https://avatar-shapes.glitch.me

Restyle user avatars & team avatars on profile and team pages to help distinguish between entities.

More info here: https://www.notion.so/glitch/Distinguish-between-user-avatar-vs-team-avatar-c475ed420dd444cba0b1ecadcca4e0a7

profile page

team page


judeallred:
spotted a couple of style quirks on the teams page:
- alignment of @ sign
- styles of the 'invited' display

image

judeallred:
the user icon in the corner should probably be a circle too, yeah?
image

ttseng:
yup, I just realized the user avatar in the header should also be round and pushed an update.
re: the alignment of @ sign, looks like that's an issue on production.  I'll fix it and look into the invited pop-over styling.

judeallred:
thanks

judeallred:
also these -- 
image

judeallred:
also noticed that some of the inputs are a little smushed in avatar-shapes  (maybe because of the fix for the @ sign above :-p  )
image

ttseng:
looks like the inline-styling for teams is incorrect on production:

image

I think it might be best to move that fix into a separate PR.

ttseng:
A few more updates:

* Update `project-result-item` to include new user styling
screen shot 2018-11-13 at 5 33 53 pm

* Updated `add-team-user` to include new user styling and condense layout 
screen shot 2018-11-13 at 5 46 23 pm

ttseng:
also needs to be fixed in the embed, but I'm not sure where that code lives
image

ttseng:
updated styling on user profile page to remove border separating user info with team associations

image

ttseng:
Updated search results
image

ttseng:
fixed the glitch avatar, which seems to be its own category of avatar 
image

ttseng:
@GregWeil ohh I didn't realize it was the glitch team, I thought it was a Glitch user.  I'll fix it.

ttseng:
fixed wrapping of user avatars in team search results (thanks for the suggestion Greg!)
image

also changed avatar shapes for projects owned by a team
image

GregWeil:
User page team hovers are off center
image

styles/pages/user-and-team-page.styl
```diff
@@ -20,4 +38,68 @@
...
+      background-color: success-background
+    .users
+      opacity: 0.5
+  
+  .join-team-button
+    vertical-align: top
+    margin-top: 3px
+    
+    
+  .beta
```
pketh: `.beta` doesn't seem to be referenced anywhere else in this pr. Is this for a seperate thing?

styles/styles.styl
```diff
@@ -33,7 +33,6 @@
 @require 'includes/zine-items'
 @require 'pages/user-and-team-page'
 @require 'pages/project'
-@require 'pages/team'
```
GregWeil: This file should get deleted if it isn't referenced

pketh:
the 'made with glitch' placeholder avatar isn't round
screen shot 2018-11-15 at 5 27 28 pm


the teams that I'm a part of are part of my identifying metadata just like name and description, they can be listed out just like other metadata without the awkwardness complexity of the lighter background and the fully left alignment. Most users won't belong to too many teams, or any teams, so the extra horizontal space of a separate bar isn't really needed.
screen shot 2018-11-15 at 5 29 33 pm


this isn't a blocker, but with the circle avatars the 'upload avatar' button looks awk/ not as nice/integrated into the design anymore
screen shot 2018-11-15 at 5 31 28 pm

styles/styles.styl
ttseng: I removed it and moved all its contents to an existing styl file `pages/user-and-team-page` to keep our files more organized

styles/pages/user-and-team-page.styl
ttseng: I think it has to do with the beta flag that @whimsicallyson added to team pages

ttseng:
Thanks for catching the tooltip misalignment @GregWeil - I fixed it:

screen shot 2018-11-15 at 5 49 09 pm
screen shot 2018-11-15 at 5 50 19 pm

styles/styles.styl
GregWeil: `pages/team.styl` is still there though 😕

ttseng:
@pketh I had originally changed the 'made with glitch' icon to round, but @GregWeil pointed out that clicking on it links to the Glitch team, so I think it should actually be rectangular with rounded corners like other team avatars.

Putting the team avatars below the user info, rather than below the user description, takes up less vertical space, especially for profiles like Jenn and yours, where you're part of lots of teams, so I prefer the existing design on avatar-shapes.glitch.me.

Here's a comparison side by side.

image

ttseng:
re: the upload avatar button positioning, we could center it:
image
 or even better IMO, we could just get rid of it trigger it when the user hovers over their existing avatar.  there might be accessibility issues with this, though - maybe @jennschiffer can comment

styles/pages/user-and-team-page.styl
whimsicallyson: all my beta flag style work was in `team.styl`, so this isn't me.

ttseng: @whimsicallyson I deleted the team.styl page because there was already team page styling in user-and-team-page.styl, and I thought having a separate team.styl file could be confusing.

whimsicallyson: gotcha. so that's probably from combining the files :)

judeallred:
[for the upload avatar button]
My favorite options are centering it, or hiding it and exposing it (at the base of the image) on hover.  I agree it looks like a visual bug to keep it anchored to the left,  but I'm comfy with all of the fixes.

jennschiffer:
i personally like having the button there, but if we were to remove it or hide it, the avatar should itself be a button element or have a [button role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role) if clicking it would trigger the upload. we should avoid hiding things behind hover states.

pketh:
ya I lean away from hiding it for both initial affordance reasons, and because hover only doesn't work on touch devices

pketh:
most users won't be like jenn and me or even be part of teams so I wouldn't optimize with our needs in mind. Of the 400+ teams currently created, most have only a single user. That'll go up but I wouldn't expect a user belonging to 4+ teams ever being hyper common, using github usage as an example.

ttseng:
Centered upload button underneath avatars
image
image

PR #162, opened 2018-11-13
Spiky juice -- fix NaN in analytics

judeallred:
https://www.notion.so/glitch/87bda09f85184c1d81e1659aa62e4392?v=9f7963becc9341ed9af502682c6e9eef&p=fa625f0fccc84fd294511b16a55eff8e

### DIFF
--- a/src/curated/sample-analytics.js
+++ b/src/curated/sample-analytics.js
@@ -9,91 +9,91 @@ export default {
     {
       "@timestamp": 1538280001000,
       "analytics": {
-        "requests": 44497,
+        "visits": 44497,
         "remixes": 0
       }
     },
     {
       "@timestamp": 1538366401000,
       "analytics": {
-        "requests": 52185,
+        "visits": 52185,
         "remixes": 4
       }
     },
     {
       "@timestamp": 1538452801000,
       "analytics": {
-        "requests": 54217,
+        "visits": 54217,
         "remixes": 0
       }
     },
     {
       "@timestamp": 1538539201000,
       "analytics": {
-        "requests": 56978,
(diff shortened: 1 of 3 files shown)


judeallred:
@etamponi just an fyi;  we accidentally broke our sample analytics page because curated/sample-analytics.js didn't include the new field that we're using

src/presenters/includes/team-analytics-referrers.jsx
```diff
@@ -39,63 +45,68 @@ const filterReferrers = (referrers) => {
   return filteredReferrers;
 };
 
-const TeamAnalyticsReferrers = ({analytics, totalRemixes, totalAppViews}) => {
```
judeallred: this big blob is just an indentation change as i swap it to a PureComponent, since it was doing a lot of rendering work but it's a pure function.

src/presenters/pages/team.jsx
```diff
@@ -74,7 +74,6 @@ class TeamPage extends React.Component {
   }
 
   render() {
-    const isLoggedIn = this.props.currentUser && this.props.currentUser.login;
```
judeallred: lint fix

src/curated/sample-analytics.js
```diff
@@ -9,90 +9,103 @@ export default {
     {
       "@timestamp": 1538280001000,
       "analytics": {
+        "visits": 44497,
         "requests": 44497,
         "remixes": 0
```
judeallred: the changes in this file are what fix the bug

etamponi: oh, here's where they came from 😅

PR #165, opened 2018-11-15
Nova chance: Ensure that there is always a currentUser object

GregWeil:
If currentUser would be null, instead act like we're an anon user. The two differences between this and a real anon is the id (`0`) and the avatar color (`#aaa`). The user pop is visible while signed out now, but it is disabled.

There's probably a lot of potential cleanup, I just hit the main 'do not render if there is no user' spots.

### DIFF
--- a/src/presenters/current-user.jsx
+++ b/src/presenters/current-user.jsx
@@ -12,11 +12,11 @@ const {Provider, Consumer} = React.createContext();
 // Default values for all of the user fields we need you to have
 // We always generate a 'real' anon user, but use this until we do
 const defaultUser = {
-  id: -1,
+  id: 0,
   login: null,
   name: null,
   description: '',
-  color: '#ccc',
+  color: '#aaa',
   avatarUrl: null,
   avatarThumbnailUrl: null,
   hasCoverImage: false,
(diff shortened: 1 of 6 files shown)


src/presenters/current-user.jsx
```diff
@@ -181,10 +192,9 @@ class CurrentUserManager extends React.Component {
   
   render() {
     const {children, sharedUser, cachedUser, setSharedUser, setCachedUser} = this.props;
-    const currentUser = cachedUser || sharedUser;
     return children({
       api: this.api(),
-      currentUser: currentUser ? cleanUser(currentUser) : null,
+      currentUser: {...defaultUser, ...sharedUser, ...cachedUser},
```
judeallred: ahahahah, fun!

src/presenters/pop-overs/user-options-pop.jsx
```diff
@@ -92,6 +92,7 @@ Are you sure you want to sign out?`)) {
         return;
       }
     }
+    togglePopover();
```
judeallred: ?

GregWeil: When you signed out and there was no currentUser the avatar menu would get removed from the dom. Now that there's always a currentUser we have to close it manually.

ttseng:
cool this is a good idea.  is there anything in particular I should test out?

src/presenters/pop-overs/user-options-pop.jsx
judeallred: ah cool, thanks

PR #166, opened 2018-11-16
spiky-juice: use cdn.glitch.com instead of cdn.[gomix,hyperdev].com

judeallred:
./src/presenters/includes/not-found.jsx:8/11
./src/presenters/includes/project-actions.jsx:8/11
./src/presenters/users-list.jsx:8/11
./src/presenters/header.jsx:8/33
./src/curated/categories.js:8/109
./src/presenters/pages/index.jsx:8/14


judeallred:
I'm going to experiment with some github API stuff on this PR, so we should see some noise here.

PR #170, opened 2018-11-20
Phase time: restyling bad browser page

whimsicallyson:
https://phase-time.glitch.me/ (turn off javascript)

https://www.notion.so/glitch/design-improvements-on-bad-browser-page-054d0b5aa76c4804bb57e53a84f457b7

...this CSS feels kind of hacky, but that might just be because I haven't written layout with floats and percentages in so long! 😂


judeallred:
I poked around with a bunch of old browsers in browserstack and the behavior looks good :-)

ttseng:
I sent Allyson a note that the SVG I exported had the shadow cut off.  I sent her an updated SVG.  Otherwise, I think it looks great!

views/index.ejs
```diff
@@ -86,8 +96,10 @@
         document.getElementById('fallback').removeAttribute('style');
         if (BUILD_COMPLETE) {
           document.getElementById('fallback-yesscript').removeAttribute('style');
+          document.getElementById('fallback-img').removeAttribute('style');
         } else {
           document.getElementById('fallback-notbuilt').removeAttribute('style');
+          document.getElementById('fallback-img').removeAttribute('style');
```
judeallred: rather than duplicating `document.getElementById('fallback-img').removeAttribute('style');`, it'd be better to pull it outside of the if/else block

judeallred:
(not a blocker for this PR, just a convenient place to think:)
with regard to the message phrasings,
```
          <div class="error-msg" id="fallback-noscript" style="display: none">
            <h1>Your browser does not support Javascript</h1>
            <p>Try using Glitch in a different browser.</p>
          </div>
          <div class="error-msg" id="fallback-yesscript" style="display: none">
            <h1>Your browser does not support the latest Javascript features</h1>
            <p>Try updating your browser to use Glitch.</p>
          </div>
```
I like to think about the sorts of people _who are not software developers_ who are likely to see these sorts of messages.  Often they're on things like school, university, or business computers which don't give them agency over browser choice.  Some amount will be on mobile devices.  Some are just folks who aren't computer savvy and are running an un-updated original version of their operating system.

So with that context in mind, I try to be super kind and assume that they're a good person stuck in a bad environment.  I'd try:
A)
-> The web browser that you're using does not support Javascript.
-> Since Glitch requires Javascript to run, please try using Glitch in a different web browser.

B) 
->The web browser you're using is missing some important Javascript features.
-> To use Glitch, please try applying your latest system updates, or try us with a different web browser.

If others like this sort of phrasing then cool;  if it's all in my head then nbd :-)

PR #173, opened 2018-11-26
New 404 page

clottman:
https://fire-element.glitch.me 
https://fire-element.glitch.me/gibberish <-- see me in action!

New - desktop: 
image

New - ☎️ (mobile)
image

Old 🗄 
image

[Notion](https://www.notion.so/glitch/87bda09f85184c1d81e1659aa62e4392?v=9f7963becc9341ed9af502682c6e9eef&p=284e201a946d4b6796d00cc47113d65a)

There are duplicate styles in error.styl because some have to work on the server. We'll consolidate them more when we add the additional error pages that are coming. 

If I'm doing anything that doesn't make sense in Glitch's style standards, please let me know! It's my first PR but I <3 feedback. 

@ttseng There was no spec for the mobile view; how do you like it? Open to changes if you want any!


ttseng:
Generally this looks good to me!  I would add a little bit of margin to `.error-page-container` so that there is space between the header and the content.  (this is what it looks like with `margin: 20px 0px;`
image

judeallred:
code lgtm;  I agree with Tiff's suggestion on the spacing.  Once that's in this should be gtg :-)

clottman:
Added that spacing!
image

PR #175, opened 2018-11-26
Superb guide: Fix homepage categories having wrong project counts

GregWeil:
The category project counts were getting stored in an array in the order that the requests came back, so they get jumbled relative to the categories they belong to. This stores them in a map of category id to count so the order they come back doesn't matter.


GregWeil:
@ttseng I think you have the wrong pr!

ttseng:
good catch! it's working well for me @GregWeil

PR #179, opened 2018-11-30
Team collections half implementation

GregWeil:
https://team-collections.glitch.me/@gregs-test-team-1/zany-playlist
This pr is a bit all over the place, it's mostly refactoring stuff handling collections to support team owned collections.

In general, I changed stuff from taking `collection, userLogin` to expecting that the collection has either a `user` or `team` field, which it gets from `collections/:id` but not `collections?userId=:id` or `POST collections`. In those cases the client knows the owning team/user and can stick it in the collection locally.

If the switch is enabled you can create collections from team pages and edit them like normal collections. Team pages list collection name/links but only to team members.

If the switch is disabled you can see team owned collections, but they'll be read only even if you're on the team. Team pages make no mention of collections.

There's also the start of a collections preview on team pages that Tiff added in before I took over the branch.

image
image
image


src/presenters/pages/collection.jsx
```diff
@@ -204,48 +220,51 @@ async function getUserIdByLogin(api, userLogin){
...
+  const collectionMatch = collections.find(c => c.url == collectionName);
+  const collection = collectionMatch && await getOrNull(api, `collections/${collectionMatch.id}`);
+  if (!collection) return null;
   
-  // get collection
-  const collection = await getCollection(api, collectionId);
+  // inject the full team so we get their projects and members
+  if (team) {
+    collection.team = team;
+  }
```
GregWeil: We don't need the members to tell if you're a member since we can use `currentUser.teams`, but we do want the team's projects for the add project menu (which currently still shows your projects and not the team's)

src/presenters/collection-item.jsx
```diff
@@ -49,7 +49,7 @@ class CollectionItem extends React.Component{
 
         {(collection &&
           <Link to={`/@${userLogin}/${collection.url}`}>
```
judeallred: I think we can get rid of this userLogin prop here and instead use your new getLink function on the collection model?

src/presenters/pages/collection.jsx
judeallred: 👍

PR #183, opened 2018-12-05
Petite frost: email login UI

whimsicallyson:
https://www.notion.so/glitch/add-email-login-header-dialog-e93b8fe045974512ad71b712f9ead315

https://petite-frost.glitch.me/ (turn on Email Login on /secret)


src/presenters/pop-overs/sign-in-pop.jsx
```diff
@@ -32,32 +32,66 @@ const SignInPopButton = (props) => (
...
-};
+  
+  async onSubmit(e) {
+    e.preventDefault();
+    this.setState({done: true});
+    try {
+      await this.props.api.post('/email/sendLoginEmail', {emailAddress:this.state.email});
+      this.setState({error: false});
+    } catch (error) {
+      this.setState({error: true});
```
GregWeil: Can you a call to `captureException(error)` here, imported from `../../utils/sentry`, that way we'll get an error report whenever email signin fails.

src/presenters/pop-overs/sign-in-pop.jsx
```diff
@@ -32,32 +32,66 @@ const SignInPopButton = (props) => (
...
-  const email = window.prompt("We'll send you a login link.\n\nWhat's your email address?");
-  if(!email) {
-    // blank or cancelled.
-    return;
+class EmailHandler extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      email: '',
+      submitted: false
```
GregWeil: The submitted state here isn't used at all, it should get replaced with done:false and error:false.

src/presenters/pop-overs/sign-in-pop.jsx
```diff
@@ -72,16 +106,16 @@ const SignInPopWithoutRouter = ({header, prompt, api, location, hash}) => (
...
         <div className="pop-over sign-in-pop">
           {header}
-          <section className="pop-over-actions last-section">
+          <section className="pop-over-actions first-section">
             {prompt}
             <SignInPopButton href={facebookAuthLink()} company="Facebook" emoji="facebook" onClick={onClick}/>
             <SignInPopButton href={githubAuthLink()} company="GitHub" emoji="octocat" onClick={onClick}/>
-            <DevToggles>
-              {(enabledToggles) => (
-                enabledToggles.includes("Email Login") && <EmailSignInButton api={api} onClick={onClick}/>
```
GregWeil: This lost the onClick handler, which records your current url so you can get back there after signing in.

GregWeil:
I left a few small comments, but the actual signin flow works without a hitch. The error handling looks good too.

GregWeil:
Oh also, Having the sign in pop error if you click submit with an empty field feels odd, maybe the button should be disabled or ignore clicks until you type something in?

ttseng:
I would suggest removing the padding on the bottom of the form element, and also on the bottom of the Email Sign in Button:

image

image

src/presenters/pop-overs/sign-in-pop.jsx
```diff
@@ -32,32 +32,66 @@ const SignInPopButton = (props) => (
...
+    );
+  }
+}
 
-const EmailSignInButton = ({api, onClick}) => (
-  <button className="button-small button-link has-emoji" onClick={() => { onClick(); jankyEmailPrompt(api); }}>
-    Sign in with Email <span aria-label="" role="img">📧</span>
+const EmailSignInButton = () => (
+  <button style={{marginTop: 10}} className="button-small button-link has-emoji" >
+    Email Sign In <span aria-label="" role="img">📧</span>
```
ttseng: I think the emojis are usually not displayed inline (since they may be OS specific) - instead, we have an emoji.styl where we point to an image of the icon, and then we use spans to display the emoji.

for example, in `entity-page-pinned-projects.jsx`

```
const pinnedTitle = (
    <>
      Pinned Projects
      <span className="emoji pushpin emoji-in-title"></span>
    </>
  );
```
and in `emoji.styl`

```
 &.pushpin
    background-image: url('https://cdn.glitch.com/55f8497b-3334-43ca-851e-6c9780082244%2Fpushpin.png?1500918744339')
```

ttseng: I can help add this in the project directly, because there is a workflow for compressing the emoji image.

ttseng: I added it (:

PR #184, opened 2018-12-05
Nova chance: replace moment-mini with dayjs

GregWeil:
https://www.notion.so/glitch/87bda09f85184c1d81e1659aa62e4392?v=78db6cf11f8a43859bd17bc800cebde0&p=43c26630b04043d8bdab4ef65fe2dd41

I added a /shared/dayjs-convert.js that is used by both the client and server, because dayjs doesn't support moment.duration().asUnits(). The dayjs.extend() side effects don't feel great (since it hides away the relativeTime dependency from where it actually gets used), but as long as we only have a few plugins it's ok.


src/client.jsx
```diff
@@ -2,8 +2,15 @@
 
 // Import Sentry early to help it initialize.
 import {configureScope} from './utils/sentry';
-
 import './polyfills.js';
+
+// Init our dayjs plugins
+import dayjs from 'dayjs';
+import relativeTimePlugin from 'dayjs/plugin/relativeTime';
+dayjs.extend(relativeTimePlugin);
```
whimsicallyson: I don't see this used anywhere?

GregWeil: It adds the dayjs().fromNow() stuff used in team-analytics-project-details

whimsicallyson: good to know!

GregWeil: I'll toss a comment there and mention that it's needed

GregWeil:
About 15kb minified, not a huge difference but it is an improvement

PR #185, opened 2018-12-07
Petite frost: use nestedPopover for email login

whimsicallyson:
https://www.notion.so/glitch/add-email-login-header-dialog-e93b8fe045974512ad71b712f9ead315

https://petite-frost.glitch.me (still behind a feature switch)

And I think I screwed something up by trying to reuse a remix. I only personally changed anything in `src/presenters/pop-overs/sign-in-pop.jsx`


GregWeil:
In the future, reuse branches by running `git pull && refresh` in the console

ttseng:
I think we should change the wording to "Sign in with email" for the button, to be parallel with the other 2 sign in options.

I'm also getting an error when I try requesting an email login twice.  Do you think we should be showing the Almost Done if the same email has been entered multiple times, but the user has not clicked the link?

whimsicallyson:
> I'm also getting an error when I try requesting an email login twice. Do you think we should be showing the Almost Done if the same email has been entered multiple times, but the user has not clicked the link?

based on the error code (429) in the console, I think we heavily rate-limit email requests to something like one every few minutes. I'm not sure how to go about changing that, or if we'd want to.

src/presenters/pop-overs/sign-in-pop.jsx
```diff
@@ -8,6 +8,8 @@ import LocalStorage from '../includes/local-storage';
 import PopoverContainer from './popover-container';
 import {DevToggles} from '../includes/dev-toggles';
 import {captureException} from '../../utils/sentry';
+import NestedPopover from './popover-nested.jsx';
+import {NestedPopoverTitle} from './popover-nested.jsx';
```
judeallred: in general the 'from' items on the right should be distinct;  to combine the items on the left, do:
`import {NestedPopover, NestedPopoverTitle} from ...`, on the left, and make sure there's an "export" on the NestedPopover item.

(you can also do ` import NestedPopover, {NestedPopoverTitle} from ...`, but at that point there's not really a good reason not to export NestedPopover directly as well. )

src/presenters/pop-overs/sign-in-pop.jsx
```diff
@@ -62,39 +64,47 @@ class EmailHandler extends React.Component {
...
-        }
-        
-      </section>
+      <dialog className="pop-over sign-in-pop">
+        <NestedPopoverTitle>
+          Email Signin <span className="emoji email" />
+        </NestedPopoverTitle>
+        <section className="pop-over-actions first-section">
+          {!this.state.done &&
+            <form onSubmit={(e) => this.onSubmit(e)} style={{marginBottom: 0}}>
```
judeallred: I believe this can be written more simply as `<form onSubmit={this.onSubmit} ...`

src/presenters/pop-overs/sign-in-pop.jsx
```diff
@@ -62,39 +64,47 @@ class EmailHandler extends React.Component {
...
+              Sign in with email
+              <input value={this.state.email} onChange={this.onChange} className="pop-over-input" type="email" placeholder="new@user.com"></input>
```
judeallred: In other areas we say "Sign in with Email";  should this text have that same casing or is the context here different enough that lower case is better?

whimsicallyson: its more of a paragraph than a header/button, so I think it works as lowercase: 
screen shot 2018-12-10 at 2 00 58 pm

src/presenters/pop-overs/sign-in-pop.jsx
whimsicallyson: I figured there was a reason it was set up the way it was (and therefore to leave it as-is), but I'll update!

src/presenters/pop-overs/sign-in-pop.jsx
```diff
@@ -62,39 +64,47 @@ class EmailHandler extends React.Component {
...
+            </>
+          }       
+        </section>
+      </dialog>
     );
   }
 }
 
-const SignInPopWithoutRouter = ({header, prompt, api, location, hash}) => (
+const EmailSignInButton = (props) => (
```
judeallred: Seems like this component is missing its proptypes;  what I'd expect to see would be to use the {destructuring} notation to extract the prop you need, and to assert that prop with proptypes, e.g.:
`const EmailSignInButton = ({onClick}) => (`
and
```
EmailSignInButton.PropTypes = {
    onClick: PropTypes.func.isRequired
}
``

File: src/presenters/pop-overs/sign-in-pop.jsx [hunk 5]
```diff
@@ -62,39 +64,47 @@ class EmailHandler extends React.Component {
...
+  <button className="button button-small button-link has-emoji" onClick={() => {props.onClick();}}>
```
judeallred: since onClick is a function and we're not modifying it inline here, we should be able to pass onClick directly (`onClick={props.onClick}`) instead of wrapping it in a closure

src/presenters/pop-overs/sign-in-pop.jsx
judeallred: for the most part, if a module has "a thing" that is the overwhelming reason to import that module, it's the default item and it saves us typing some curlies,  and then any additional exports are special cases.  But once we're slurping in a few things, using curlies for all of them feels a bit cleaner to me

src/presenters/pop-overs/sign-in-pop.jsx
judeallred: Got it, seems fine, although I think the title should be "Email Sign In"

judeallred:
functionally lgtm;  had some code cleanup comments above ^^ 👍

src/presenters/pop-overs/sign-in-pop.jsx
whimsicallyson: cross-referenced with Tiff's mockup and she didn't have that line at all, so I removed it entirely. Updated the header, as well as made the main button  have a capital I.

GregWeil:
Clicking the user options pop while signed out makes the page explode, the error sounds like a broken import

whimsicallyson:
> Clicking the user options pop while signed out makes the page explode, the error sounds like a broken import

changed all `import NestedPopover`s to `import {NestedPopover}`s

src/presenters/pop-overs/sign-in-pop.jsx
```diff
@@ -138,7 +149,10 @@ export default function SignInPopContainer(props) {
       {({togglePopover, visible}) => (
         <div className="button-wrap">
           <button className="button button-small" onClick={togglePopover}>Sign in</button>
-          {visible && <SignInPop {...props}/>}
+          {visible && (
+            <NestedPopover alternateContent={() => <EmailHandler {...props}/>} startAlternateVisible={false}>
+              {showEmailLogin => <SignInPop {...props} {...{togglePopover, showEmailLogin}}/>}
+            </NestedPopover>)}
```
GregWeil: The nested popover reference should exist as part of the sign in pop, not given to it by the sign in pop button, because the sign in pop is used in a few different places and not always with the button. The user pop -> sign in to create teams link gets an error because the pop needs all of the nested popover showEmailLogin stuff.

whimsicallyson: ok this should be fixed (seems to work with both the sign in button and user pop -> sign in). I feel like what I had to do to pass the props along to `EmailHandler` is messy and wrong though? (ln 119)

GregWeil: It would probably feel better in the opposite direction. `(props) => (` and `const {a,b,c} = props` so they only get listed out once.

judeallred:
note: this will need to merge with master, which has removed the dev toggle

src/presenters/pop-overs/sign-in-pop.jsx
whimsicallyson: agreed, that does feel better.

PR #194, opened 2018-12-18
Superb guide: Refactor webpack to use the dev server

GregWeil:
This splits up our webpack config a little bit between dev and production. Production works the same way (`webpack --watch` running in parallel) but in dev it uses the webpack-dev-middleware. That identifies requests for files controlled by webpack and stalls the request until the current build is complete. This makes working on the site a bit nicer.

Unfortunately, the plugin doesn't play nice with our filename querystring cachebuster, because it somehow writes files to the disk with the querystring intact, which breaks other stuff. Instead, I've switched to outputting the files with a clean name and using the stats.json (pared down to the bits we need, otherwise it's a massive file) to identify what's needed by the site.

Advantages
- Now you can just hit refresh in the browser without worrying about waiting until the build finishes (if there's an error you'll still get the old site though)
- Builds with errors or warnings show up red in the console now
- Webpack uses less memory in dev because it shares the node runtime

Disadvantages
- The site is very slow before the initial build finishes (only in dev, prod is unchanged)
- Divergence between dev and prod, but all of the differences are in one file


judeallred:
Seems worth trying to me.  I think it's been inevitable that we'd reach a size that it'd be worth specializing between our dev and prod environments more, even at some minor cost in testing fidelity.

Code looks fine, but it's hard for me to really comment on it because it's mostly webpack config hijinks :-p

The main concern I'd have is that the production environment is still amicable towards 'live coding',  specifically for the changes to /curated/ that our marketing team makes.  I'd want to be confident that we keep our old behavior of not pushing broken webpack builds out to the client or floating build errors to the client in prod.

If the behavior is changed, I'd suggest trying to find the tradeoffs between the advantages of this versus the advantages of being able to be "glitchy" about how our live app can be edited.  

Also if the webpack output style has changed, we may need to also update our CI runner to understand the errors.

Once you've considered the above, I trust you to make a call on deployment and communicating whatever you need to to Gareth and co :-)

jennschiffer:
with the divergence of prod and dev, we should document in the contributing doc that the environment can be set in .env - as well as comment in this PR how the developer workflow would otherwise change bc of this

GregWeil:
@jennschiffer I added some stuff to the faq, plus some updates to outdated text

Developer workflow should be unchanged. Remixes should still work as-is, with the same console output for errors (but now red in dev mode)

PR #195, opened 2018-12-19
Thankful agenda: couple bug fixes

whimsicallyson:
Project descriptions getting cut off:
https://www.notion.so/glitch/Project-description-getting-cut-off-fcebde1215d64cd3a1439756ad75e51d
decreased `truncatedmarkdown` to 80 characters to fit in the smallest location of project descriptions (https://thankful-agenda.glitch.me/community-picks)

Long email addresses:
https://www.notion.so/glitch/long-email-addresses-can-word-wrap-poorly-and-overflow-their-dialog-09eb1d875e314e599052c8e5695d71ac
added `word-break: break-word` style so that long addresses will break on @ and fit in the popover
screen shot 2018-12-19 at 11 17 23 am

### DIFF
--- a/src/presenters/pop-overs/sign-in-pop.jsx
+++ b/src/presenters/pop-overs/sign-in-pop.jsx
@@ -80,7 +80,7 @@ class EmailHandler extends React.Component {
           {(this.state.done && !this.state.error) &&
             <>
               <div className="notification notifySuccess">Almost Done</div>
-              <div style={{wordBreak: 'break-word'}}>Please click the confirmation link sent to {this.state.email}.</div>
+              <div>Please click the confirmation link sent to {this.state.email}.</div>
             </>
           }
           {(this.state.done && this.state.error) &&
--- a/src/presenters/pop-overs/sign-in-pop.jsx
+++ b/src/presenters/pop-overs/sign-in-pop.jsx
@@ -80,7 +80,7 @@ class EmailHandler extends React.Component {
           {(this.state.done && !this.state.error) &&
             <>
               <div className="notification notifySuccess">Almost Done</div>
-              <div>Please click the confirmation link sent to {this.state.email}.</div>
+              <div style={{wordBreak: 'break-word'}}>Please click the confirmation link sent to {this.state.email}.</div>
             </>
           }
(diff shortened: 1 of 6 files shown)


src/presenters/pop-overs/sign-in-pop.jsx
```diff
@@ -80,7 +80,7 @@ class EmailHandler extends React.Component {
           {(this.state.done && !this.state.error) &&
             <>
               <div className="notification notifySuccess">Almost Done</div>
-              <div>Please click the confirmation link sent to {this.state.email}.</div>
+              <div style={{wordBreak: 'break-word'}}>Please click the confirmation link sent to {this.state.email}.</div>
```
jennschiffer: functionality-wise, this is great and works. i just wanted to know before i sign off on this if we have a standard for css - like why are some inline but other elements have classes?

whimsicallyson: if we have one, I'm not aware of it. I've seen both inline and class-based styles throughout the community site, and haven't noticed a pattern to the choices (though I'll admit I haven't dug into it much). I did this one inline because there wasn't currently any classnames to add styling to, so it seemed simpler to just add it inline than add a class and associated style.

jennschiffer: ok cool i'll look into that more. not a blocker here!

ttseng: In this case, I think it would be helpful for it not to be an inline style because it would be useful elsewhere (for other pop-overs that have long messages).  maybe we can just add wordBreak to divs in pop-over sections?

whimsicallyson: that makes sense to me tiff, will update!

styles/pop-overs.styl
```diff
@@ -49,11 +49,14 @@ details.popover-container
       border-bottom: none
       border-bottom-left-radius: 3px
       border-bottom-right-radius: 3px
+    div
```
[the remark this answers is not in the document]
ttseng: yup

PR #196, opened 2018-12-19
Interesting riddle two - Report Abuse

clottman:
Change in collection-item.jsx: makes the collection-item keyboard navigable/screenreader discoverable by restricting the Link to just the button parts. 

Component-izes the way we typically use popovers: https://glitch.com/edit/#!/interesting-riddle-two?path=src/presenters/pop-overs/popover-with-button.jsx:1:0 
 - Updates all instances of that pattern to use the new component

Adds a report abuse button that opens a form that looks like this: 
image

image

with error handling wow! 
image

And a friendly message when the report has been submitted

Relies on https://glitch.com/~support-poster which is a private Community team project whose sole job is posting stuff to our support discourse. (Code review welcome on that app, too; just send me a slack message)


src/presenters/collection-item.jsx
```diff
@@ -60,10 +60,12 @@ class CollectionItem extends React.Component{
                     </div>
                   </div>
                   <div className="collection-name-description">
-                    <div className="button">
-                      <span className="project-badge private-project-badge" aria-label="private"></span>
-                      <div className="project-name">{collection.name}</div>
-                    </div>
+                    <Link to={`/@${userLogin}/${collection.url}`}>
```
jennschiffer: `Link` does not seem to be imported on this file even though it's called here. I get an error when I go to my or gareth's profile so I assume it's any profile with a collection (try https://interesting-riddle-two.glitch.me/@jennschiffer)

pketh:
looks like this introduces a layout regression in the project page
screen shot 2018-12-19 at 2 49 10 pm

src/presenters/includes/add-team-project.jsx
```diff
@@ -1,27 +1,24 @@
-import React from 'react';
-import PropTypes from 'prop-types';
+import React from "react";
```
jennschiffer: I'm seeing single quotes being replaced with double quotes in different parts of the PR, I'd update them back to single quotes for formatting consistency.

src/presenters/collection-item.jsx
clottman: I think something went funky when I merged master, this was working before.

jennschiffer: Ok my profile works now. The issue that's exposed now is that the cursor over the whole collection card is of a pointer, but only the button is clickable. There's a link in the footer of the card "view ... ->" that should be clickable as well. We should probably also have the project links be clickable too since they look like they are.

clottman: The previous behavior was that the whole thing was clickable, but took you to the collection page instead of the thing it seemed you were clicking on (an individual project or the 'view more' button). Do we want to enable clicking through to projects and also make the view button go to the collection page?

pketh:
i was able to submit the form and get a success message by just putting in my email. The message input should check to see whether it's been changed on submit and present a field-error to the user if it hasn't

src/presenters/pop-overs/report-abuse-pop.jsx
```diff
@@ -0,0 +1,308 @@
...
+
+  renderSuccess() {
+    return this.renderSent(
+      <>
+        Report Sent{" "}
+        <span role="img" aria-label="">
+          📧
+        </span>
+      </>,
+      <p className="pop-description">
```
pketh: should use `.info-description` instead to correctly space the text inside the section (removes the extra margin)

clottman:
> i was able to submit the form and get a success message by just putting in my email. The message input should check to see whether it's been changed on submit and present a field-error to the user if it hasn't 

That's a consequence of making the label for the textarea the starting text value of the textarea. Initially we were just checking if the user entered anything at all. So we want to check if they've entered anything other than the placeholder actually?

styles/pop-overs.styl
```diff
@@ -430,6 +469,12 @@ details.popover-container
     right: 0
     top: 27px
     display: inline-block
+    &.top-right
```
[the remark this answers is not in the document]
clottman: I'm not scrolling the viewport - the browser just does that I guess?

pketh:
feedback for the signed out state: 
signed out 2x

pketh:
ui 
signed in 2x
feedback for the signed in state

pketh:
ui feedback for the error state
error state 2x

pketh:
ui feedback for the success state
complete 2x

pketh:
it's a lot of feedback, def feel free to ask me for help, or irl clarification if you want it

PR #197, opened 2018-12-19
Interesting riddle 2 2

clottman:
NEW PR for same code! Glitch project at glitch.com/~interesting-riddle-two

Change in collection-item.jsx: makes the collection-item keyboard navigable/screenreader discoverable by restricting the Link to just the button parts. 

Component-izes the way we typically use popovers: https://glitch.com/edit/#!/interesting-riddle-two?path=src/presenters/pop-overs/popover-with-button.jsx:1:0 
 - Updates all instances of that pattern to use the new component

Adds a report abuse button that opens a form that looks like this: 
image

image

with error handling wow! 
image

And a friendly message when the report has been submitted

Relies on https://glitch.com/~support-poster which is a private Community team project whose sole job is posting stuff to our support discourse. (Code review welcome on that app, too; just send me a slack message)


src/presenters/pop-overs/project-options-pop.jsx
```diff
@@ -1,119 +1,181 @@
...
-  </button>
-);
-
 // Project Options Content
-const ProjectOptionsContent = ({addToCollectionPopover, ...props}) => {
+const ProjectOptionsContent = ({ addToCollectionPopover, ...props }) => {
   function animate(event, className, func) {
-    const projectContainer = event.target.closest('li');
-    projectContainer.addEventListener('animationend', func, {once: true});
+    const projectContainer = event.target.closest("li");
```
jennschiffer: some formatting regression (quotes & space between brackets)

src/presenters/pop-overs/project-options-pop.jsx
```diff
@@ -1,119 +1,181 @@
...
+    projectContainer.addEventListener("animationend", func, { once: true });
     projectContainer.classList.add(className);
-    props.togglePopover();
   }
-  
+
   function leaveProject(event) {
-    const prompt = `Once you leave this project, you'll lose access to it unless someone else invites you back. \n\n Are sure you want to leave ${props.project.name}?`;
+    const prompt = `Once you leave this project, you'll lose access to it unless someone else invites you back. \n\n Are sure you want to leave ${
+      props.project.name
```
jennschiffer: formatting fix: get the template literal all on the same line together again

src/presenters/pop-overs/project-options-pop.jsx
```diff
@@ -1,119 +1,181 @@
...
-  
+
   function joinTeamProject() {
     props.joinTeamProject(props.project.id, props.currentUser);
   }
-    
+
   function animateThenAddPin(event) {
-    animate(event, 'slide-up', () => props.addPin(props.project.id));
+    animate(event, "slide-up", () => props.addPin(props.project.id));
```
jennschiffer: formatting fix: double quote to single quote on lines 36, 40, 44

src/presenters/pop-overs/report-abuse-pop.jsx
```diff
@@ -0,0 +1,292 @@
...
+    if (this.state[stateField] === "") {
+      errorObj = { [errorField]: `${fieldDescription} is required` };
+    } else {
+      errorObj = { [errorField]: "" };
+    }
+    this.setState(errorObj);
+    return errorObj;
+  }
+
+  validateEmail() {
```
jennschiffer: i would have a validateMessage function that not only checks that there is a message (calling validateNotEmpty) but checks that it's not identical to to the default this.state.reason. this way i can't keep sending "this project doesn't belong on glitch because..." messages

jennschiffer:
i also wanted to bring back up how the collections hover state has a pointing cursor but the footer "view more" link isn't clickable and should be.

ttseng:
Some smaller design issues:

screen shot 2018-12-28 at 12 15 39 pm
Gap between editable field input and error messaging.

image
Can we change the styling of the confirmation message here to `pop-over-action` rather than `pop-over-info`? (so the background is white rather than grey)

clottman:
@ttseng made both of those changes! Anything else you've noticed that hasn't been addressed yet?

src/presenters/pop-overs/add-collection-project-pop.jsx
```diff
@@ -266,7 +266,7 @@ AddCollectionProjectPop.propTypes = {
   api: PropTypes.func.isRequired,
   collection: PropTypes.object.isRequired,
   addProjectToCollection: PropTypes.func.isRequired,
-  togglePopover: PropTypes.func.isRequired,
+  togglePopover: PropTypes.func, // required but added dynamically
```
jennschiffer: can you elaborate on this change - what happens if you set it to isRequired?

clottman: It gives an error that you're not passing the prop in and it's required. :(

clottman:
@ttseng  
### Ways to trigger various error states: 

1. Type nothing
image

2. don't change default text 
image

3. don't provide email when logged out
image

4. provide invalid email
image

5. Modify https://glitch.com/~support-poster to return a 500 immediately from `/post` 
```
app.post("/post", function(request, response) {
  response.sendStatus(500); // add this here
```
image

ttseng:
I am noticing another styling regression.  Looks like the margins of text in pop-overs got edited, so now the rows are taller than they should be.

I think it's because of this styling that was added:
```
.project-page .buttons.space-between div > * {
    margin: 3.5px;
}
```
image

(regression shown on the left)

jennschiffer:
everything looks good to me :shipit:

PR #198, opened 2018-12-20
Nova chance: Lots of analytics events

GregWeil:
https://www.notion.so/glitch/Add-new-community-tracking-events-9f6cf10373ea47e38750aa06c886acf1
Check the network tab in the debugger to see the new requests to segment.io and what's being sent

The new code is all in presenters/analytics.js, with lots of usage and indentation changes throughout the rest of the code.

First I have TrackClick, which injects an onClick prop into its children. It hides what's happening when you use it, which is a lot more readable than `<Tracker>{track=><asdf onClick={()=>{origOnClick(); track()}}/>}</Track>`. The downside is it won't work if the thing in it ignores onClick props.
AnalyticsContext does a shallow merge of whatever you put in it with the other contexts in the page, which gets sent with the tracking event for TrackClick.
TrackedExternalLink exists because when you click a link to outside the community site it triggers a full page load that will block the request from a TrackClick. It'll wait a moment before firing off the page load so the tracking request can go through.

I'm using the contexts so everything on a given page gets an origin property for the page type (extra relevant for user vs team since you can't differentiate by url)

@garethx This should have everything you were asking for, with the exception of groupId on team creation and page events.


GregWeil:
No, I usually will group things together if they're all closely related in code (for example react context stuff) or will all be used together in a page (like with team-elements.jsx).

garethx:
lgtm - merge!

PR #199, opened 2018-12-21
Rift-aftershave: Only show leave project button if user is actually on the project

clottman:
See discussion in slack at: https://fogcreek.slack.com/archives/CB4EHAJ2J/p1545432703006800


src/presenters/pop-overs/project-options-pop.jsx
```diff
@@ -83,7 +83,7 @@ const ProjectOptionsContent = ({addToCollectionPopover, ...props}) => {
         </section>
       }
       
-      {(props.leaveProject && props.project.users.length > 1) &&
+      {(props.leaveProject && props.currentUserIsOnProject) &&
```
jennschiffer: another good catch - but I do think we should keep the `props.project.users.length > 1` expression in there for projects that don't have a team to fall back to if everyone leaves.

clottman: Ah - so we can't have orphaned projects?

jennschiffer: We'd like to avoid them, yeah. In the case of a project part of a team, if everyone leaves, the project still has the team as a member.

PR #200, opened 2018-12-21
Wheat-earwig: Fixed wrong project name in leave project dialog

clottman:
Before: 
image

After:  
image

See: https://wheat-earwig.glitch.me/ .


clottman:
I did a quick search for all the places we use 'name' as well as project.name more specifically, and it doesn't look like we're doing it anywhere else.

PR #203, opened 2019-01-04
Featured zine

ttseng:
Featured Zine section on homepage
image

**Remix:** 
https://featured-zine.glitch.me/

**Spec:**
https://www.notion.so/glitch/Featured-Embed-Updates-bef80fc8d5ea468583bfca1056f5e4e7


src/presenters/featured-embed.jsx
```diff
@@ -0,0 +1,39 @@
...
+        <img className={'mask ' + feature.mask} src={feature.image} alt=""/>
+      </Link>
+    </div>
+      
+    <div className="content" style={{backgroundColor: feature.color}}>
+      <div className="description">
+        <Link to={`culture${feature.url}`}>
+          <h1>{feature.title}</h1>
+        </Link>
+        <p dangerouslySetInnerHTML={{__html: feature.body}}/>
```
clottman: Why do you need `dangerouslySetInnerHTML` here instead of doing `<p>{{feature.body}}</p>`?

src/presenters/featured-embed.jsx
```diff
@@ -0,0 +1,39 @@
...
+        <Link to={`culture${feature.url}`} className="learn-more">
+            Learn More →
+        </Link>
+      </div>
+      <div className="embed">
+        <span dangerouslySetInnerHTML={{__html: feature.embed}}/>
```
clottman: Same question here around dangerouslySetInnerHTML.

src/presenters/featured.jsx
```diff
@@ -2,17 +2,19 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import {sampleSize} from 'lodash';
 
-import EmbedHtml from '../curated/embed';
+// import EmbedHtml from '../curated/embed';
```
clottman: Commented code intended to be left here?

src/presenters/featured-embed.jsx
ttseng: some info on `dangerouslySetInnerHtml` here: https://reactjs.org/docs/dom-elements.html#dangerouslysetinnerhtml

src/presenters/featured.jsx
ttseng: Yeah, I'm not sure if we want to keep that in to use the full-width embed in the future (I think we might switch between the current embed and the feature one, in case there isn't anything to heavily feature).  Not sure though?

src/presenters/featured-embed.jsx
clottman: I know what it does - why do you need to set inner html instead of `children`?

src/presenters/featured.jsx
clottman: My preference would be to take it out until we need it again, especially in this case where not having it will throw an error that it needs to be imported if it's being used.

src/presenters/featured-embed.jsx
clottman: Ah, I see. It looks like feature.body and feature.embed are defined as strings. Was that intentional? Since we use jsx, they can just be defined like this and they should get rendered as html. 
```
var embed = <div class="glitch-embed-wrap" style="height: 420px; width: 100%;">
  <iframe
    allow="geolocation; microphone; camera; midi; encrypted-media"
    src="https://glitch.com/embed/#!/embed/egg-drop?path=README.md&previewSize=100"
    alt="egg-drop on Glitch"
    style="height: 100%; width: 100%; border: 0;">
  </iframe>
</div>;

var body = <>
<p>Follow along with Mythbusters Jr. by remixing your own Glitch apps inspired by each episode.</p>
<p><b>Episode 1:</b></p>
<p>Can you make a viable, life-saving parachute out of duct tape?</p></>;
```
 
Note that the thing you save to a variable must be wrapped in a single element, so in `body` I'm using the React.Fragment shorthand `<> </>`.

ttseng: Oh, that makes sense to me.  

I was basing it off of `curated/embed.js`, which I think is a string because it's easiest to just copy and paste the default embed code from the editor, rather than strip out the `Copy and Paste Me` message:

```
export default `
<!-- Copy and Paste Me -->
<div class="glitch-embed-wrap" style="height: 420px; width: 100%;">
  <iframe
    allow="geolocation; microphone; camera; midi; encrypted-media"
    src="https://glitch.com/embed/#!/embed/click-fireworks?path=firework.js&previewSize=100"
    alt="click-fireworks on Glitch"
    style="height: 100%; width: 100%; border: 0;">
  </iframe>
</div>
`;
```

I don't have a strong preference either way - I figured this was how Maurice and Gareth were used to editing the featured content.  Do you have any thoughts @jennschiffer ?

jennschiffer: i'd go with the dangerous route - it's how we do the feature embed as tiff mentioned (see: https://github.com/FogCreek/Glitch-Community/blob/master/src/presenters/featured.jsx#L66) - and until we have an app for marketing/media to use to update this, i don't want to change their copy/paste workflow

src/presenters/featured.jsx
jennschiffer: +1 to removing the comment

ttseng: ok, I removed it

PR #206, opened 2019-01-09
Beneficial moose

TimKingtonFC:
This is for private apps.  We need a page hosted at glitch.com that uses the user's token to generate a private app token.

### DIFF
--- a/public/api-auth.html
+++ b/public/api-auth.html
@@ -1,0 +1,1 @@
+Hi there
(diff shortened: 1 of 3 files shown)


GregWeil:
`PROJECT_DOMAIN` and `ENVIRONMENT` aren't used by the template, can you remove them?
Other than that it looks good

TimKingtonFC:
This is ready to go now.  I added handling for unauthorized users.  You'll still get a 404 on that API endpoint right now, but that's okay.

PR #208, opened 2019-01-09
Astonishing orchestra: use code to sign in

whimsicallyson:
This is, at the very least, the UI for signing in with the code provided in an email. currently I'm seeing CORS errors when trying to actually use a code... not sure if the issue is me, or the fact that its a remix.

https://glitch.manuscript.com/f/cases/3326594/login-update-email-sign-in-with-manual-sign-in-code-support-and-ui-to-match-editor

https://astonishing-orchestra.glitch.me/

### DIFF
--- a/src/presenters/pop-overs/sign-in-pop.jsx
+++ b/src/presenters/pop-overs/sign-in-pop.jsx
@@ -7,6 +7,7 @@ import Link from '../includes/link';
 import LocalStorage from '../includes/local-storage';
 import PopoverWithButton from './popover-with-button';
 import {captureException} from '../../utils/sentry';
+import {CurrentUserConsumer} from '../current-user';
 import {NestedPopover, NestedPopoverTitle} from './popover-nested.jsx';
 
 /* global GITHUB_CLIENT_ID, FACEBOOK_CLIENT_ID, APP_URL */
--- a/src/presenters/pop-overs/sign-in-pop.jsx
+++ b/src/presenters/pop-overs/sign-in-pop.jsx
@@ -120,7 +120,7 @@ class SignInCodeHandler extends React.Component {
     console.log(this.props);
     try {
       const {data} = await this.props.api.post('/auth/email/' + this.state.code);
-      this.props.login(data);
+      this.login(data);
       this.setState({error: false});
     } catch (error) {
       captureException(error);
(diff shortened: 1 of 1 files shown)


src/presenters/pop-overs/sign-in-pop.jsx
```diff
@@ -115,19 +179,31 @@ const SignInPopWithoutRouter = (props) => (
           hash: hash,
         },
       });
+      <CurrentUserConsumer>
+        {(currentUser, fetched, {login}) => <LoginPage setUser={login} destination={destination} setDestination={setDestination} {...props}/>}
+      </CurrentUserConsumer>;
```
GregWeil: The LoginPage is only meant to be used from the router, it'll take over the current page and redirect you somewhere else. You need to grab the login function that the LoginPage gets from the CurrentUserConsumer

src/presenters/pop-overs/sign-in-pop.jsx
```diff
@@ -95,6 +97,68 @@ class EmailHandler extends React.Component {
...
+  onChange(e) {
+    this.setState({code: e.target.value});
+
+  }
+  
+  async onSubmit(e) {
+    e.preventDefault();
+    this.setState({done: true});
+    try {
+      await this.props.api.post('/login/email/?token=' + this.state.code);
```
GregWeil: This is where the login actually happens, so this is where that function from CurrentUserConsumer should get called.
```
const {data} = await this.props.api.post(...);
this.props.login(data);
```

src/presenters/pop-overs/sign-in-pop.jsx
```diff
@@ -95,6 +96,69 @@ class EmailHandler extends React.Component {
...
+  onChange(e) {
+    this.setState({code: e.target.value});
+
+  }
+  
+  async onSubmit(e) {
+    e.preventDefault();
+    this.setState({done: true});
+    try {
+      const {data} = await this.props.api.post('/login/email/?token=' + this.state.code);
```
GregWeil: I think the issue is that your url here is wrong, this is the glitch.com url that the email sends you to, the api uses /auth/email/{code} (check out /login.jsx line 125)

whimsicallyson: good catch, thanks!

whimsicallyson:
it works now! I feel confident about this!

GregWeil:
It looks like the almost done/error in the email sent message is pulling some unwanted notification animations, it fades out unless you mouseover it. That's an issue on glitch.com too, so it doesn't have to block this review

PR #213, opened 2019-01-11
Report button throughout the community site

clottman:
Spec: https://www.notion.so/glitch/Expand-report-abuse-button-location-to-throughout-the-community-site-bda54e53cc3d4e0b97b4b17c8883a688 

Remix: https://gc-report.glitch.me/
See: 
- [a user page](https://gc-report.glitch.me/@cassey)
- your own user page when logged in
- [a collection](https://gc-report.glitch.me/@cassey/cool-stuff-i-made)
- one of your own collections when logged in
- [the bottom of the home page](https://gc-report.glitch.me/)
- [a team](https://gc-report.glitch.me/@community) 
   - Since you can't currently remove yourself from a team, you can see the report abuse button from a team that you are part of as well as one you're not on


src/models/project.js
```diff
@@ -6,7 +6,7 @@ export function getAvatarUrl(id, cdnUrl=CDN_URL) {
   return `${cdnUrl}/project-avatar/${id}.png`;
 }
 
-export function getLink(domain) {
+export function getLink({domain}) {
```
clottman: Reason for this change - standardizes how we call `getLink` - it always takes the model as a parameter, for each model type.

src/presenters/pages/project.jsx
```diff
@@ -91,19 +91,17 @@ ReadmeLoader.propTypes = {
 };
 
 const ProjectPage = ({
-  project: {
-    description, domain, users, teams,
-    ...project // 'private' can't be used as a variable name
-  },
+  project,
```
clottman: Reason for this change: the way we were using this before, the variable `project` in the function body contained just the properties that weren't split up. So it appears at first glance like you can access `project.domain` later on, but actually `project.domain` is undefined. 

The ReportButton needs the entire `project` object, including the domain. To preserve the way we were using `domain, users, teams` before, those props are assigned from `project` now on line 103.

styles/pages/team.styl
```diff
@@ -1,50 +0,0 @@
-.team-page
```
clottman: This is not referenced anywhere/hasn't been getting served so it's definitely not in use.

src/presenters/pages/team.jsx
```diff
@@ -241,6 +242,8 @@ class TeamPage extends React.Component {
             users={this.props.team.users}
           />
         )}
+        
+        <ReportButton thingType="team" thingModel={this.props.team} />
```
jennschiffer: can we rename these to reporteeType and reporteeModel or something with context as to what these things are - i want us to avoid using "thing" because it's very vague

jennschiffer: reportedType/reportedModel i like better even

ttseng:
@clottman Can we add a `<Loader/>` so that when the user clicks "Submit Report" they know that something is happening before they see the confirmation page?  You can check out the `CreateCollectionButton` example in `collections-list.jsx`

loading-animation

ttseng:
I agree that team members should not see a "report abuse" button on their own teams.  

Similarly, if you are viewing a team collection and are a member of that team, you shouldn't see a report abuse button for the collection.

clottman:
> Similarly, if you are viewing a team collection and are a member of that team, you shouldn't see a report abuse button for the collection.

This should be handled by the `isAuthorized` check on the collection, right?

src/presenters/pages/team.jsx
clottman: Ooh I like reportedType! I couldn't think of a better word for that.

ttseng:
> > Similarly, if you are viewing a team collection and are a member of that team, you shouldn't see a report abuse button for the collection.
> 
> This should be handled by the `isAuthorized` check on the collection, right?

yup

clottman:
Addressed a few comments: 
- Users shouldn't see the report abuse button on a team they're part of
- Added the eslint rule for switch/case that was discussed in #community on slack and updated models.js to use it
- renamed the thing @jennschiffer asked for.

Currently working on adding the loading indicator.

ttseng:
loader looks great! thanks for adding this.

I just noticed that category pages (for example https://gc-report.glitch.me/games) don't have report abuse, but I think it's ok because once team collections are in place, they'll turn into team pages and the report abuse button will automatically appear.

PR #216, opened 2019-01-14
Astonishing orchestra

whimsicallyson:
https://astonishing-orchestra.glitch.me/

Now if you request a sign-in email, there's a button on the success state to take you to the sign-in code pop!

screen shot 2019-01-14 at 3 17 50 pm

As noticed/requested by cori, https://glitch.manuscript.com/f/cases/3327667/sign-in-pop-text-after-email-has-been-sent-could-be-clearer

also fixed the fact that the notification banners would disappear (as you can see in the screenshot above)

### DIFF
--- a/src/presenters/pop-overs/sign-in-pop.jsx
+++ b/src/presenters/pop-overs/sign-in-pop.jsx
@@ -95,12 +95,8 @@ class EmailHandler extends React.Component {
               }
             </section>
             {(this.state.done && !this.state.error) &&
-              <section className="pop-over-actions last-sectionn pop-over-info">
-                <button className="button-small button-tertiary button-on-secondary-background"  onClick={() => { onClick(); showCodeLogin(api); }}>
-                  <span>Use a sign in code</span>
-                </button>
-              </section>
-              }
+              <SignInCodeSection onClick={() => { onClick(); showCodeLogin(api); }}/>
+            }
           </dialog>
         }
       </NestedPopover>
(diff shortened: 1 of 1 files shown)


src/presenters/pop-overs/sign-in-pop.jsx
```diff
@@ -171,6 +178,17 @@ EmailSignInButton.propTypes = {
   onClick: PropTypes.func.isRequired
 };
 
+const SignInCodeSection = ({onClick}) => (
+  <section className="pop-over-actions last-section pop-over-info">
+    <button className="button-small button-tertiary button-on-secondary-background"  onClick={() => { onClick();}}>
```
GregWeil: This can just do `onClick={onClick}` instead of defining an extra function

src/presenters/pop-overs/sign-in-pop.jsx
```diff
@@ -67,31 +67,38 @@ class EmailHandler extends React.Component {
...
-          }
-          {(this.state.done && this.state.error) &&
-            <>
-              <div className="notification notifyError">Error</div>
-              <div>Something went wrong, email not sent.</div>
-            </>
-          }       
-        </section>
-      </dialog>
+      <NestedPopover alternateContent={() => <CurrentUserConsumer>{(currentUser, fetched, {login}) => <SignInCodeHandler setUser={login} {...this.props}/>}</CurrentUserConsumer>} startAlternateVisible={false}>
```
GregWeil: Now that this is being done in two places, can you add a variant of SignInCodeHandler with the CurrentUserConsumer already bundled in? Having to grab and plug in the login function inline makes it a bit hard to read.

whimsicallyson: updated!

PR #218, opened 2019-01-15
Add h1 tag to glitch.com for SEO purposes

sheridanvk:
Fixes https://glitch.manuscript.com/f/cases/3327668/Re-add-h1-to-header-of-homepage
Site here for verification: https://jealous-hearing.glitch.me/

Added the h1 tag around the image as requested, and modified the existing h1 tag to h2. According to this SO answer: https://stackoverflow.com/questions/665037/replacing-h1-text-with-a-logo-image-best-method-for-seo-and-accessibility#comment68674842_669847, alt text has the same weighting as regular text  for SEO and is preferable to adding text and hiding it, so I didn't add any additional text in. @garethx, shout if you have concerns with this though.

Also rejigged the CSS classnames on this section to disambiguate from the upcoming featured-collections section, and removed some background colour that was messing with the display, as this is not used elsewhere according to @ttseng.


sheridanvk:
Also added a site verification tag so we can now check that we're crawlable.

PR #219, opened 2019-01-16
Profile feature 2

ttseng:
**Remix:** 
https://profile-feature-2.glitch.me/

image

**Summary:** 
Users can feature a single project in an embed on their profile page

**Original Spec:** 
https://www.notion.so/glitch/Feature-Pinned-Project-in-Embed-on-Profile-Page-8f114bcab6a34a9888f26f012a4215d1
* Note: feature has been modified from spec based on feedback https://fogcreek.slack.com/archives/CB4EHAJ2J/p1547591694327100

Will follow up with featured embed on team pages in a separate PR.


ttseng:
Added "featured project" section header to make stronger connection between the full-width embed and the option to "feature" a project (so you know that you can do that on your own profile page)

src/presenters/entity-page-featured-project.jsx
```diff
@@ -0,0 +1,76 @@
...
+          : reportBtn
+        }
+
+        <div className="buttons buttons-right">
+
+          {currentUser.login && <AddProjectToCollection className="button-small" api={api} currentUser={currentUser} project={featuredProject} fromProject={true} addProjectToCollection={addProjectToCollection}/>}
+          
+          <RemixButton className="button-small"
+            name={featuredProject.domain} isMember={isAuthorized}
+            onClick={() => trackRemix(featuredProject.id, featuredProject.domain)}
```
GregWeil: Can you use TrackClick from analytics.jsx instead of this? You can find an example in add-collection-project-pop.jsx, it takes name and properties as props and will apply an onClick event to its children. You can also omit origin and it'll use the current page.

src/presenters/pages/user.jsx
```diff
@@ -70,65 +72,86 @@ const UserPage = ({
...
-      >
-        <NameAndLogin
-          name={user.name} login={user.login} {...{isAuthorized, updateName}}
-          updateLogin={login => updateLogin(login).then(() => syncPageToLogin(login))}
+}) =>
+{ 
+  const pinnedSet = new Set(user.pins.map(({projectId}) => projectId));
+  const pinnedProjects = user.projects.filter( ({id}) => pinnedSet.has(id)).filter ( ({id}) => id != featuredProjectId); 
+  const recentProjects = user.projects.filter(({id}) => !pinnedSet.has(id)).filter( ({id}) => id != featuredProjectId);
+  const featuredProject = Object.is(featuredProjectId, undefined) ? undefined : user.projects.filter( ({id}) => id == featuredProjectId)[0];
```
GregWeil: You can just use `user.projects.find(({id}) => id === featuredProjectId)` here, it'll get the first one that matches or undefined

src/presenters/pop-overs/project-options-pop.jsx
```diff
@@ -51,17 +51,24 @@ const ProjectOptionsContent = ({addToCollectionPopover, ...props}) => {
...
         <section className="pop-over-actions">
+          {!props.project.private && <PopoverButton onClick={featureProject} text="Feature" emoji="clapper"/>}
           <TrackClick name="Project Pinned">
             <PopoverButton onClick={animateThenAddPin} text="Pin " emoji="pushpin"/>
           </TrackClick>
         </section>
       }
       {!!props.removePin &&
         <section className="pop-over-actions">
+          {!props.project.private && <PopoverButton onClick={featureProject} text="Feature" emoji="arrow-up"/>}
```
GregWeil: Are these two supposed to have different icons?

src/presenters/entity-page-pinned-projects.jsx
```diff
@@ -22,15 +14,16 @@ const EntityPagePinnedProjects = ({api, projects, pins, currentUser, isAuthorize
...
-      {!!pinnedVisible && !!pinnedProjects.length && (
+     {projects.length > 0 && 
         <ProjectsList title={pinnedTitle}
-          projects={pinnedProjects}
+          projects={projects}
           api={api} 
           projectOptions={isAuthorized ? {removePin, ...projectOptions} 
             : (currentUser && currentUser.login ? {...projectOptions} : {})
           }
+          extraClasses="pinned"
```
GregWeil: Does this pinned class do anything? If not, it seems like EntityPagePinnedProjects and EntityPageRecentProjects could finally get merged back into a single component 🤞

src/presenters/entity-page-featured-project.jsx
ttseng: I changed it to TrackClick.  What's the best way to confirm that it's working?  (All I know is that I don't have app errors, but I don't know how to monitor the results of TrackClick)

src/presenters/pages/user.jsx
ttseng: that's a lot cleaner - thanks!

src/presenters/pop-overs/project-options-pop.jsx
ttseng: good catch - changed it to "clapper"

src/presenters/entity-page-pinned-projects.jsx
ttseng: I think not - it must have been a relic from prototyping.  I removed it and consolidated everything to `entity-page-projects.jsx` (removing `entity-page-pinned-projects.jsx` and `entity-page-recent-projects.jsx`) .

src/presenters/entity-page-featured-project.jsx
GregWeil: Check the network tab in your debugger, there should be xhr requests going out to segment.io

src/presenters/entity-page-projects.jsx
```diff
@@ -0,0 +1,44 @@
...
+  isAuthorized: PropTypes.bool.isRequired,
+  projects: PropTypes.array.isRequired,
+  addPin: PropTypes.func,
+  removePin: PropTypes.func,
+  projectOptions: PropTypes.object,
+};
+
+const EntityPageProjectsContainer = ({api, projects, maybeCurrentUser, ...props}) => (  
+  <EntityPageProjects api={api} projects={projects} currentUser={maybeCurrentUser} {...props}/>  
+);
```
GregWeil: This container isn't doing anything except converting maybeCurrentUser to currentUser (but it's still optional)

ttseng: thanks for pointing this out - I removed the container and pass currentUser directly to EntityPageProjects

ttseng:
Note - planning on incorporating @sheridanvk's changes to embeds.

ttseng:
added new Embed component from @sheridanvk

src/presenters/entity-page-featured-project.jsx
```diff
@@ -0,0 +1,66 @@
...
+    </>
+  );
+  
+  return(
+    <>
+      <section id="featured-project-embed">      
+        <h2>{featuredTitle}</h2>
+      
+        {isAuthorized && <FeaturedProjectOptionsPop api={api} unfeatureProject={unfeatureProject}/>}
+        <div className="glitch-embed-wrap">
```
sheridanvk: this class is already included in the Embed component, so I'd remove it here

sheridanvk: remove the whole div I mean - there's two levels of nesting in this current implementation

ttseng: oh good catch - thanks! I removed the container div.

PR #220, opened 2019-01-16
Profuse tea: Fix issues with cache busting

GregWeil:
I had switched from including the content hash in the webpack config filename to adding it  while rendering the template because the webpack-dev-middleware tries to write the file to disk with the querystring still intact. Lazy loaded js is linked by webpack, so c3 lost its cache buster.

I've got a 'real' fix out, which is for the dev middleware to clip the querystring off of filenames when writing to disk (like webpack normally does) https://github.com/webpack/webpack-dev-middleware/pull/361

In the meantime, I switched back to cache busting in the webpack config and disabled writing to disk in dev mode. That has the side effect that when you switch a server from dev to prod mode, it will show an outdated site until the next prod build finishes.

I also cut the cache time down from a month to a week, because we usually deploy at least that often anyway.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,9 +1,4 @@
-public/*.css
-public/*.css.map
-public/*.ks
-public/*.js.map
-public/*.js
-public/*.json
+public/*
 styles/style.css
 styles/style.css.map
 *.gz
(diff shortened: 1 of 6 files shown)


GregWeil:
Looks like they've already merged in my change! I'll update this to just use the new middleware release

GregWeil:
Updated, this is now the 'real' fix where cache busting happens in webpack filenames.

jennschiffer:
looks good and also congrats on your contribution to web-dev-middleware!

PR #221, opened 2019-01-21
Make Glitch.com show up for the Google crawler

sheridanvk:
Demo available at https://mountainous-pixie.glitch.me/ (note this app was compiled using production mode Webpack)

The major areas of change are:
1. babel config files modified to allow transpiled ES2015 code to be used in older browsers.
2. window.bootstrap in client.jsx will now always get defined (instead of only being made available when ES6 is present) and will be run once the build is done. It still checks for whether we’re trying to show an editor page right now, and switches this out for the editor app if needed (note: editor pages are currently broken for browsers without ES6 support and will continue to be broken after this fix - I will open a case to better support that)
--> as a result, I stripped out html and JS for this no-ES6 state in index.ejs, as we're not checking for it there any more.
3. Added a new component embed.jsx. This is now the only place where we check for ES6 compatibility. If it’s not possible to show the embed, we show an error view. The error part of this component needs to be refactored once we have better error components (currently we only have one-off full-page error components).

Marketing now has to use a slightly different format for specifying feature embeds, providing only the app domain instead of the whole embed. I will confirm that this is not an issue with Gareth.

Note that this was a little complicated to fix because of the way we’re integrating with webpack. Chatted with Emanuele, and we think we can make this simpler down the line, so I’m just fixing this for the crawler issue right now.

Screenshots of the app now on Chromium 41 (Google webcrawler version):

screenshot 2019-01-21 at 5 04 03 pm
screenshot 2019-01-21 at 5 05 47 pm
screenshot 2019-01-21 at 5 06 06 pm
screenshot 2019-01-21 at 5 06 20 pm


sheridanvk:
I need to make some changes to this to address the build size (which is making the CI fail); will send a note once this can be looked at again

src/presenters/includes/embed.jsx
```diff
@@ -0,0 +1,60 @@
...
+      const [b, ...bRest] = y; // Can we use array destructuring?
+      const str = `${b}23`; // Can we use formatted strings?
+      const func = (f, ...args) => f(...args); // Can we define arrow functions?
+      func(async arg => await arg, Promise.resolve()); // Can we do async/await?
+      new URLSearchParams(); // Do we have URLSearchParams? 
+      /* eslint-enable no-unused-vars */
+
+      return true;
+    } catch (error) {
+      console.log("Sorry, you don't have the necessary JS permissions to run Glitch code editors", error);
```
clottman: Is this the right terminology we want? As a user it would make me want to try to find some permission settings I need to toggle, not to update my browser.

sheridanvk: yeah that's a good point. We message what to do in the actual embed itself, so I'll just change the messaging here to be consistent for the console log as well 👍

src/presenters/includes/embed.jsx
```diff
@@ -0,0 +1,67 @@
...
+      /* eslint-disable no-unused-vars */
+      let x = {a: 1, b: 2}; // Can we use let?
+      const y = [1, 2, 3]; // Can we use const?
+      const {a, ...aRest} = x; // Can we use object destructuring?
+      const [b, ...bRest] = y; // Can we use array destructuring?
+      const str = `${b}23`; // Can we use formatted strings?
+      const func = (f, ...args) => f(...args); // Can we define arrow functions?
+      func(async arg => await arg, Promise.resolve()); // Can we do async/await?
+      new URLSearchParams(); // Do we have URLSearchParams? 
+      /* eslint-enable no-unused-vars */
```
GregWeil: Will these checks still work? I'd expect them to get transpiled and work the same as the rest of the site

GregWeil: Just tested, it does block embeds, but only by the URLSearchParams check

GregWeil:
URLSearchParams isn't getting transpiled, so if you're on an older browser you can't login or search. If the only goal is to get seo working that isn't a blocker though

src/presenters/includes/embed.jsx
sheridanvk: hah that is an excellent point :/ I did test this on chromium 41 to make sure it was working as expected, but didn't notice that it was only one of these cases causing it to fail.

The main goal here though is seo support and not so much old versions of JS. So we could just go ahead and ship this, and then do something smarter (which could be properly transpiling the editor so it works in ES2015 too). 

Eyeballing our traffic, much less than 1% of our traffic is using a browser without ES6 support, so I think it's not a huge goal to have that working properly.

PR #222, opened 2019-01-22
Breezy ocelot: Pupdate for sign-in codes

whimsicallyson:
https://breezy-ocelot.glitch.me/

Not much actual code here, more like a tone review :) New Stuff is under the avatar menu!


ttseng:
I think it might be good to talk about email log in more generally, with and without the code, because both are useful.  Logging in from the library is a really good example!  Maybe something like this?

```
title: 
"Use your email to get Glitchy",
    body: `
Don't have Facebook or GitHub, but still want to use Glitch? Want to code at the library without remembering all your logins?
Now you can sign into Glitch with just your email.  Choose Sign in with Email to get a magic link and temporary sign-in code sent straight to your inbox, made just for you. 
![](https://cdn.glitch.com/69bf2d27-a029-4632-86a4-a722974778ae%2FScreen%20Shot%202019-01-22%20at%204.26.50%20PM.png?1548192423838)
`,
  },
 ```
Also, I thought this GIF might be useful for showing both options:
email

(viewable at https://wax-wine.glitch.me/)

whimsicallyson:
@ttseng I like your suggestions! And I definitely need a class in how to GIF...

PR #225, opened 2019-01-28
Superb guide: Featured collections + misc cleanup

GregWeil:

image

More ideas is still just the categories list for now, but the code is in place to load collections from a team and use them instead.

@garethx before this can deploy I need a starting set of featured collections


src/presenters/more-ideas.jsx
```diff
@@ -0,0 +1,64 @@
...
+    {({data}) => (
+      <DataLoader get={() => data !== 'NOT FOUND' ? api.get(`collections?teamId=${data}`) : null}>
+        {({data}) => (
+          <MoreIdeasCollectionsDisplay collections={data.map(collection => ({...collection, team: {url: moreIdeasTeam}}))}/>
+        )}
+      </DataLoader>
+    )}
+  </DataLoader>
+);
+
```
GregWeil: The stuff above here is not used yet

src/presenters/includes/collection-avatar.jsx
```diff
@@ -4,6 +4,19 @@ import PropTypes from 'prop-types';
...
+  if(/^#([A-Fa-f0-9]{3}){1,2}$/.test(hex)){
+    let c = hex.substring(1).split('');
+    if(c.length== 3){
+      c = [c[0], c[0], c[1], c[1], c[2], c[2]];
+    }
+    c = '0x'+c.join('');
+    return 'rgba('+[(c>>16)&255, (c>>8)&255, c&255].join(',')+',0.5)';
+  }
+  return false;
+};
```
GregWeil: I moved this here since it's only/always used with the avatar and does slightly more than convert to rgba (it sets the alpha)

package.json
```diff
@@ -3,7 +3,7 @@
   "version": "0.0.1",
   "description": "Find cool projects and people using Glitch",
   "scripts": {
-    "prestart": "bash sh/watchers.sh &",
+    "prestart": "nodemon --watch server --watch shared --exec 'eslint --config server/.eslintrc.server.js server shared webpack.config.js' &",
```
clottman: What's the benefit of this change? (no preference here, just curious)

src/curated/collections.js
```diff
@@ -0,0 +1,11 @@
+/// Collections shown on the homepage and in the more ideas area
+
+// What collections to show in the expanded area on the homepage
+export const featuredCollections = [
+  { user: 'scientiffic', url: 'microbit' },
```
clottman: Should this be `collection_name` for the second field instead of `url`? The url for a collection would be something like `https://glitch.com/@scientiffic/microbit`.

src/presenters/featured-collections.jsx
```diff
@@ -0,0 +1,100 @@
...
+import {CollectionLink} from './includes/link';
+import {DataLoader} from './includes/loader';
+import {TruncatedMarkdown} from './includes/markdown';
+import ProjectsLoader from './projects-loader';
+import {ProjectsUL} from './projects-list';
+import {TeamTile} from './teams-list';
+import {UserTile} from './users-list';
+
+const CollectionWide = ({collection, api}) => {
+  const dark = getContrastTextColor(collection.coverColor) === 'white' ? 'dark' : '';
```
clottman: The name's a little confusing here - can we call it `contrastTextColor` or something? 
Though I do like the poetic potential of `var dark = white`...

src/presenters/featured-collections.jsx
```diff
@@ -0,0 +1,100 @@
...
+      <article className="projects" style={{backgroundColor: collection.coverColor}}>
+        <header className={`collection ${dark}`}>
+          <CollectionLink className="collection-image-container" collection={collection}>
+            <CollectionAvatar color={collection.coverColor}/>
+          </CollectionLink>
+          <CollectionLink className="collection-name" collection={collection}>
+            <h2>{collection.name}</h2>
+          </CollectionLink>
+          {!!collection.team && <TeamTile team={collection.team}/>}
+          {!!collection.user && <UserTile {...collection.user}/>}
```
clottman: Just a guard against user error here - what if an edit goes wrong and we end up with a `team` and a `user` defined for a single featured collection? Will it fail gracefully?

src/presenters/featured-collections.jsx
```diff
@@ -0,0 +1,100 @@
...
+          <div className="collection-description">
+            <TruncatedMarkdown length={80}>{collection.description}</TruncatedMarkdown>
+          </div>
+        </header>
+        <div className="collection-contents">
+          <ProjectsLoader api={api} projects={collection.projects}>
```
clottman: We pass api down through a lot of levels in a lot of places - should we create an ApiProvider/ApiConsumer and start avoiding passing it down through the component chain when we can?

src/presenters/featured-collections.jsx
```diff
@@ -0,0 +1,100 @@
...
+            {projects => <ProjectsUL projects={projects}/>}
+          </ProjectsLoader>
+          <CollectionLink collection={collection} className="collection-view-all">View all {collection.projectCount} projects →</CollectionLink>
```
clottman: This looks really similar to this line in the ProjectsUL component - 
        `<Link to={collectionUrl} className="collection-view-all">View all {projectCount} projects →</Link>`

Can we share that code somehow? Maybe by passing in a prop that lets you set a class on the View All link so we can control its styling in each place its used?

src/presenters/featured-collections.jsx
```diff
@@ -0,0 +1,100 @@
...
+import {TruncatedMarkdown} from './includes/markdown';
+import ProjectsLoader from './projects-loader';
+import {ProjectsUL} from './projects-list';
+import {TeamTile} from './teams-list';
+import {UserTile} from './users-list';
+
+const CollectionWide = ({collection, api}) => {
+  const dark = getContrastTextColor(collection.coverColor) === 'white' ? 'dark' : '';
+  return (
+    <div className="collection-wide">
```
clottman: Can we refactor the CSS a bit here so we can get rid of this wrapper div?

src/presenters/featured-collections.jsx
```diff
@@ -0,0 +1,100 @@
...
+    name: PropTypes.string.isRequired,
+    url: PropTypes.string.isRequired,
+  }).isRequired,
+  api: PropTypes.any.isRequired,
+};
+
+const loadCollection = async (api, info) => {
+  let collections = [];
+  if (info.team) {
+    const {data: teamId} = await api.get(`teamid/byUrl/${info.team}`);
```
clottman: `teamId` instead of `teamid`, right? (I'm not 100% sure how case works in urls)

src/presenters/featured-collections.jsx
```diff
@@ -0,0 +1,100 @@
...
+    if (teamId !== 'NOT FOUND') {
+      const {data} = await api.get(`collections?teamId=${teamId}`);
+      collections = data;
```
clottman: In this section of code we're declaring `const {data}` at different scopes several times. Would you mind making those variable names a bit more explicit about what data `data` holds, with the added benefit that we won't have to think so much about what scope `data` refers to?

src/presenters/featured-collections.jsx
```diff
@@ -0,0 +1,100 @@
...
+    }
+  } else if (info.user) {
+    const {data: userId} = await api.get(`userid/byLogin/${info.user}`);
```
clottman: `userId` instead of `userid`

src/presenters/featured-collections.jsx
```diff
@@ -0,0 +1,100 @@
...
+    if (userId !== 'NOT FOUND') {
+      const {data} = await api.get(`collections?userId=${userId}`);
+      collections = data;
+    }
+  }
+  const collection = collections.find(c => c.url === info.url);
```
clottman: Can we reuse `loadCollection()` from collection.jsx for this? In our current file structure, I'd move that function to src/utils/models.js or maybe src/utils/collectionHelpers.js (a new file) and reference it from both places.

src/presenters/featured-collections.jsx
```diff
@@ -0,0 +1,100 @@
...
+  if (collection) {
+    const {data} = await api.get(`collections/${collection.id}`);
+    data.projectCount = data.projects.length;
+    data.projects = sampleSize(data.projects, 3).map(p => ({...p, users: p.users||[]}));
```
clottman: How does this handle if a collection has less than 3 projects? (Remembering that even if it has 3 projects when it is featured for the first time, if it's a user-curated collection, they could remove projects at any time and break the page.

src/presenters/more-ideas.jsx
```diff
@@ -0,0 +1,64 @@
...
+import {DataLoader} from './includes/loader';
+
+const MoreIdeasCollectionsDisplay = ({collections}) => (
+  <section className="more-ideas" role="navigation">
+    <h2>More Ideas</h2>
+    <ul>
+      {collections.map(collection => (
+        <li key={collection.id}>
+          <CollectionLink className="more-ideas-box-link" collection={collection}>
+            <div className="more-ideas-box centered" style={{backgroundColor: collection.coverColor}}>
```
clottman: What's the div helping with here? Can we pass the classnames through to collectionAvatar and get rid of the container div?

src/presenters/more-ideas.jsx
```diff
@@ -0,0 +1,64 @@
...
+            }}>{collection.name}</div>
+          </CollectionLink>
+        </li>
+      ))}
+    </ul>
+  </section>
+);
+
+export const MoreIdeasCollections = ({api}) => (
+  <DataLoader get={() => api.get(`teamid/byUrl/${moreIdeasTeam}`)}>
```
clottman: Another good place to reuse loadCollection from collection.jsx, which does all this nesting for us.

src/presenters/more-ideas.jsx
```diff
@@ -0,0 +1,64 @@
...
+    <h2>More Ideas</h2>
+    <ul>
+      {categories.map(category => (
+        <li key={category.id}>
+          <Link className="more-ideas-box-link" to={category.url}>
+            <div className="more-ideas-box centered" style={{backgroundColor: category.color}}>
+              <img src={category.avatarUrl} alt=""/>
+            </div>
+            <div className="more-ideas-box-label centered" style={{backgroundColor: category.color}}>
+              {category.name}
```
clottman: Do we need a wrapper div for this, or the img on the line a couple lines up? We should be able to style the <Link> to have the img to the left of the text without needing either wrapper.

src/curated/collections.js
ttseng: I think it's probably called `url` for legacy reasons (because it was referred to in that way within `categories.js` . I agree that it would make sense to rename it to something like `name` (I think collection is implied since they're featuredCollections)

ttseng: do you think it might make sense to rename these files, to make it clearer what they are?
`src/curated/featured.js` -> `src/curated/feautred-projects.js`
`src/curated/collection.js` -> `src/curated/featured-collections.js`

ttseng: note that renaming `collection.js` to `featured-collections.js` might prevent search result conflicts between `model`/collection.js` and `curated/collection.js`

package.json
GregWeil: I like being able to see all of the sh script together in the package.json. We used to do a lot more in watchers.sh, but most of it got removed as we moved to react and webpack.

src/curated/collections.js
GregWeil: I called it that because it's `collection.url`, but I'm not following that for `user.login` or `team.url` so it's probably better off getting a more specific name

styles/collections.styl
```diff
@@ -250,4 +223,15 @@ collection-avatar-size = 55px
   &:focus
     outline: transparent
     background: #efefef
-    
\ No newline at end of file
+    
+.collection-wide
```
ttseng: it looks like the collection-wide containers are ~40px taller than featured categories on the current homepage: 
image
(modified version, based on suggested CSS below, shown on top; current version shown on bottom)

If you add these styling changes, we can get rid of some of the larger padding:

```
.collection-wide
   header.collection
      width: 25%
      padding: 16px
      margin: 0
      .collection-image-container
           margin: 0 0 15px
     .collection-name
           h2
               margin-bottom: 0px
   .collection-contents
       padding: 16px
```

src/presenters/featured-collections.jsx
GregWeil: I'm going to switch the function to 'isDarkColor' or something more generic like that. Most uses just check if it returned black or white anyways

src/presenters/featured-collections.jsx
GregWeil: It's working right now, so urls probably aren't case sensitive. I'll change it though if it's more canonically right

src/presenters/featured-collections.jsx
GregWeil: It'll just grab every project in the collection, the second featured collection on superb-guide.glitch.me only has two projects

src/curated/collections.js
GregWeil: That makes sense, but could we do that in a smaller pr, this branch already has merges with featured changes from gareth and maurice and renaming the files can make that messy

src/presenters/featured-collections.jsx
GregWeil: The collections page loads a bit differently because it wants the full team data (including projects/members) which takes a lot longer to load than just the id here.

GregWeil: I would like to standardize our api though, mostly so 404/null/NOT FOUND can get handled behind the scenes

clottman: Can we make that function take an option to determine whether to get the full team data? I mostly want to make sure that we're sharing the conditional flow.

garethx:
> @garethx before this can deploy I need a starting set of featured collections
@GregWeil Melissa is working on them and should be ready for Monday. Will keep you updated.

src/presenters/featured-collections.jsx
GregWeil: It'll just show two avatars. I'm not sure if they'd be on separate lines or the same one, but I think it's fine either way for an edge case

src/presenters/featured-collections.jsx
GregWeil: yes

src/presenters/more-ideas.jsx
GregWeil: We can't pass classnames to the collection avatar because it's an inlined svg that gets inserted using a string. That said I refactored it a bit and now it doesn't need any of the divs

src/presenters/featured-collections.jsx
```diff
@@ -0,0 +1,101 @@
...
+        </CollectionLink>
+        {!!collection.team && <TeamTile team={collection.team}/>}
+        {!!collection.user && <UserTile {...collection.user}/>}
+        <div className="collection-description">
+          <TruncatedMarkdown length={80}>{collection.description}</TruncatedMarkdown>
+        </div>
+      </header>
+      <div className="collection-contents">
+        <ProjectsLoader api={api} projects={collection.projects}>
+          {projects => <ProjectsUL projects={projects}/>}
```
clottman: The relevant CSS this is controlling is this: 
```
  .collection-contents
    padding-left: 20px
    .projects-container
      margin: 0
    .collection-view-all
      display: block
      clear: both
``` 

Could we pass an extra classname into ProjectsUL to set the padding and margin, and bring .collection-view-all out from the nesting? As we transition to CSS modules, we won't need the nesting to keep those styles isolated. (And I think we can get rid of the div that starts on line 35)

clottman:
Created https://glitch.manuscript.com/f/cases/3327763/New-API-endpoint-for-get-collection-by-owner-name-collection-name to help deal with some of the load collection complexity.

GregWeil:
Updated with the new /collections/:owner/:name endpoint to speed up and simplify loading.

I also reworked how collection pages load, so they don't get blocked waiting for the team. The add project pop will show only the search box until the team loads.

src/presenters/featured-collections.jsx
```diff
@@ -0,0 +1,93 @@
...
+
+const loadAllCollections = async (api, infos) => {
+  // don't await until every request is sent so they can all run at once
+  const promises = infos.map(info => loadCollection(api, info));
+  return await Promise.all(promises);
+};
+
+export const FeaturedCollections = ({api}) => (
+  <DataLoader get={() => loadAllCollections(api, featuredCollections)}>
+    {collections => collections.filter(c => !!c).map(collection => (
```
clottman: I can't figure out what the filter is doing here: `collections.filter(c => !!c)` - is it guarding against bad data? It seems like loadCollection does that already.

src/presenters/includes/add-collection-project.jsx
```diff
@@ -4,19 +4,44 @@ import PropTypes from 'prop-types';
...
+      this.setState({projects: currentUser.projects});
+    } else if (collection.userId > 0) {
+      const {data: user} = await api.get(`users/${collection.userId}`);
+      this.setState({projects: user.projects});
+    }
+  }
+  componentDidMount() {
+    this.loadProjects();
+  }
+  componentDidUpdate(prevProps) {
```
clottman: Does this ever get triggered? My understanding is that even though we're using multiple AddCollectionProject pops on a page, they'd be in different points in the React component tree, and so a new component will always be created instead of reusing an old one. I could be wrong, though.

src/presenters/includes/add-collection-project.jsx
```diff
@@ -4,19 +4,44 @@ import PropTypes from 'prop-types';
...
+    const collection = this.props.collection;
+    const prevCollection = prevProps.collection;
+    if (collection.teamId !== prevCollection.teamId || collection.userId !== prevCollection.userId) {
+      this.loadProjects();
+    }
+  }
+  render() {
+    return (
+      <PopoverWithButton buttonClass="button add-project opens-pop-over" buttonText="Add Project" passToggleToPop>
```
clottman: As far as I can tell, `.button` isn't needed on `<button>` because the rules are the same. 
```
.button,
button { 
  button styles
...
```

clottman: Does `opens-pop-over` actually do anything? I see it getting used in our JSX, but can't actually find it defined in styles/ anywhere. If not let's ax it and make Allyson's job easier. :)

src/presenters/more-ideas.jsx
```diff
@@ -0,0 +1,57 @@
...
+import categories from '../curated/categories';
+import {moreIdeasTeam} from '../curated/collections';
+import {isDarkColor} from '../models/collection';
+
+import CollectionAvatar from './includes/collection-avatar';
+import Link, {CollectionLink} from './includes/link';
+import {DataLoader} from './includes/loader';
+
+const MoreIdeasCollectionsDisplay = ({collections}) => (
+  <section className="more-ideas" role="navigation">
```
clottman: It's hard to tell for sure without having NVDA or JAWS here, but I think we probably don't want `role='navigation'` on this, at least not right now. We have a lot of other navigation-like things on the page, but I think this might be the only thing using role=navigation or a <nav> tag, so I'm worried that people using the rotor on NVDA/JAWS will see this as one of the only major navigation landmarks and so will get confused. 

My recommendation would be to remote the role here for now, and create a case about checking out the homepage landmarks in JAWS/NVDA to see where more ARIA would be helpful.

clottman:
Oh and a couple about unnecessary class names. :) Easy fixes.

src/presenters/featured-collections.jsx
GregWeil: loadCollections will return [c, null, c] if one of the collections doesn't exist or has an api error. This'll guard against that and just ignore it.

src/presenters/includes/add-collection-project.jsx
GregWeil: Doesn't look like it does anything. Mind if I just change this one spot and open another review with the rest of them? There's about ten uses of it and I'd rather not bring that many extra files into this pr.

clottman: Fine by me.

src/presenters/includes/add-collection-project.jsx
GregWeil: It never should. I cleaned it up a bit to remove the code that will never happen and to not put stuff in state that doesn't have to be there

GregWeil:
Removed extra class names/roles and simplified add collection pop loading

PR #226, opened 2019-01-29
Earthy submarine

ttseng:
Pupdate for featured project embed on profile pages

image


styles/overlays.styl
```diff
@@ -103,17 +103,5 @@ dialog.overlay
   .wistia_embed
     height: 100%
     width: 100%
-
-dialog.overlay-collection
```
whimsicallyson: do we not use this class anymore?

whimsicallyson: we do not. Not sure where we might have in the past, but it doesn't really matter.

PR #228, opened 2019-01-29
Add sitemaps to community site and split build files out from public folder

sheridanvk:
Site available at https://nickel-dill.glitch.me/

Uploaded two sitemap files at Gareth's request (https://nickel-dill.glitch.me/sitemap.xml and https://nickel-dill.glitch.me/marketing-sitemap.xml). We were previously using the public folder purely as a build target, but this PR also modifies webpack to put those files into build/ instead, and the public folder can be used for content like this.

Fixes https://glitch.manuscript.com/f/cases/3327697/Add-sitemap-for-marketing-pages-to-glitch-com


sheridanvk:
> Overall looks good 👍
> We have a bit in server/routes.js that sets the max-age header for .js and .css files. Can you update that to use the express.static maxAge field instead?

done, and tested (604800ms = 7 days):
screenshot 2019-01-30 at 2 41 32 pm

webpack.config.js
```diff
@@ -27,7 +27,7 @@ module.exports = {
   },
   output: {
     filename: '[name].js?[contenthash]',
-    path: PUBLIC,
+    path: BUILD,
```
GregWeil: :+1:

PR #232, opened 2019-01-31
Team feature project

ttseng:
https://team-feature-collection.glitch.me/

(sorry this has nothing to do with collections, but I'm worried that changing the name will mess up syncing with the project, so I'm leaving the project name for now)

Added featured project embed to team pages.

Note - also fixed a bug with filtering of pinned projects on team pages. (You can notice it here: https://glitch.com/@spotify)

Also renamed the "Remove Project" button to "Remove project from team" to clarify that the act is non-destructive.


src/presenters/pages/team.jsx
```diff
@@ -103,6 +105,13 @@ class TeamPage extends React.Component {
   }
 
   render() {
+    
+    const pinnedSet = new Set(this.props.team.teamPins.map(({projectId}) => projectId));
+    const pinnedProjects = this.props.team.projects.filter( ({id}) => pinnedSet.has(id)).filter ( ({id}) => id != this.props.team.featuredProjectId);
```
clottman: This section (109-112) is kind of hard to read, and I think there's probably some tiny performance optimizations we could make. 

What about storing the result of the union between pinnedSet and teamProjects so you don't have to repeat that check?

src/presenters/pages/team.jsx
```diff
@@ -178,10 +187,22 @@ class TeamPage extends React.Component {
           />
         </ErrorBoundary>
         
+        {featuredProject && 
+          <>
```
clottman: What's the fragment helping with here? I think should be able to get rid of it and directly do `{featuredProject && <EntityPageFeaturedProject />}`

src/presenters/team-editor.jsx
```diff
@@ -206,6 +218,9 @@ class TeamEditor extends React.Component {
       updateUserPermissions: (id, accessLevel) => this.updateUserPermissions(id, accessLevel).catch(handleError),
       joinTeamProject: projectId => this.joinTeamProject(projectId).catch(handleError),
       leaveTeamProject: projectId => this.leaveTeamProject(projectId).catch(handleError),
+      addProjectToCollection: (project,collection) => this.addProjectToCollection(project, collection).catch(handleError),
```
clottman: Where do these get used?

src/presenters/pages/team.jsx
ttseng: made some changes to store both pinned projects + the featured project in a set, and then filtering by the set.  do you have any other suggestion to make it more efficient?

src/presenters/pages/team.jsx
ttseng: yup, I removed it

src/presenters/pop-overs/add-team-project-pop.jsx
```diff
@@ -88,10 +89,14 @@ export class AddTeamProjectPop extends React.Component {
     return await this.props.api.post(inviteUserPath);
   }
 
-  onClick(event, project) {
+  onClick(event, project, createNotification) {
     event.preventDefault();
     this.props.togglePopover();
     this.props.addProject(project);
+    
+    console.log('create notification');
```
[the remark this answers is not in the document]
ttseng: yup, removed

src/presenters/pop-overs/add-team-project-pop.jsx
```diff
@@ -88,10 +89,14 @@ export class AddTeamProjectPop extends React.Component {
...
+    
+    createNotification(<p>Added <b><span className="project-name">{project.domain}</span></b></p>, "notifySuccess");
```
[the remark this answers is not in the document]
ttseng: I created a new function in Notification.jsx for this.

src/presenters/pages/team.jsx
ttseng: ok - made better with lodash partition: `const [pinnedProjects, recentProjects] = partition(user.projects, ({id}) => pinnedAndFeaturedSet.has(id));`

src/presenters/pages/team.jsx
```diff
@@ -153,35 +162,45 @@ class TeamPage extends React.Component {
                   inviteEmail={this.props.inviteEmail}
                   inviteUser={this.props.inviteUser}
                   setWhitelistedDomain={this.props.currentUserIsTeamAdmin ? this.props.updateWhitelistedDomain : null}
-                  members={this.props.team.users.map(({id}) => id)}
-                  whitelistedDomain={this.props.team.whitelistedDomain}
+                  members={team.users.map(({id}) => id)}
```
clottman: Just an fyi as I see you using this pattern a lot - you can just call `_.map(team.users, 'id')` if you'd like! No need to change it now though.

src/presenters/team-editor.jsx
clottman: I guess it's probably out of scope for this PR, but the setup here seems really odd. I'll flag it for @sheridanvk & others to look at, though.

webpack.config.js
```diff
@@ -30,7 +30,7 @@ module.exports = {
     path: BUILD,
     publicPath: '/',
   },
-  devtool: mode === 'production' ? 'source-map' : 'eval',
+  devtool: mode === 'production' ? 'source-map' : 'cheap-module-source-map',
```
clottman: Can you confirm with @sheridanvk or @GregWeil that this is an intended change and not a regression caused by git being silly?

src/presenters/pages/team.jsx
ttseng: oh cool, that's definitely more concise. thanks!

webpack.config.js
ttseng: this matches what's currently on production

PR #234, opened 2019-02-04
Quixotic rocket: Hopeful fix for proxied sitemaps

GregWeil:
Over in server/proxy.js I deployed a bit of code that alters any sitemap.xml that gets proxied, rewriting urls to match the request. It works for glitch.me urls, but when you request https://glitch.com/culture/sitemap.xml req.hostname gets proxied as community.glitch.me. This change tells express to respect `X-Forwarded-Host`, which I expect to be set to glitch.com. https://expressjs.com/en/api.html#req.hostname

### DIFF
--- a/server/server.js
+++ b/server/server.js
@@ -28,6 +28,8 @@ const app = express();
 
 app.use(Sentry.Handlers.requestHandler());
 
+app.enable('trust proxy');
+
 // Accept JSON as req.body
 const bodyParser = require("body-parser");
 app.use(bodyParser.urlencoded({ extended: false }));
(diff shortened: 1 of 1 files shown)


GregWeil:
It's what it should be on remixed, I'd expect the community site to keep showing community.glitch.me as the worst case. I can deploy a change to just console.log the header to begin with though.

clottman:
It doesn't sound like it's going to make things worse than current behavior even if it does do community.glitch.me, so it's fine by me if you want to just go ahead and deploy it.

PR #236, opened 2019-02-05
Team feature project

ttseng:
There was a bug with featuredProjects not being properly filtered from recent projects.  
Now, when generating pinnedProjects and recentProjects, the user's projects is completely filtered of the featuredProject.


ttseng:
@clottman I added you as a reviewer because you have the context of the previous PR - let me know if you'd prefer someone else to look at this!

src/presenters/pages/team.jsx
```diff
@@ -108,9 +108,9 @@ class TeamPage extends React.Component {
 
   render() {
     const {team} = this.props;
-    const pinnedAndFeaturedSet = new Set(team.teamPins.map(({projectId}) => projectId)).add(team.featuredProjectId);
-    const [pinnedProjects, recentProjects] = partition(team.projects, ({id}) => pinnedAndFeaturedSet.has(id));
-    const featuredProject = team.projects.find(({id}) => id === team.featuredProjectId);
+    const pinnedSet = new Set(team.teamPins.map(({projectId}) => projectId));
+    const [pinnedProjects, recentProjects] = partition(team.projects.filter(({id}) => id !== team.featuredProjectId), ({id}) => pinnedSet.has(id));
```
clottman: The goal here is that neither side of the partition contains the featured project, is that right? 

If so, maybe add a short comment explaining that so it's a bit easier to grok at a glance.

PR #238, opened 2019-02-05
Prism-flax: Sentry gets more and sometimes less data 

clottman:
- Changed instances of captureMessage to take only one argument (only 1 argument is used by Sentry)
- Changed instance of captureMessage to captureError when possible, to record a stack trace. 
- In certain cases, don't send an error to Sentry if it's an error type that we expect to happen and that we handle gracefully

https://glitch.com/edit/#!/prism-flax
https://prism-flax.glitch.me


src/presenters/pages/join-team.jsx
```diff
@@ -38,7 +38,9 @@ class JoinTeamPageBase extends React.Component {
       // The team is real but the token didn't work
       // Maybe it's been used already or expired?
       console.log('Team invite error', error && error.response && error.response.data);
-      captureMessage('Team invite error', {extra: {error}});
+      if (error && error.status != 401) {
```
GregWeil: I think this needs to be error.response.status

src/presenters/current-user.jsx
```diff
@@ -154,10 +154,10 @@ class CurrentUserManager extends React.Component {
         const newSharedUser = await this.getSharedUser();
         this.props.setSharedUser(newSharedUser);
         console.warn('Fixed shared cachedUser from', sharedUser, 'to', newSharedUser);
-        captureMessage('Invalid cachedUser', {extra: {
+        captureMessage('Invalid cachedUser: ' + JSON.stringify({
           from: sharedUser || null,
           to: newSharedUser || null,
-        }});
+        }));
```
GregWeil: This will send tokens into sentry and bypass the filtering in utils/sentry.js.
Maybe the old/new user should be sent as breadcrumbs, since they support kv data?
https://docs.sentry.io/enriching-error-data/breadcrumbs/?platform=node

clottman:
Also added a filter for errors like these to get ignored: https://sentry.io/fog-creek-software/community/issues/855432405/?query=is%3Aunresolved%20network%20error 
We get A TON and they generally mean that axios couldn't get any kind of connection to the server it was supposed to try reaching. We have other monitoring tools to determine if the API is down, so we can pretty safely assume this is a user with a bad connection, not our API going down. 

Also we get so many that it'd be hard to notice if the API actually was down anyways, as things are now.

PR #243, opened 2019-02-07
Add teams to add-project-to-collection-pop

ttseng:
### Remix
https://add-to-collection-pop.glitch.me

### Description
Add the ability for users with teams to add projects to their team collections.  Adds a new Dropdown element at `/src/presenters/pop-overs/dropdown.jsx`

Test Cases:
1. User without any collections should see an invitation to create their first collection.
first-collection

2. If the user attempts to create a new collection with a name that already exists, then they get an error message that prevents them from clicking the "create" button
collection-exists-error

3. If a user has more than 3 collections, a filter input appears that allows them to search by name.
filtering

4. If a user is part of a team, and some of their team have collections, the results returns all collections the user has write access to, ordered in reverse chronological order
team-and-user

5. When creating a new collection, by default the user creates it for themselves.  They can choose from a dropdown if they want to add the collection to a team instead.  The error messaging handles whether the URL exists already for the particular user, or the selected team.
create-for-team

### Initial Spec
https://www.notion.so/glitch/Updated-Add-to-Collection-Pop-over-ce039990de54441cb5f189a102a9a468


src/presenters/pop-overs/popover-with-button.jsx
```diff
@@ -9,15 +9,20 @@ const PopoverWithButton = props => {
...
+          // what's happening here?
         }
         return (
           <div className={"button-wrap " + props.containerClass}>
             <button
               className={props.buttonClass}
               data-track={props.dataTrack}
               onClick={togglePopover}
+              type="button"      
+              aria-haspopup="listbox"
```
clottman: From the [spec](https://www.w3.org/TR/wai-aria/#aria-haspopup):
> Authors MUST ensure that the role of the element that serves as the container for the popup content is menu, listbox, tree, grid, or dialog, and that the value of aria-haspopup matches the role of the popup container.  

Have you made sure that all of the children of this component are listboxes? If they're not, I think we don't want to set `aria-haspopup` here.

src/presenters/pop-overs/popover-with-button.jsx
```diff
@@ -9,15 +9,20 @@ const PopoverWithButton = props => {
...
+              aria-expanded={visible}
```
clottman: From the [spec](https://www.w3.org/TR/wai-aria/#aria-expanded)
> If the element with the aria-expanded attribute controls the expansion of another grouping container that is not 'owned by' the element, the author SHOULD reference the container by using the aria-controls attribute.

I think we don't want aria-expanded to be set unless we also set aria-controls on the popover content, or it could be confusing.

src/presenters/pop-overs/popover-with-button.jsx
```diff
@@ -9,15 +9,20 @@ const PopoverWithButton = props => {
...
             >
               {props.buttonText}
+              {props.dropdown && <span className="down-arrow icon" aria-label="options"></span>}
```
clottman: Usually when we want to show an icon on the popover-triggering-button, we pass in `buttonText={<>MyLabel <span className='icon'></span></>}` on the PopoverWithButton. I would prefer doing that here, too, so we don't clutter up the component with one-off uses.

clottman: Also, I think if we're going to use aria-label here, we'd want a more descriptive name. I'm not confident how this will display right now - it might be read as "myself options" in the default case?

src/presenters/pop-overs/project-options-pop.jsx
```diff
@@ -128,7 +128,7 @@ const ProjectOptionsContent = ({addToCollectionPopover, ...props}) => {
 // Project Options Pop
 const ProjectOptionsPop = ({...props}) => {
   return(
-    <NestedPopover alternateContent={() => <AddProjectToCollectionPop {...props} api={props.api} togglePopover={props.togglePopover}/>}>
+    <NestedPopover alternateContent={() => <AddProjectToCollectionPop {...props} api={props.api} togglePopover={props.togglePopover}/>} startAlternateVisible={false}>
```
clottman: This is the default, so you don't need to pass it in. 
```
NestedPopover.defaultProps = {
  startAlternateVisible: false,
}; 
```

src/presenters/pop-overs/report-abuse-pop.jsx
```diff
@@ -290,6 +290,7 @@ const ReportAbusePopButton = props => (
   </PopoverWithButton>
 );
 
+
```
clottman: Extra space

styles/pop-overs.styl
```diff
@@ -469,7 +475,7 @@ popover-secondary-background()
   margin-bottom: side-padding
   .users
     margin-bottom: 0
-
+
```
clottman: Extra whitespace.

src/presenters/pop-overs/add-project-to-collection-pop.jsx
```diff
@@ -1,180 +1,172 @@
...
-            <p className="info-description">
-              Organize your favorite projects in one place
-            </p>
-          </section>)
-        ) : <Loader/>}
+        { this.props.collections ?        
+          <>
+            {this.props.collections.length > 3 && (
+              <section className="pop-over-info">
+                <input id="collection-filter"
```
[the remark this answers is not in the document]
ttseng: re: labels for input fields, I agree that the lack of a label is not good from an accessibility standpoint.  I'm going to hold off on making this change here because it affects every other form on the site, and I don't want this form to be inconsistent with similar forms elsewhere.  will create a separate task for evaluating our lack of form labels.

src/presenters/pop-overs/add-project-to-collection-pop.jsx
```diff
@@ -1,180 +1,172 @@
...
   }
   
-  async componentDidMount() {
-    this.loadCollections();
-    try {
-      const collectionPair = await getCollectionPair();
-      this.setState(prev => ({query: prev.query || collectionPair, collectionPair}));
-    } catch (error) {
-      // it's ok rocky. you go when you feel like it
+  componentWillReceiveProps(nextProps){
```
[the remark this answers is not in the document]
ttseng: I removed state.collections, but I'm finding that if I remove the componentWillReceiveProps function altogether, `this.props.collections` and thus `filteredCollections` is null – I guess this.props.collections isn't defined until after `AddProjectToCollectionPop` is initially loaded. Is there some other way around this, beyond `componentWillReceiveProps`?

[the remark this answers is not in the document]
clottman: You can add a default prop for props.collections to make sure it's always an array, even before it gets populated. Below the `propTypes` definition, add this: 

```
AddProjectToCollectionPopContents.defaultProps = {
 collections: [],
};
```

src/presenters/pop-overs/add-project-to-collection-pop.jsx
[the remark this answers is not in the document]
clottman: What about just adding `aria-label='Filter collections'`? Then it won't change the look at all, but people using a screenreader will have some more context.

sheridanvk:
> Let me know if you have any questions about comments I made. Also if you need any help installing the Prettier extension if you haven't used it yet. A lot of my comments are styling things and I didn't necessarily comment on all of them if there were multiple in a file.
> 
> Also, Greg mentioned in slack that we definitely do want public/marketing-sitemap.xml and public/sitemap.xml so you'll want to bring those back.

+1, definitely need those sitemaps back

sheridanvk:
I'll do a full review once the remix is working again (can't get the popover to load right now) and Cassey's comments have been addressed, but wanted to call this out - I have a lot of teams, so it's a lot of calls to the API to find all my collections:
screenshot 2019-02-11 at 2 32 14 pm


Should we ask the infra team whether it makes sense for the first call to /collections?userId=<X> to return all the collections for teams of which I am member in this one call also? They already populate userId = -1 for those collections, so you'd know which ones were team versus user.

src/presenters/includes/collection-result-item.jsx
```diff
@@ -35,25 +35,30 @@ const addProject = (addProjectToCollection, project, collection, collectionPath,
...
+const CollectionResultItem = ({onClick, project, collection, isActive, togglePopover}) => {
   let resultClass = "button-unstyled result result-collection";
   if(isActive) {
     resultClass += " active";
   }
-  const collectionPath = `/@${currentUserLogin}/${collection.url}`;
-
-  return (
+  let collectionPath;
+  if(collection.userId !== -1){
```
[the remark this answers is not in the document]
ttseng: good idea

src/presenters/includes/collection-result-item.jsx
```diff
@@ -35,25 +35,30 @@ const addProject = (addProjectToCollection, project, collection, collectionPath,
...
         <div>
           <button className={resultClass} onClick={() => addProject(onClick, project, collection, collectionPath, createNotification, togglePopover)} data-project-id={project.id}>
             <div className="avatar" id={"avatar-collection-" + collection.id}>
-              <CollectionAvatar backgroundColor={collection.coverColor}/>
+              <CollectionAvatar color={collection.coverColor}/>
             </div>
             <div className="results-info">
               <div className="result-name" title={collection.name}>{collection.name}</div>
               { collection.description.length > 0 && <div className="result-description">{collection.description}</div> }
+              { collection.userId !== -1? <UserAvatar user={collection.owner}/> : <TeamAvatar team={collection.owner}/>}
```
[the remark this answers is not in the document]
ttseng: Switched to 
```
{ collection.team && <TeamAvatar team={collection.team}/> }
{ collection.user && <UserAvatar user={collection.user}/> }
```

src/presenters/pop-overs/add-collection-project-pop.jsx
```diff
@@ -212,7 +212,8 @@ class AddCollectionProjectPop extends React.Component {
   
   render() {
     // load user's recent projects
-    const results = this.state.query ? this.state.maybeResults : this.props.initialProjects;
+    const ownProjects = this.props.collection.team ? this.props.collection.team.projects : this.props.currentUser.projects;
+    const results = this.state.query ? this.state.maybeResults : ownProjects.slice(0,20);
```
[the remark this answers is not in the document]
ttseng: we only give a preview of the user's most recent projects, rather than all of their projects.  this is the same as on team pages as well.

src/presenters/pop-overs/add-project-to-collection-pop.jsx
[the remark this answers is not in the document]
ttseng: 👍

src/presenters/pop-overs/add-project-to-collection-pop.jsx
```diff
@@ -1,180 +1,172 @@
...
-                }
-              </ul>
-            </section>
-          ) : (<section className="pop-over-info">
-            <p className="info-description">
-              Organize your favorite projects in one place
-            </p>
-          </section>)
-        ) : <Loader/>}
+        { this.props.collections ?
```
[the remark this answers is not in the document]
ttseng: good idea - I moved it out

src/presenters/pop-overs/add-project-to-collection-pop.jsx
```diff
@@ -1,180 +1,172 @@
...
+    
+    // next load all of the user's team's collections
+    const userTeams = this.props.currentUser.teams;
+    for(const team of userTeams){
+      const {data} = await this.props.api.get(`collections/?teamId=${team.id}`);
+      const teamCollections = data;
+      if(teamCollections){
+        teamCollections.forEach(teamCollection => {
+          teamCollection.owner = this.props.currentUser.teams.find(userTeam => userTeam.id == team.id);
+          userCollections.data.push(teamCollection);
```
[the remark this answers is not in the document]
ttseng: changed name to `allCollections`

src/presenters/pop-overs/create-collection-pop.jsx
```diff
@@ -0,0 +1,175 @@
...
+import {getLink, defaultAvatar} from '../../models/collection';      
+
+import {NestedPopoverTitle} from './popover-nested.jsx';
+import Dropdown from './dropdown.jsx';
+import {PureEditableField} from '../includes/editable-field.jsx';
+import Loader from '../includes/loader.jsx';
+
+import {kebabCase, orderBy} from 'lodash';
+
+class CreateNewCollectionPop extends React.Component {
```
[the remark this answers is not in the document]
ttseng: changed to `CreateCollectionPop`

src/presenters/pop-overs/create-collection-pop.jsx
```diff
@@ -0,0 +1,175 @@
...
+  setTeamId(buttonContents){
+    const teamId = buttonContents.props.id;
+    this.setState({
+      teamId: teamId
+    });
+  }
+
+  async handleSubmit(event){
+    event.preventDefault();
+    this.setState({working: true});
```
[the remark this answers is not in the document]
ttseng: changed `working` to `loading`.  Used to toggle between the Create Collection button and the Loader.  The redirect is handled separately, after the new collection has been successfully created.

src/presenters/pop-overs/create-collection-pop.jsx
```diff
@@ -0,0 +1,175 @@
...
+      const {data} = await this.props.api.post('collections', {
+        name,
+        url,
+        avatarUrl,
+        coverColor,
+        teamId,
+      });
+
+      // add the selected project to the collection
+      await this.props.api.patch(`collections/${data.id}/add/${this.props.project.id}`);
```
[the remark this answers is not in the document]
ttseng: it should be caught in the catch statement

src/presenters/pop-overs/dropdown.jsx
```diff
@@ -0,0 +1,81 @@
...
+    super(props);
+    this.state = { 
+      selected: 0, 
+      buttonContents: this.props.buttonContents
+    };
+    
+    this.updateSelected = this.updateSelected.bind(this);
+  }
+  
+  componentDidMount(){
```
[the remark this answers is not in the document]
ttseng: oh yeah, wanted to add some default behavior here so that when you access the dropdown from a team page, it defaults to selecting the team, rather than the user.  will add this in.

src/presenters/pop-overs/add-project-to-collection-pop.jsx
```diff
@@ -1,180 +1,172 @@
...
+      maybeCollections: null, // null means still loading
+    };
+  }
+  
+  async loadCollections() {
+    // first, load all of the user's collections
+    const userCollections = await this.props.api.get(`collections/?userId=${this.props.currentUser.id}`);
+    // add current user as owner for collection (for generating user avatar for collection result item)
+    userCollections.data.forEach(userCollection => {
+      userCollection.owner = this.props.currentUser;
```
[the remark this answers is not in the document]
ttseng: the collections endpoint by default does not include the actual owner object (team or user) for the collection - just their associated ID.  However, we need the object in order to display its avatar using our UserAvatar and TeamAvatar components.  I renamed owner to user and team (respectively).

src/presenters/pop-overs/add-project-to-collection-pop.jsx
```diff
@@ -1,180 +1,172 @@
...
+    });
+    
+    // next load all of the user's team's collections
+    const userTeams = this.props.currentUser.teams;
+    for(const team of userTeams){
+      const {data} = await this.props.api.get(`collections/?teamId=${team.id}`);
```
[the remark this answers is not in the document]
ttseng: Simplified with a single request to `collections/?userId=${this.props.currentUser.id}&includeTeams=true` (Isaac updated the endpoint)

src/presenters/pop-overs/create-collection-pop.jsx
```diff
@@ -0,0 +1,175 @@
...
+    const teams = this.props.currentUser.teams;
+    const currentUserMenuItem = <span>myself <UserAvatar user={this.props.currentUser} isStatic={true}/></span>;
+    
+    function getTeamMenuContents(){
+      const orderedTeams = orderBy(teams, team => team.name.toLowerCase());   
+      const menuContents = [];
+      menuContents.push(currentUserMenuItem); // add user as first option
+      
+      orderedTeams.map(team => {
+        let content = <span id={team.id}>{team.name} {<TeamAvatar team={team} className="user"/>}</span>;
```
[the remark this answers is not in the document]
ttseng: yeah, agreed - I removed it

src/presenters/pop-overs/create-collection-pop.jsx
```diff
@@ -0,0 +1,175 @@
...
+    return (
+      <dialog className="pop-over create-collection-pop wide-pop">
+        <NestedPopoverTitle>
+          Add {this.props.project.domain} to a new collection
+        </NestedPopoverTitle>
+        
+        <section className="pop-over-actions">
+          <form onSubmit={this.handleSubmit}> 
+            <PureEditableField
+              id="collection-name"
```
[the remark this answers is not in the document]
ttseng: nope - removed the ID and added an aria-label

src/presenters/pop-overs/add-project-to-collection-pop.jsx
```diff
@@ -1,180 +1,172 @@
...
+      if(teamCollections){
+        teamCollections.forEach(teamCollection => {
+          teamCollection.owner = this.props.currentUser.teams.find(userTeam => userTeam.id == team.id);
+          userCollections.data.push(teamCollection);
+        });
+      }
+    }
+    
+    // order reverse chronological
+    let orderedCollections = orderBy(userCollections.data, collection => collection.updatedAt).reverse();
```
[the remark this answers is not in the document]
ttseng: hm - for some reason this is working when I test it in the console (loading lodash https://medium.com/@matt.leo/how-to-import-lodash-to-chrome-console-3e5e30b4933e), but it's not working when I add it to `add-project-to-collection-pop` (the app seems to hang)

src/presenters/pop-overs/create-collection-pop.jsx
```diff
@@ -0,0 +1,175 @@
...
+    let queryError = error;
+    let submitEnabled = this.state.query.length > 0;
+    let nameTakenError = "You already have a collection with this url";
+
+    let placeholder = "New Collection Name";
+    
+    const teams = this.props.currentUser.teams;
+    const currentUserMenuItem = <span>myself <UserAvatar user={this.props.currentUser} isStatic={true}/></span>;
+    
+    function getTeamMenuContents(){
```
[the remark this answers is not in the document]
ttseng: ok - I moved the function out of render

src/presenters/pop-overs/popover-with-button.jsx
ttseng: got rid of the dropdown prop from popover-with-button

src/presenters/pop-overs/add-project-to-collection-pop.jsx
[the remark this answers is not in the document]
clottman: Weird - are you getting any errors?

ttseng:
I tried out adding (React-Select)[https://github.com/JedWatson/react-select], which is functionally working at https://blushing-hexagon.glitch.me

However, in order to override its default styling, you have to use their auto-generated class names, which use a different convention than our classes, e.g.

image

I'm not sure if this is a problem but feel free to test it out at https://blushing-hexagon.glitch.me - if it looks fine, then I'll add this to `add-project-to-collection`

ttseng:
added dropdown using `react-select` and custom styling to override defaults from `react-select` 
dropdown-styling

this adds keyboard support to the dropdown component.

src/presenters/pop-overs/add-project-to-collection-pop.jsx
```diff
@@ -1,180 +1,241 @@
...
-import _ from 'lodash';
+import { NestedPopoverTitle } from "./popover-nested.jsx";
 
-class AddProjectToCollectionPop extends React.Component {
+import { orderBy } from "lodash";
+
+const AddProjectPopoverTitle = ({ project }) => {
+  return (
+    <NestedPopoverTitle>
+      <img
```
clottman: Can we use the <Avatar> component for this, so we're standardized on alt text? I believe Tara just made a change so the alt text on a project avatar is ''.

src/presenters/pop-overs/add-project-to-collection-pop.jsx
```diff
@@ -1,180 +1,241 @@
...
-      const coverColor = randomColor({luminosity: 'light'});
-      
-      const {data} = await this.props.api.post('collections', {
-        name,
-        description,
-        url,
-        avatarUrl,
-        coverColor,
+  componentWillReceiveProps(nextProps) {
+    if (nextProps.collections !== this.props.collections) {
```
clottman: When would the collections provided to the component change?

src/presenters/pop-overs/add-project-to-collection-pop.jsx
```diff
@@ -1,180 +1,241 @@
...
-    const placeholder = 'New Collection Name';
-    const {error, maybeCollections, query} = this.state;
-    let queryError = this.state.error;
-    if (!!maybeCollections && !!query && maybeCollections.some(c => c.url === _.kebabCase(query))) {
-      queryError = 'You already have a collection with this url';
-    }
-    if(this.state.newCollectionUrl){
-      return <Redirect to={this.state.newCollectionUrl}/>;
+    const { filteredCollections, query } = this.state;
+    const NoSearchResultsPlaceholder = (
```
clottman: We should declare this outside the render() method so it doesn't get re-instantiated on every re-render.

src/presenters/pop-overs/add-project-to-collection-pop.jsx
```diff
@@ -1,180 +1,241 @@
...
+      <p className="info-description">
+        No matching collections found – add to a new one?
+      </p>
+    );
+    const NoCollectionPlaceholder = (
```
clottman: Same with this - declare it outside `render`

src/presenters/pop-overs/add-project-to-collection-pop.jsx
```diff
@@ -1,180 +1,241 @@
...
+      allCollections.forEach(collection => {
+        if(collection.teamId == -1){
+          collection.user = this.props.currentUser;
+        }else{
+          collection.team = this.props.currentUser.teams.find( userTeam => userTeam.id == collection.teamId);
+        }
+      });
+      // const orderedCollections = orderBy(allCollections, ['updatedAt'], ['desc']);
+
+      const orderedCollections = orderBy(
```
clottman: does `orderBy(allCollections, collection => collection.updatedAt, ['desc']` work?

src/presenters/pop-overs/create-collection-pop.jsx
```diff
@@ -0,0 +1,210 @@
...
+import { kebabCase, orderBy } from "lodash";
+
+class CreateCollectionPop extends React.Component {
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      loading: false,
+      query: "", //The entered collection name
+      teamId: undefined // by default, create a collection for a user, but if team is selected from dropdown, set to teamID,
```
clottman: The comma at the end make me think this comment might be missing some words?

src/presenters/pop-overs/create-collection-pop.jsx
```diff
@@ -0,0 +1,175 @@
...
+              className="pop-over-input create-input"
+              value={query} 
+              update={this.handleChange}
+              placeholder={placeholder}
+              error={error || queryError}
+            />
+            
+            { this.props.currentUser.teams.length > 0 &&
+              <div>
+                for <Dropdown buttonContents={currentUserMenuItem} menuContents={getTeamMenuContents()} onUpdate={this.setTeamId}/>
```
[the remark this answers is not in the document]
ttseng: clarified variable naming to pass both user and team menu items

src/presenters/pop-overs/dropdown.jsx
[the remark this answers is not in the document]
ttseng: we don't store in context whether the user is currently on a user or team page - to avoid passing this down unnecessarily, I'm going to skip the auto-selection of user or team based on page to a separate PR

src/presenters/pop-overs/create-collection-pop.jsx
```diff
@@ -0,0 +1,175 @@
...
+  async handleSubmit(event){
+    event.preventDefault();
+    this.setState({working: true});
+    // get text from input field
+    const newCollectionName = this.state.query;
+    
+    // create a new collection
+    try{
+      const name = newCollectionName;
+      const url = kebabCase(newCollectionName);
```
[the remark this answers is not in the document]
ttseng: I created a function in `collection.js` called `createCollection` that does name & description generation + the API request to create the new collection and handles error notification.

`createCollection` is called from `collections-list.jsx` and `create-collection-pop`

src/presenters/pop-overs/create-collection-pop.jsx
```diff
@@ -0,0 +1,175 @@
...
+        teamId,
+      });
+
+      // add the selected project to the collection
+      await this.props.api.patch(`collections/${data.id}/add/${this.props.project.id}`);         
+      
+      // redirect to that collection
+      if(data && data.url){
+        if(this.state.teamId){
+          const {data: team} = await this.props.api.get(`/teams/${this.state.teamId}`);
```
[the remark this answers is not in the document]
ttseng: I removed the API request by filtering the `currentUsers`'s teams, which are already loaded from `currentUserConsumer`

src/presenters/pop-overs/create-collection-pop.jsx
```diff
@@ -0,0 +1,175 @@
...
+        const newCollectionUrl = getLink(data);
+        this.setState({newCollectionUrl});
+      }
+      
+
+    }catch(error){
+      if (error && error.response && error.response.data && error.response.data.message) {
+        this.setState({error: error.response.data.message});
+      } else {
+        captureException(error);
```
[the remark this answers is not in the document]
ttseng: Move error messaging to `collection.js`

ttseng:
After speaking with @jennschiffer, I've left the dropdown as keyboard accessible using `react-select` with the plan that we can add voiceover support in a later PR – testing with popular screenreaders like JAWs, which I guess are often used over the built-in Voiceover tool.

The main changes I made are:
1. I created a `createCollection` method in `collection.js`
2. I depend on good error messaging from the API to pass the default messages using `handleCustomError` in `error-handlers.jsx` . And some changes to improve error handling more generally for collection creation / adding projects to collections.

At this point, I didn't move a separate wrapper into dropdown.jsx to handle option creation because I wasn't sure of the most generic way to do so without passing in info specific to `create-collection-pop`

src/presenters/collection-editor.jsx
```diff
@@ -25,40 +27,53 @@ class CollectionEditor extends React.Component {
...
   async addProjectToCollection(project, collection) {
-    if(collection.id == this.state.id){
-      // add project to collection page
-      this.setState(({projects}) => ({
-        projects: [...projects, project],
-      }));
-    }
-    await this.props.api.patch(`collections/${collection.id}/add/${project.id}`);
+    await this.props.api.patch(`collections/${collection.id}/add/${project.id}`)
+      .then(() => {
```
clottman: Usually we use either await, or then(). I'm not sure how to read what happens when we use both. Maybe switch it over to just await?

clottman: Also need to catch errors.

src/presenters/includes/collection-result-item.jsx
```diff
@@ -1,49 +1,33 @@
...
-    const content = <AddProjectMessage projectName={project.domain} collectionName={collection.name} url={collectionPath}/>;
+    const content = <AddProjectToCollectionMsg projectDomain={project.domain} collectionName={collection.name} url={collectionPath}/>;
     notification(content, "notifySuccess");
-  }catch(error){
-    const content = <p>Something went wrong. Try refreshing and adding the project again.</p>;
-    notification(content, "notifyError");
-  }
+  });
+
+  // toggle popover
```
clottman: This comment probably isn't needed.

src/presenters/notifications.jsx
```diff
@@ -80,4 +85,31 @@ export class Notifications extends React.Component {
...
+export const AddProjectToCollectionMsg = ({
+  projectDomain,
+  collectionName,
+  url
+}) => (
+  <>
+    <p>
+      Added {projectDomain} { collectionName ? (`to collection ${collectionName}`) : null }
+    </p>
+    { url ?
```
clottman: Can simplify this to just `{url && <a />}` vs `{url ? <a/> : null}`

src/presenters/pop-overs/dropdown.jsx
```diff
@@ -0,0 +1,49 @@
...
+  constructor(props) {
+    super(props);
+    this.state = {
+      selectedOption: this.props.options[0] // set to the first option by default
+    };
+    this.updateSelected = this.updateSelected.bind(this);
+  }
+
+  shouldComponentUpdate(nextProps, nextState) {
+    return nextState.selectedOption.value !== this.state.selectedOption.value; // render is continually called without this check...
```
clottman: I think this is probably because you're storing state in two places - the Dropdown component as well as it's parent (Dropdown passes the selected option up via props.onUpdate, so the parent will update its own state and set a new value in Dropdown's props.selectedOption, which triggers the onChange, which starts the process over. 

Pick one component to be the 'owner' of the state to break the loop.

src/presenters/pop-overs/dropdown.jsx
```diff
@@ -0,0 +1,49 @@
...
+    return (
+      <Select
+        autoWidth={true}
+        value={this.state.selectedOption}
+        options={this.props.options}
+        className={"dropdown " + this.props.containerClass}
+        classNamePrefix="dropdown"
+        onChange={this.updateSelected}
+        isSearchable={false}
+        aria-label="test"
```
clottman: Remove or replace with a real value :)

src/presenters/collection-editor.jsx
ttseng: I think the error is caught already by 
```addProjectToCollection: (project, collection) =>
        this.addProjectToCollection(project, collection).catch(
          handleCustomError
        ),
```
I tested this by adding a project to a collection that has since been added by another user (e.g., adding a project that already exists in a collection) and was able to see the error message "Project already belongs to the collection"
feb-19-2019 16-05-35

I guess it might be better to reload the collection projects just to be safe (to show that the project is already part of the collection), but I think that's a separate issue.

src/presenters/includes/collection-result-item.jsx
ttseng: ah right - it was just a note for myself (:

src/presenters/notifications.jsx
ttseng: oh ok, I was basing it off of the conversation in #community about whether or not to use && instead of : null, but sure I can switch it

ttseng: will do the same for collectionName too

src/presenters/pop-overs/dropdown.jsx
ttseng: good catch!

src/presenters/pop-overs/dropdown.jsx
ttseng: got it - I move state out of dropdown complete and set it in `create-collection-pop`

ttseng:
Changes to `create-collection-pop` to move state out of `dropdown-jsx`

Still noticing the warning that cassey pointed out regarding the new-project-pop - not sure why, as I think it has to do with `project-result-item`, which hasn't been modified here

```
index.js:3725 Warning: validateDOMNesting(...): <a> cannot appear as a descendant of <a>.
    in a (created by Link)
    in Link (created by ForwardRef(Link))
    in ForwardRef(Link) (created by ProjectLink)
    in ProjectLink (created by ProjectResultItem)
    in div (created by ProjectResultItem)
    in ProjectResultItem (created by NewProjectPop)
    in a (created by TrackedExternalLinkWithoutContext)
    in TrackedExternalLinkWithoutContext (created by Context.Consumer)
    in TrackedExternalLink (created by NewProjectPop)
    in div (created by NewProjectPop)
    in section (created by NewProjectPop)
    in div (created by NewProjectPop)
    in NewProjectPop (created by NewProjectPopButton)
    in div (created by PopoverContainer)
    in Wrapper (created by OnClickOutside(Wrapper))
    in OnClickOutside(Wrapper) (created by PopoverContainer)
    in PopoverContainer (created by PopoverWithButton)
    in PopoverWithButton (created by NewProjectPopButton)
    in NewProjectPopButton (created by Header)
    in nav (created by Header)
    in header (created by Header)
    in Header (created by PopoverContainer)
    in PopoverContainer (created by NewStuff)
    in NewStuff (created by Context.Consumer)
    in UserPref (created by Context.Consumer)
    in UserPref (created by NewStuffContainer)
    in NewStuffContainer (created by Context.Consumer)
    in CurrentUserConsumer (created by HeaderContainer)
    in HeaderContainer (created by Layout)
    in div (created by Layout)
    in Layout (created by CollectionPage)
    in CollectionPage (created by Route)
    in Route (created by Router)
    in Switch (created by Router)
    in Router (created by CurrentUserManager)
    in CurrentUserManager (created by LocalStorage)
    in LocalStorage (created by LocalStorage)
    in LocalStorage (created by CurrentUserProvider)
    in CurrentUserProvider (created by App)
    in AnalyticsContext (created by App)
    in UserPref (created by DevTogglesProvider)
    in DevTogglesProvider (created by App)
    in LocalStorage (created by UserPrefsProvider)
    in UserPrefsProvider (created by App)
    in Notifications (created by App)
    in Router (created by BrowserRouter)
    in BrowserRouter (created by App)
    in ErrorBoundary (created by App)
    in App
(anonymous) @ index.js:3725
warningWithoutStack @ react-dom.development.js:506
validateDOMNesting @ react-dom.development.js:8597
createInstance @ react-dom.development.js:8704
completeWork @ react-dom.development.js:16450
completeUnitOfWork @ react-dom.development.js:18619
performUnitOfWork @ react-dom.development.js:18817
workLoop @ react-dom.development.js:18829
renderRoot @ react-dom.development.js:18912
performWorkOnRoot @ react-dom.development.js:19809
performWork @ react-dom.development.js:19721
performSyncWork @ react-dom.development.js:19695
interactiveUpdates$1 @ react-dom.development.js:19964
interactiveUpdates @ react-dom.development.js:2169
dispatchInteractiveEvent @ react-dom.development.js:4878
sentryWrapped @ index.js:3223
```

Going to look more into this tomorrow.

clottman:
The error about an <a> nested in <a> in NewProjectPop happens in a fresh remix of community also, so I think it's just not an error that shows in production builds. I [created a case](https://glitch.manuscript.com/f/cases/3327887/New-Project-Pop-contains-an-a-nested-in-a-but-shouldn-t) for it; it's not a blocker here since it's not related to your PR.

src/presenters/pop-overs/add-collection-project-pop.jsx
```diff
@@ -111,137 +131,163 @@ class AddCollectionProjectPop extends React.Component {
...
-    const results = this.state.query ? this.state.maybeResults : this.props.initialProjects;
-    
+    const ownProjects = this.props.collection.team
+      ? this.props.collection.team.projects
+      : this.props.currentUser.projects;
+    const results = this.state.query
+      ? this.state.maybeResults
+      : ownProjects
+        ? ownProjects.slice(0, 20)
+        : null; // TO DO - comment out
```
clottman: Still need this?

src/presenters/pop-overs/add-project-to-collection-pop.jsx
clottman: Maybe just remove alt text content on this and we can call it good? (`alt=''`)

src/presenters/pop-overs/add-project-to-collection-pop.jsx
```diff
@@ -1,180 +1,237 @@
...
+          <AddProjectPopoverTitle project={this.props.project} />
         )}
-        
-        {maybeCollections ? (
-          maybeCollections.length ? (
-            <section className="pop-over-actions results-list">
-              <ul className="results">
-                {maybeCollections.map(collection =>   
+
+        {filteredCollections.length > 3 && (
```
clottman: You have a null check on line 86 - do you need one here too, or is the one on line 86 unnecessary?

src/presenters/pop-overs/add-project-to-collection-pop.jsx
```diff
@@ -1,180 +1,237 @@
...
+          <section className="pop-over-info">
+            <input
+              className="pop-over-input search-input pop-over-search"
+              onChange={evt => {
+                this.updateFilter(evt.target.value);
+              }}
+              placeholder="Filter collections"
+              arial-label="Filter collections"
```
clottman: aria-label, not arial :)

src/presenters/pop-overs/create-collection-pop.jsx
```diff
@@ -0,0 +1,210 @@
...
+      this.props.api,
+      this.state.query,
+      this.state.selection.value,
+      createNotification
+    );
+    // add the project to the collection
+    if (collectionResponse && collectionResponse.id) {
+      const collection = collectionResponse;
+      // add the selected project to the collection
+      await this.props.api
```
clottman: Another case of mixing `await` and `then()`.

src/presenters/pop-overs/create-collection-pop.jsx
```diff
@@ -0,0 +1,210 @@
...
+      <span>
+        myself <UserAvatar user={this.props.currentUser} />
+      </span>
+    );
+    const currentUserOption = { value: -1, label: currentUserOptionLabel };
+
+    this.state = {
+      loading: false,
+      query: "", //The entered collection name
+      options: [currentUserOption].concat(
```
clottman: I don't see `setState` ever getting called to change `options` - does it need to be in state?

src/presenters/pop-overs/dropdown.jsx
```diff
@@ -0,0 +1,33 @@
+import React from "react";
+import PropTypes from "prop-types";
+import Select from "react-select"; // https://react-select.com/
+
+// Options passed to Dropdown are expected to be formatted like options = [ {value: optionValue, label: optionLabel}, ... ]
+class Dropdown extends React.Component {
+  constructor(props) {
```
clottman: You can get rid of this constructor since it's not doing anything. (`super(props)` is handled automatically if there's no constructor)

src/presenters/pop-overs/add-collection-project-pop.jsx
ttseng: nope - removed

src/presenters/pop-overs/add-project-to-collection-pop.jsx
ttseng: you're right - don't need the check on 86

ttseng:
In this PR, I made the change that adding a project to a new collection shows a notification with a link to the new collection, rather than automatically redirecting.  This enables us to show error notifications if necessary, and also lets users keep working on whatever they were working on.

I updated the PR so that if you're creating a new collections from create-collection-pop on a user profile, then it refreshes the collections on the page:

new-collection

I wanted to do the same for team collections, but it seems like team collections were never instrumented for updating collections dynamically as projects are added to them.  (for example, if you add a project to a team collection from a team page, the collection card does not update to show a new project count & the newly added project - however, doing the same on your own profile page does update the collections).

As this is an existing bug, and the team page is structured differently than user pages, I think that needs to be a separate PR.  In the meantime, I designed it so that if you create a new collection using create-collection-pop from a team page, then it redirects you automatically, which is aligned with the current behavior.

team-collection

(tagging @GregWeil on this just so he's aware)

shrinkwrap.yaml
```diff
@@ -1,13 +1,13 @@
 dependencies:
-  '@babel/core': 7.3.3
+  '@babel/core': 7.2.2
```
clottman: Any idea why we're backtracking on package versions? Maybe you missed a merge in package.json?

ttseng: that seems to be the version on community https://glitch.com/edit/#!/community?path=package.json:23:29

ttseng: checked out shrinkwrap.yaml from master

clottman: Weird... must have backtracked in master for some reason. 🤷‍♀️

src/presenters/collections-list.jsx
```diff
@@ -1,13 +1,12 @@
...
-import randomColor from 'randomcolor';
-import { kebabCase, orderBy } from 'lodash';
+import { orderBy } from 'lodash';
 import { TrackClick } from './analytics';
 import CollectionItem from './collection-item';
-import { defaultAvatar, getLink } from '../models/collection';
-import { getCollections, getPredicate } from '../models/words';
+import { getLink, createCollection } from '../models/collection';
 import { Loader } from './includes/loader';
+import { NotificationConsumer } from './notifications';
```
clottman: Can we switch this to `useNotification`? (What Greg added yesterday?)

clottman: If we don't that's ok. :) You've put in a lot of work on this.

ttseng: ok - will do in a separate PR.  on quick testing, looks like I gotta create a separate functional component to use the hook (outside of the createcollectionbutton component)

PR #244, opened 2019-02-08
Adding CSS Modules, Storybook and our first Atomic Design Component (Button)

sheridanvk:
Demo app at https://splendid-carol.glitch.me/

This is the first of the PRs for our move towards Atomic Design. It contains the following:

1. Addition of CSS Modules for the src/components directory, alongside the existing global CSS. 
- This ensures that we can incrementally move the site towards the new approach, while still keeping the old approach running.
2. Addition of Storybook, available at https://splendid-carol.glitch.me/storybook. 
- Currently building this is not hooked into any of the watcher processes, because the app was frequently running out of memory when it was doing both. 
- Hopefully down the line this will be possible, but for now you have to manually run "npm run storybook" from the console to get the latest components to show up in the storybook (after adding all the states into the /stories folder).
3. Creation of the Button component. 
- Note that this has a hover property on it that will be passed in from higher level components (like Project Items), but which is unused for now. 
- The component only covers in-page actions for now - it will be extended later to allow for off-page links too.
4. Implementation of the Button component for the Upload Avatar/Cover Image functions on User and Team profile, as a first example of how to use it.

Soliciting feedback from the whole team!


src/components/buttons/button.jsx
```diff
@@ -0,0 +1,44 @@
...
+
+Button.propTypes = {
+  /** callback when button clicked */
+  onClick: PropTypes.func,
+  /** button disabled */
+  disabled: PropTypes.bool,
+  /** type of button */
+  type: PropTypes.oneOf(TYPES),
+  /** size of button */
+  size: PropTypes.string,
```
clottman: Small thing - can this be `PropTypes.oneOf(SIZES)` ?

src/components/buttons/button.styl
```diff
@@ -0,0 +1,55 @@
...
+    text-decoration: none
+  &:active,
+  &.active
+    background-color: secondary-background-active
+    color: primary
+
+  &[disabled]
+    opacity: .5
+    pointer-events: none
+    cursor: pointer
```
clottman: Why do we need to set `cursor` here if `pointer-events: none` is set?

src/components/buttons/button.jsx
```diff
@@ -0,0 +1,44 @@
...
+  });
+  
+  return (
+    <button onClick={onClick} className={className} disabled={disabled}>
+      {children}
+    </button>
+  );
+};
+
+Button.propTypes = {
```
clottman: Should we also add `children: PropTypes.node.isRequired` so we don't accidentally end up with an empty button?

stories/index.js
```diff
@@ -0,0 +1,23 @@
...
+  .add('small', () => (
+    <Button size="small">Small Button</Button>
+  ))
+.add('tertiary', () => (
+    <Button type="tertiary">Tertiary Button</Button>
+  ))
+.add('danger zone', () => (
+    <Button type="dangerZone">Destructive Action</Button>
+  ))
+.add('data tracking', () => (
```
clottman: Does this one work currently? I don't see dataTrack in the proptypes, or that it's getting passed through to <button> anywhere.

src/components/buttons/button.jsx
sheridanvk: yes good call!

src/components/buttons/button.jsx
sheridanvk: 👍

src/components/buttons/button.styl
sheridanvk: no need, good catch

stories/index.js
sheridanvk: right, it doesn't work - I actually had removed this between the initial PR submission and your comment, but hadn't removed it from the story yet 😄

package.json
```diff
@@ -4,7 +4,8 @@
   "description": "Find cool projects and people using Glitch",
   "scripts": {
     "prestart": "nodemon --watch server --watch shared --exec 'eslint --config server/.eslintrc.server.js server shared webpack.config.js' &",
-    "start": "nodemon --watch server --watch shared --watch views ./server/server.js"
+    "start": "nodemon --watch server --watch shared --watch views ./server/server.js",
+    "storybook": "build-storybook -c .storybook -o build/storybook"
```
jennschiffer: make sure this npm script is documented in CONTRIBUTING.md

package.json
```diff
@@ -38,6 +40,7 @@
     "express-http-proxy": "^1.5.1",
     "group-by-time": "^1.0.0",
     "html-truncate": "^1.2.2",
+    "http-proxy": "^1.17.0",
```
jennschiffer: (remove bc unused)

GregWeil:
The storybook doesn't have the right font set for the buttons, because it's missing the webtype reference and falling back on sans-serif. `<link rel="stylesheet" href="https://cloud.webtype.com/css/3a8e55c6-b1f3-4659-99eb-125ae72bd084.css">`

sheridanvk:
> The storybook doesn't have the right font set for the buttons, because it's missing the webtype reference and falling back on sans-serif. `<link rel="stylesheet" href="https://cloud.webtype.com/css/3a8e55c6-b1f3-4659-99eb-125ae72bd084.css">`

Damn, good catch :D added this into the head file of the storybook preview iframe using the method described [here](https://storybook.js.org/configurations/add-custom-head-tags/): 
screenshot 2019-02-11 at 11 44 38 am
.

ttseng:
Just a few quick questions:
* I noticed that in other storyboard demos, some default inline styling seems to be added out-of-the-box: (e.g. https://lonelyplanet.github.io/backpack-ui/?selectedKind=Styles&selectedStory=Design%20tokens&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs)
Is there something about how we configured Storyboard that would remove these default styles for storybox/index.html?

* By default, the storyboard config files aren't visible in the Glitch editor.  Is it possible to move them into a director that makes them viewable within Glitch?

sheridanvk:
> Just a few quick questions:
> 
> * I noticed that in other storyboard demos, some default inline styling seems to be added out-of-the-box: (e.g. https://lonelyplanet.github.io/backpack-ui/?selectedKind=Styles&selectedStory=Design%20tokens&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs)
>   Is there something about how we configured Storyboard that would remove these default styles for storybox/index.html?

I didn't remove anything, but it seems like something might be happening with the CSS Modules integration that is stripping out styles (storybook has them all inline). I spent some time on it but couldn't work out why it was happening - I will check with Greg and Allyson to see if there's anything we can do. If it's ok with you, I won't consider this blocking, but I agree it's something we should try to resolve.

> * By default, the storyboard config files aren't visible in the Glitch editor.  Is it possible to move them into a director that makes them viewable within Glitch?

Right, seems like we don't show any folders that begin with "." in the editor :/ looks like I can change the name of the directory with a config setting though, that's a great idea - I'll change to storybook-config (don't want to confuse new hires with the fact that the URL for the guide itself is /storybook)

ttseng:
cool, that sounds good to me!

pketh:
I it looks like none of the buttons support emoji, is that intentional?

sheridanvk:
> it looks like none of the buttons support emoji (ie as a param), is that intentional?

yes, I should have clarified - ButtonWithEmoji is coming next, which will make it clear that emoji should always be on the right hand side now. Also related to that change is this question: https://glitch.manuscript.com/f/cases/3327633/make-Emoji-component-so-we-are-consistent-in-how-we-display-cross-site-w-correct-role-alt, which I just saw you replied to but I think we should catch up on quickly in person. 

I can send ButtonWithEmoji in as the next PR.

pketh:
^ ok left a reply in the fogbugz case linked above

pketh:
there are some pixel level differences between the storybook buttons and the editor ones
screen shot 2019-02-12 at 1 37 43 pm

(the heights of buttons with emojis are the same as without - heights are hardset)

tertiary button styling also has inconsistencies
screen shot 2019-02-12 at 1 39 22 pm
screen shot 2019-02-12 at 1 40 09 pm

sheridanvk:
> there are some pixel level differences between the storybook buttons and the editor ones
> screen shot 2019-02-12 at 1 37 43 pm
> 
> (the heights of buttons with emojis are the same as without - heights are hardset)
> 
I think you're comparing this between the editor and the community site, which are already not consistent apparently. Regular buttons in the editor are 32px in height; regular buttons on the community site are 29px.

> tertiary button styling also has inconsistencies
> screen shot 2019-02-12 at 1 39 22 pm
> screen shot 2019-02-12 at 1 40 09 pm

again this seems to be due to inconsistencies between editor and community. Editor tertiary buttons are 25px high; community are 24px.

Ideally I could just focus on replicating the community site as-is for now, and down the line we can start fixing consistency between the two apps - what do you think? Otherwise we'll get into inconsistencies if the new button components have different heights to the old community buttons, as we'll be in a state of transition as we componentise everything.

pketh:
ya that's fine, I don't want to over-expand the scope of this pr

PR #245, opened 2019-02-11
Flip the logic on shouldsenderror checks, ignore server errors too

GregWeil:
Bleh, sorry I missed this in review.

We should try to share this config between the client and server, but they use separate npm packages and env variables. What do you think of just a function taking `Sentry` and `PROJECT_DOMAIN`?


GregWeil:
That shouldn't be possible, the inline script to define PROJECT_DOMAIN is before any of the scripts get referenced.

clottman:
Okay, I thought I had seen some before but I'm not finding any now. Never mind then! Lgtm!

PR #246, opened 2019-02-11
Collection cards

taravancil:
Fixes [case 3327632](https://glitch.manuscript.com/f/cases/3327632/make-collections-display-on-profile-page-screenreader-keyboard-navigable)

## Changes
1. Breaks down collection cards into multiple links instead of one big link:
    - the header links to the collection
    - each project in the projects list is a link to the project
    - "View {n} projects" links to the collection
2. Removes alt captions for avatars that are purely decorative

## Testing

See [collection-cards.glitch.me/@taravancil](https://collection-cards.glitch.me/@taravancil) to test out collections with various numbers of projects.

**Note**: Tabbing through the page works fine in Chrome, but in Firefox it jumps from the user settings menu back up to the URL bar? Anyone know what that's about?


src/presenters/collection-item.jsx
```diff
@@ -1,90 +1,138 @@
...
+              />
+            </ProjectLink>
           </li>
-        )) }
+        ))}
       </ul>
-      <div className="collection-link">
-        View <Pluralize count={projects.length} singular="project"/> →
-      </div>
+      <CollectionLink collection={collection} className="collection-link link">
```
clottman: What's the `link` class contributing? In Chrome it looks like the only 2 properties it provides are being overwritten.

styles/collections.styl
```diff
@@ -167,6 +160,11 @@ collection-avatar-size = 55px
           background-color: private
           .private-project-badge
             display: inline-block
+        &:hover
```
clottman: Can we add &:focus to this to make the focus style match the hover style? Same for the one on the collection title button (hover makes the button get a thicker bottom border; focus should, too.)

src/presenters/collection-item.jsx
taravancil: Looks like nothing, good catch!

styles/collections.styl
taravancil: Yep, done! The `:hover` styles for the collection title come from `.button-area` in `buttons.styl`. As far as I can tell adding `:focus` state to that class shouldn't cause problems for other components that use `.button-area`

PR #247, opened 2019-02-11
Change getDominantColor to match the api

GregWeil:
getDominantColor returns null if it finds any transparency in the image. That upsets the api which wants empty colors to be specified as an empty string.
I also did some very minor cleanup to make it more obvious what's happening

### DIFF
--- a/src/utils/assets.js
+++ b/src/utils/assets.js
@@ -116,7 +116,7 @@ export function getDominantColor(image) {
     colors = outlyingColors;
   }
   if (transparentPixels) {
-    return null;
+    return '';
   } 
   const colorMap = quantize(colors, 5);
   const [r, g, b] = Array.from(colorMap.palette()[0]);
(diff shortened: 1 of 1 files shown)


ttseng:
@GregWeil From looking at how the user and team color / coverColor properties are used, it seems like they're used as CSS background-color attributes in the case that the uploaded avatar/cover images don't not load properly.  Is that right? 

I think it would be worth including a comment block before the function explaining what the function is used for, like we do for `resizeImage` and `drawCanvasThumbnail` in `assets.js`

GregWeil:
Updated with a comment blob!

ttseng:
functionally, `getDominantColor` is looking good to me.  However, when I actually go about testing it, it looks like the user's color attribute is not set as expected, but only for user avatars (cover images seem fine).

I tried uploading the following image as my avatar:

test

The correct dominant color is identified, but the user's color attribute is not updated.

GregWeil:
It looks like the user's color field isn't writable, any changes just get ignored.

ttseng:
ok - it looks to me like your changes can be merged, and we can let infrastructure know to make that field rewritable independently.

PR #248, opened 2019-02-11
Update to the latest react, add hooks linting

GregWeil:
I refactored one component to use hooks, not the greatest example, but it let me make sure they worked and test out linting.
Also added bonus network error logging on the server.


package.json
```diff
@@ -33,6 +33,7 @@
     "eslint-loader": "^2.1.1",
     "eslint-plugin-jsx-a11y": "^6.1.2",
     "eslint-plugin-react": "^7.12.2",
+    "eslint-plugin-react-hooks": "^1.0.0",
```
clottman: I think we also need to turn on the hooks rules eslint rule in our eslintrc. 
```
{
  "plugins": [
    // ...
    "react-hooks"
  ],
  "rules": {
    // ...
    "react-hooks/rules-of-hooks": "error"
  }
}
```

It doesn't seem included in eslint-plugin-react.

server/api.js
```diff
@@ -23,6 +24,8 @@ async function getFromCacheOrApi(id, cache, api) {
     const value = await promise;
     return value;
   } catch (error) {
+    console.error(error);
+    captureException(error);
```
clottman: I haven't been updating the sentry config on the server side to filter things out - I'm guessing we'll get double the flood of 401s and timeouts if we turn this on now. :( Maybe we should wait to add this until we also are ready to spend time cleaning up our `beforeSend` here?

src/presenters/featured.jsx
```diff
@@ -10,41 +10,34 @@ import Link from './includes/link.jsx';
...
-class ZineItems extends React.Component {
-  constructor(props) {
-    super(props);
-    this.state = {
-      posts: window.ZINE_POSTS.slice(0, 4), 
-      masks: sampleSize([1, 2, 3, 4, 5], 4),
-    };
+const ZineItems = () => {
+  const [posts] = React.useState(window.ZINE_POSTS.slice(0, 4));
+  const [masks] = React.useState(sampleSize([1, 2, 3, 4, 5], 4));
```
clottman: Since neither of these are using the `setState` or the hooks equivalents, do they actually need to be stored in state? I'm not sure what we're gaining by having state for these at all.

package.json
GregWeil: https://github.com/FogCreek/Glitch-Community/pull/248/files#diff-6afabf7a87ff122d39ba12ec08fbc041R40

It's included here

src/presenters/featured.jsx
GregWeil: masks is random, so it needs to be in state so it stays the same across renders. posts is a global constant and will never change, but I like using state for it since it's pulling from an external source.

package.json
clottman: Sorry I missed that!

server/api.js
GregWeil: Ok, I'll remove this for now

PR #249, opened 2019-02-11
Edit contributing.md to add references to scripts & pushing to glitch

clottman:
Include how to use sh/ scripts in the normal deployment flow, and add info about updating Glitch from your local machine.


GregWeil:
`git pull origin my-remix` will only work in a remix if their branch is on our github repo. They'll have to create a new remote for their fork in github.

clottman:
Oh good point - so if they're on a fork, it would be `git pull myfork my-remix`? After adding a remote for their fork?

PR #253, opened 2019-02-13
Webpack changes to reduce out of memory issues in dev

sheridanvk:
https://lumpy-organ.glitch.me (currently building in dev mode)

Three changes:
1. Changed to using the oneOf syntax so that not all file paths are evaluated against all the rules in the webpack file (excludes eslint, which does need to be run in addition to the matching loader)
2. Removed node_module transpilation in dev.
3. Updated the webpack version.

Confirmed that this still results in the site rendering in Chromium 41 on production settings. Dev builds now use far less memory and don't result in OOM issues any more.


webpack.config.js
```diff
@@ -76,52 +78,57 @@ module.exports = {
...
-              sourceMap: mode !== 'production', // no css source maps in production
-            },
+            test: /\.styl/,
+            include: CSS_MODULES,
+            use: [
+              MiniCssExtractPlugin.loader,
+              {
+                loader: 'css-loader?modules',
+                options: {
+                  sourceMap: true,
```
clottman: Can we make this ` mode !== 'production'`? I must have missed this when making that change to the styles/ directory.

package.json
```diff
@@ -67,7 +67,7 @@
     "stylus-loader": "^3.0.2",
     "terser-webpack-plugin": "^1.2.1",
     "url-join": "^4.0.0",
-    "webpack": "^4.28.4",
+    "webpack": "^4.29.2",
```
clottman: Why not 4.29.3 while we're updating anyways?

sheridanvk: huh I upgraded to the latest that the Glitch UI said was available. I wonder why it didn't give me .3 - do you know if we delay on that for some reason (as it's only 6 days old)?

Updating now though 👍

webpack.config.js
sheridanvk: done; it wasn't there yet when you made changes (it came in yesterday as part of the CSS modules change)

PR #254, opened 2019-02-13
Cuddly tarp: persistently see avatars of those invited to a team

whimsicallyson:
https://glitch.manuscript.com/f/cases/3326592/teams-persistently-see-avatars-of-people-that-have-been-invited-to-a-team

https://cuddly-tarp.glitch.me/@fantastic-band
screen shot 2019-02-13 at 2 31 38 pm

Tara and I worked on this, and neither of us were sure if the solution we came up with is "good React". Interested to hear what reviewers have to say!


src/presenters/includes/team-users.jsx
```diff
@@ -101,11 +101,17 @@ export class AddTeamUser extends React.Component {
...
       invitee: '',
-      alreadyInvited: [],
+      alreadyInvited: this.props.invitedMembers,
     };
     this.removeNotifyInvited = this.removeNotifyInvited.bind(this);
   }
   
+  componentDidUpdate(prevProps) {
+    if (this.props.invitedMembers !== prevProps.invitedMembers) {
+      this.setState(() => ({alreadyInvited: this.props.invitedMembers}));
```
clottman: If you're only using the state to keep track of what the prop is and not setting it somewhere else, you can just use `{this.props.invitedMembers}` in `render` and remove it from state/remove this method. Each time the props change, render is called again anyways so the value will always be current.

src/presenters/pages/team.jsx
```diff
@@ -70,9 +70,15 @@ class TeamPage extends React.Component {
     super(props);
     this.state = {};
     this.teamAdmins = this.teamAdmins.bind(this);
+    this.getInvitees = this.getInvitees.bind(this);
     this.addProjectToCollection = this.addProjectToCollection.bind(this);
   }
   
+  async componentDidMount() {
+    this.invitees = await this.getInvitees();
+    this.setState(this.state);
```
clottman: What's this line doing?

src/presenters/pages/team.jsx
```diff
@@ -105,6 +111,16 @@ class TeamPage extends React.Component {
     }
     return false;
   }
+  
+  async getInvitees() {
+    const ids = this.props.team.tokens.map(({userId}) => userId);
+    const invitees = [];
+    for (const id of ids) {
+      const user = await this.props.api.get(`users/${id}`);
```
clottman: We need some error handling here - either via a try/catch if you want to stick with the `await` syntax, or a `.catch()` block if you want to switch to using `.then()`.

clottman: Also, awaiting API calls while in a loop means the calls will run synchronously instead of all being fired at once. We probably want to use `Promise.all()` instead.

src/presenters/includes/team-users.jsx
whimsicallyson: if another user gets invited (see line 120), invitedMembers is updated to add the new user. I'm pretty sure this means we need to keep it in state?

src/presenters/pages/team.jsx
whimsicallyson: this is one of those "not sure if its good React" bits. Originally I only called `getInvitees()` when I was sending it as a prop in `AddTeamUser` (line 183), but because it was a promise (I think) it didn't update in the class once it had completed. Basically this seemed to be the way to get the data into the spot we needed it.

clottman: I think you want invitees to be in state, and pass in `this.state.invitees` to your AddTeamUser component.

src/presenters/includes/team-users.jsx
GregWeil: You can store only the additions in state and render with `[...this.props.invited, ...this.state.newlyInvited]`

src/presenters/pages/team.jsx
whimsicallyson: getting `Promise.all()` to work was more convoluted than expected (mapping a second time didn't occur to me, go Tara!) but this should be properly being async and filtering out nulls!

GregWeil: Can you add handling for if the request gets a 404 too

whimsicallyson: updated, I think that will work.

src/presenters/pages/team.jsx
```diff
@@ -105,6 +113,21 @@ class TeamPage extends React.Component {
...
+      const data = await Promise.all(this.props.team.tokens.map(({userId}) => (
+        this.props.api.get(`users/${userId}`)
+      )));
+      const invitees = data.map(user => user.data).filter(user => !!user);
+      return invitees;
+    } catch (error) {
+      if (error && error.response && error.response.status === 404) {
+        return null;
+      }
+      throw error;
```
clottman: What happens if you `throw` here? Are there any errors we need to handle? Do we care if this fails?

whimsicallyson: I'm not entirely sure how to test it (don't know how to force a 404) but its the same pattern used with 404 errors in other parts of the site.

clottman: I have a slight preference for calling `captureException(error);` (from `import { captureException } from "../../utils/sentry";` instead of throwing, but I think since Sentry captures errors that bubble up anyways it might not matter much. 

You could test by replacing your api call with another promise that you define somewhere else that waits a second and then resolves a failure, just to test that it doesn't break the UI in some way.

whimsicallyson: Added `captureException` to be safe, and tested with a fake call to make sure it didn't break things horribly--looked fine to me!

PR #255, opened 2019-02-13
Superb guide: Upgrade dependencies, hopeful sentry releases

GregWeil:
https://docs.sentry.io/workflow/releases/?_ga=2.200134830.1878052308.1549902989-1700619958.1539894700&platform=javascript
I'm not doing the full releases setup because it involves extra cli process around deploys, and being able to blame specific commits isn't helpful when they're all checkpoints.
What I ended up with is storing the server start time as the release. I was hoping to get the webpack build time so it would update with live changes, but webpack doesn't include the build time in stats.json unless you're using `webpack --json`


styles/includes/team-analytics.styl
```diff
@@ -112,8 +112,8 @@ analytics-remixes = #ff7f0f
   .controls,
   .project-details
     button
-    margin-left: 0
-    margin-right: 5px
+      margin-left: 0
+      margin-right: 5px
```
GregWeil: Bonus bug fix found by upgraded css-loader. This was building as `.project-details { buttonmargin-left:0; ... }`

PR #256, opened 2019-02-14
Team admin bugfixes

taravancil:
**Case** [3327689](https://glitch.manuscript.com/f/cases/3327689/Trying-to-remove-myself-as-the-only-admin-from-a-team-needs-a-clearer-error-message)
**Remix** https://team-admin-bugfixes.glitch.me

This fixes when you're the only admin on a team and try to remove yourself or demote your admin status. Instead of updating the error message when an admin tries to do something they shouldn't, I decided to hide those controls entirely so they can't do forbidden things!

Most of these changes are Prettier changes, the meaningful changes are on the following lines:
- [`userIsTheOnlyAdmin`](https://github.com/FogCreek/Glitch-Community/compare/team-admin-bugfixes?expand=1#diff-5efa2e8a59389e5b21ef386d07506282R32)
- [`38`-`41`: `canChangeUserAdminStatus`](https://github.com/FogCreek/Glitch-Community/compare/team-admin-bugfixes?expand=1#diff-b9d4bc0e6c46070efd4b9cb6e5638827R38)
- [`canRemoveUser`](https://github.com/FogCreek/Glitch-Community/compare/team-admin-bugfixes?expand=1#diff-b9d4bc0e6c46070efd4b9cb6e5638827R93)
- [remove user rendering](https://github.com/FogCreek/Glitch-Community/compare/team-admin-bugfixes?expand=1#diff-b9d4bc0e6c46070efd4b9cb6e5638827R140)

## Testing

1. Be on a team where you're the only admin
2. Click your avi
3. You should not see either the admin or remove user controls

Then
1. Be on a team where you and others are admins
2. Click your avi
3. You should see both admin and remove user controls

### DIFF
--- a/src/presenters/includes/team-users.jsx
+++ b/src/presenters/includes/team-users.jsx
@@ -1,29 +1,37 @@
-import React from 'react';
-import PropTypes from 'prop-types';
+import React from "react";
+import PropTypes from "prop-types";
 
-import {getDisplayName} from '../../models/user';
-import {TrackClick} from '../analytics';
-import {WhitelistedDomainIcon} from './team-elements.jsx';
-import {UserAvatar} from '../includes/avatar.jsx';
-import AddTeamUserPop from '../pop-overs/add-team-user-pop.jsx';
-import PopoverWithButton from '../pop-overs/popover-with-button.jsx';
-import PopoverContainer from '../pop-overs/popover-container.jsx';
-import TeamUserInfoPop from '../pop-overs/team-user-info-pop.jsx';
-import UsersList from '../users-list.jsx';
+import { getDisplayName } from "../../models/user";
+import { TrackClick } from "../analytics";
+import { WhitelistedDomainIcon } from "./team-elements.jsx";
+import { UserAvatar } from "../includes/avatar.jsx";
+import AddTeamUserPop from "../pop-overs/add-team-user-pop.jsx";
+import PopoverWithButton from "../pop-overs/popover-with-button.jsx";
+import PopoverContainer from "../pop-overs/popover-container.jsx";
+import TeamUserInfoPop from "../pop-overs/team-user-info-pop.jsx";
+import UsersList from "../users-list.jsx";
 
 // Team Users list (in profile container)
 
-export const TeamUsers = (props) => (
+export const TeamUsers = props => (
   <ul className="users">
     {props.users.map(user => (
       <li key={user.id}>
-        <PopoverWithButton 
-          buttonClass="user button-unstyled" 
-          buttonText={<UserAvatar user={user} suffix={adminStatusDisplay(props.adminIds, user)}/>}
-          passToggleToPop 
+        <PopoverWithButton
+          buttonClass="user button-unstyled"
+          buttonText={
+            <UserAvatar
+              user={user}
+              suffix={adminStatusDisplay(props.adminIds, user)}
+            />
+          }
+          passToggleToPop
         >
           <TeamUserInfoPop
             userIsTeamAdmin={props.adminIds.includes(user.id)}
+            userIsTheOnlyAdmin={
+              props.adminIds.length === 1 && props.adminIds.includes(user.id)
+            }
             userIsTheOnlyMember={props.users.length === 1}
             user={user}
             {...props}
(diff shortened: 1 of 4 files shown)


src/presenters/pop-overs/team-user-info-pop.jsx
```diff
@@ -138,12 +169,12 @@ TeamUserInfoAndRemovePop.propTypes = {
...
 };
 
 TeamUserInfoAndRemovePop.defaultProps = {
-  currentUserIsOnTeam: false,
+  currentUserIsOnTeam: false
 };
 
-export default TeamUserInfoAndRemovePop;
\ No newline at end of file
+export default TeamUserInfoAndRemovePop;
```
sheridanvk: This was super confusing to me, because I realised the name of the component we're exporting is not the same as the file name or what we call it when we import it elsewhere. Should we modify the name of the default export from TeamUserInfoAndRemovePop to TeamUserInfoPop?

src/presenters/includes/team-users.jsx
```diff
@@ -1,29 +1,37 @@
...
+              user={user}
+              suffix={adminStatusDisplay(props.adminIds, user)}
+            />
+          }
+          passToggleToPop
         >
           <TeamUserInfoPop
             userIsTeamAdmin={props.adminIds.includes(user.id)}
+            userIsTheOnlyAdmin={
+              props.adminIds.length === 1 && props.adminIds.includes(user.id)
```
clottman: We're having to check `props.adminIds.includes(user.id)` twice now (line 31 & 33) - can we cache that result in a variable so we only have to calculate it once?

src/presenters/pop-overs/team-user-info-pop.jsx
clottman: Maybe in a separate PR, but I'd definitely support this. I hate not being able to `grep` for what I'm importing.

src/presenters/pop-overs/team-user-info-pop.jsx
```diff
@@ -1,124 +1,155 @@
...
-        </TrackClick>
-      ) : (
-        <TrackClick name="Make an Admin clicked">
-          <button className="button-small button-tertiary has-emoji" onClick={() => updateUserPermissions(user.id, ADMIN_ACCESS_LEVEL)}>
-            Make an Admin <span className="emoji fast-up" />
-          </button>
-        </TrackClick>
-      )}
-    </section>
+    canChangeUserAdminStatus && (
```
clottman: Just to make the nesting easier to follow, can we check canChangeUserAdminStatus and return null right away if false, instead of using the && shorthand?

src/presenters/pop-overs/team-user-info-pop.jsx
```diff
@@ -1,124 +1,155 @@
...
-
 // Team User Info 😍
 
-const TeamUserInfo = ({currentUser, showRemove, ...props}) => {
-  const userAvatarStyle = {backgroundColor: props.user.color};
-  const canRemoveUser = props.currentUserIsTeamAdmin || (currentUser && currentUser.id === props.user.id);
+const TeamUserInfo = ({ currentUser, showRemove, ...props }) => {
+  const userAvatarStyle = { backgroundColor: props.user.color };
+  const canRemoveUser =
+    !(props.userIsTheOnlyMember || props.userIsTheOnlyAdmin) &&
```
clottman: This one is a little hard to follow too - can we split up some of the chunks into named variables so it's easier to grok what we're checking for?

src/presenters/includes/team-users.jsx
taravancil: Good call, done!

src/presenters/pop-overs/team-user-info-pop.jsx
taravancil: Sure thing

src/presenters/pop-overs/team-user-info-pop.jsx
taravancil: Done, let me k now if how I broke it down is still difficult to grok

taravancil:
> seems so weird to me that we're passing things like "is the user the only member of the team" in as a prop right now, instead of passing in context to the popover and determining this within the popup itself

Totally agree! It feels like a thorough review of how we're passing props around is in order - I bet moving over to Hooks will force us to rethink a lot of that

I fixed the things you both suggested, @sheridanvk and @clottman lmk if you think it's good to go

PR #258, opened 2019-02-14
Desert pufferfish - lint all the things

clottman:
https://desert-pufferfish.glitch.me <-- will need lots of testing to make sure this didn't break anything.

See some rules that are being used now you don't like? Speak up, a lot of this is just one set of opinions and another opinion will work too, as long as we standardize. 

Types of feedback I'm looking for: 
 - If you disagree with a rule and would prefer changing it
- If a transform I made to comply with a rule changes behavior in an unexpected way


src/app.jsx
```diff
@@ -1,24 +1,24 @@
...
-import {CurrentUserProvider} from './presenters/current-user.jsx';
-import {UserPrefsProvider} from './presenters/includes/user-prefs.jsx';
-import {DevTogglesProvider} from './presenters/includes/dev-toggles.jsx';
-import {Notifications} from './presenters/notifications.jsx';
+import ErrorBoundary from "./presenters/includes/error-boundary";
+import { AnalyticsContext } from "./presenters/analytics";
+import { CurrentUserProvider } from "./presenters/current-user";
+import { UserPrefsProvider } from "./presenters/includes/user-prefs";
+import { DevTogglesProvider } from "./presenters/includes/dev-toggles";
+import { Notifications } from './presenters/notifications'; // eslint-disable-line
```
GregWeil: Why is eslint disabled here?

clottman: Good question, I'll un-disable it.

PR #261, opened 2019-02-18
Quixotic rocket: Fix the private frontend page

GregWeil:
https://quixotic-rocket.glitch.me/~delicate-technician/404

Projects with a private glitch.me can still be public on the community site, which confused the check and caused a redirect loop. This checks if you can get a token to see the project.

### DIFF
--- a/src/presenters/current-user.jsx
+++ b/src/presenters/current-user.jsx
@@ -7,7 +7,7 @@ import axios from 'axios';
 import {configureScope, captureException, captureMessage} from '../utils/sentry';
 import LocalStorage from './includes/local-storage.jsx';
 
-const {Provider, Consumer} = React.createContext();
+const Context = React.createContext();
 
 // Default values for all of the user fields we need you to have
 // We always generate a 'real' anon user, but use this until we do
@@ -217,9 +217,9 @@ export const CurrentUserProvider = ({children}) => (
         {(sharedUser, setSharedUser, loadedSharedUser) => (loadedSharedUser && loadedCachedUser &&
           <CurrentUserManager sharedUser={sharedUser} setSharedUser={setSharedUser} cachedUser={cachedUser} setCachedUser={setCachedUser}>
             {({api, ...props}) => (
-              <Provider value={props}>
+              <Context.Provider value={props}>
                 {children(api)}
-              </Provider>
+              </Context.Provider>
             )}
           </CurrentUserManager>
         )}
@@ -232,13 +232,16 @@ CurrentUserProvider.propTypes = {
 };
 
 export const CurrentUserConsumer = (props) => (
-  <Consumer>
+  <Context.Consumer>
     {({currentUser, fetched, ...funcs}) => props.children(currentUser, fetched, funcs, props)}
-  </Consumer>
+  </Context.Consumer>
 );
-
 CurrentUserConsumer.propTypes = {
   children: PropTypes.func.isRequired,
+};
+
+export const useCurrentUser = () => {
+  return React.useContext(Context);
 };
 
 export function normalizeUser(user, currentUser) {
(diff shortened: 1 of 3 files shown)


src/presenters/pages/error.jsx
```diff
@@ -78,27 +79,25 @@ OauthErrorPage.propTypes = {
...
-    }
-  }
-  render() {
-    return null;
-  }
-}
+  };
+  React.useEffect(() => {
+    check(); // effects can't be async
+  }, [name, currentUser && currentUser.token]);
```
clottman: Note for anyone else who had to look up this syntax: the array second parameter means "only run the effect function if the values in this array have changed". 

Since our second element in the array is `currentUser && currentUser.token`, the `check` function won't get re-run if there's a different user but that user also has a token. Is that the intended behavior?

src/presenters/pages/error.jsx
```diff
@@ -78,27 +79,25 @@ OauthErrorPage.propTypes = {
...
+  return null;
```
clottman: I'm only seeing that `useEffect` takes two parameters - what's the `return null` here doing?

If you meant it to be 'no clean-up callback necessary', I think 
 1) that would go on the line after `check()`, inside the first arg's function brackets, and 
 2) that's the default, so we don't need to explicitly return null.
 
https://reactjs.org/docs/hooks-reference.html#useeffect

src/presenters/pages/error.jsx
```diff
@@ -78,27 +79,25 @@ OauthErrorPage.propTypes = {
...
-      this.check();
-    }
-  }
-  render() {
-    return null;
-  }
-}
+  };
+  React.useEffect(() => {
+    check(); // effects can't be async
```
clottman: I think the comment here is a little misleading - you can use an async function, but it's kind of confusing/cluttered syntax so the pattern you're using here is recommended. https://github.com/facebook/react/issues/14326 

Without the comment calling a function instead of using an inline anonymous one wouldn't really have stood out to me, so I think it's probably fine to remove the comment.

src/presenters/pages/error.jsx
GregWeil: The return null is after useEffect, it's the actual component not rendering anything. I suppose there's no reason for this to exist as a separate component anymore.

src/presenters/pages/error.jsx
GregWeil: Yeah, I guess what I meant was 'you can't pass an async function to useeffect' but I'll just remove the comment

src/presenters/pages/error.jsx
clottman: Oh, I see now. My bad for missing that. 👍

src/presenters/pages/error.jsx
GregWeil: Whoops, this should just be grabbing persistent token. current user always exists so there's no need for that check

src/presenters/layout.jsx
```diff
@@ -11,9 +11,7 @@ import Konami from './includes/konami.jsx';
 
 const Layout = ({children, api, searchQuery}) => (
   <div className="content">
-    <Helmet>
-      <title>Glitch</title>
-    </Helmet>
+    <Helmet title="Glitch" />
```
GregWeil: https://github.com/nfl/react-helmet/issues/437

React helmet breaks if you use the nested element syntax in a component using effect hooks. Hopefully they get the fix out soon, but this works just fine too.

src/presenters/current-user.jsx
```diff
@@ -232,15 +232,18 @@ CurrentUserProvider.propTypes = {
...
     {({currentUser, fetched, ...funcs}) => props.children(currentUser, fetched, funcs, props)}
-  </Consumer>
+  </Context.Consumer>
 );
-
 CurrentUserConsumer.propTypes = {
   children: PropTypes.func.isRequired,
 };
 
+export const useCurrentUser = () => {
```
clottman: Game-changer! 💯

sheridanvk: ooh this is awesome! will you be telling the rest of the team about this change, in case they don't look at this PR (maybe you already discussed in the meeting yesterday?)

PR #262, opened 2019-02-19
More Sentry filtering

clottman:
https://glitch.manuscript.com/f/cases/3327831/Add-a-global-filter-on-sending-401s-and-timeouts-to-Sentry

If the error was just a timeout or a 401, don't send it to Sentry as it was probably fine or the user's problem. 

If the project is configured to use the staging API or doesn't have 'production' in node_env (on the server side), also don't send errors to Sentry.


server/server.js
```diff
@@ -3,16 +3,25 @@ const compression = require("compression");
 
 // https://docs.sentry.io/error-reporting/quickstart/?platform=node
 const Sentry = require("@sentry/node");
-const shouldSendError = process.env.PROJECT_DOMAIN === "community" || process.env.PROJECT_DOMAIN === "community-staging";
+const onProductionSite = (process.env.PROJECT_DOMAIN === "community" || process.env.PROJECT_DOMAIN === "community-staging") && process.env.NODE_ENV === "production";
```
GregWeil: We have a check in `server/constants` for if we're running on staging or production. It's not exported but it would be a good test here

src/utils/sentry.js
```diff
@@ -11,7 +11,7 @@
 import * as Sentry from "@sentry/browser";
 export * from "@sentry/browser";
 
-const shouldSendError = PROJECT_DOMAIN === "community" || PROJECT_DOMAIN === "community-staging";
+const onProductionSite = (PROJECT_DOMAIN === "community" || PROJECT_DOMAIN === "community-staging") && API_URL === "https://api.glitch.com/";
```
GregWeil: `index.ejs` picks out if we're on staging or production to set `API_URL`, this should just test against that.

src/utils/sentry.js
```diff
@@ -28,18 +28,19 @@ try {
...
       if (
         hint.originalException &&
-        hint.originalException.message === "Network Error"
-      ) {
-        // axios couldn't find the server it was supposed to make a request to - probably user network error
+        (hint.originalException.message === "Network Error" ||
+         hint.originalException.message.indexOf("timeout") !== -1 ||
+         hint.orginalException.message.indexOf("status code 401" !== -1)
+        )
+      ) {
```
GregWeil: These checks are getting a bit complicated to be duplicated. Can they get moved into shared/ as some helper functions?

src/utils/sentry.js
```diff
@@ -1,4 +1,4 @@
-/* globals BUILD_TIMESTAMP, ENVIRONMENT, PROJECT_DOMAIN */
+/* globals BUILD_TIMESTAMP, ENVIRONMENT, PROJECT_DOMAIN, API_URL, _env */
```
GregWeil: There's a webpack build error because API_URL isn't used

clottman:
The runtime errors about module.exports being read only was solved by adding a line to .babelrc. 
https://github.com/webpack/webpack/issues/4039#issuecomment-419284940

.babelrc
```diff
@@ -14,4 +14,5 @@
     "@babel/syntax-dynamic-import",
     "@babel/plugin-proposal-class-properties"
   ],
+  "sourceType": "unambiguous"
```
clottman: This fixes the issue I was having with webpack not liking a file that uses module.exports and also exports something, too.

src/presenters/analytics.jsx
```diff
@@ -52,7 +52,13 @@ export const AnalyticsTracker = ({ children }) => (
           resolveProperties(context, inherited.context),
         );
       } catch (error) {
-        captureException(error);
+        /* From Segment: "We currently return a 200 response for all API requests so debugging should be done in the Segment Debugger.
```
clottman: We currently get a ton of errors that might be from Segment - I'm hoping this will clear things up! 🤞

PR #265, opened 2019-02-20
Superb guide: use hooks to make editors way more readable

GregWeil:
`useCurrentUser()` `useNotifications()` `useErrorHandlers()` and `useUploader()` are all equivalent to the their `<BlahConsumer>` versions but don't require tons of nesting to use (you can't use hooks in class components though, so there is still one layer of wrapping to gather them up)

I also cleaned up some props that are marked as optional but aren't.

### DIFF
--- a/src/presenters/error-handlers.jsx
+++ b/src/presenters/error-handlers.jsx
@@ -1,5 +1,3 @@
-import PropTypes from 'prop-types';
-
 import { useNotifications } from './notifications';
 
 function handleError(notify, error) {
@@ -17,7 +15,7 @@ function handleErrorForInput(notify, error) {
   return Promise.reject();
 }
 
-export const useErrorHandlers = () => {
+const useErrorHandlers = () => {
   const { createErrorNotification } = useNotifications();
   return {
     handleError: error => handleError(createErrorNotification, error),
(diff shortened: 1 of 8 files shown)


src/presenters/error-handlers.jsx
```diff
@@ -1,7 +1,4 @@
-import React from 'react';
-import PropTypes from 'prop-types';
-
-import { NotificationConsumer } from './notifications';
+import { useNotifications } from './notifications';
```
clottman: @ttseng if this PR gets in before yours, you'll probably have merge conflicts again in this file. Let me know if you want help sorting those out.

clottman:
@GregWeil I think you're a few commits behind on Github - I see this line in current-user.jsx on Glitch but not here: 
```
export const useCurrentUser = () => React.useContext(Context);
```

src/presenters/project-editor.jsx
```diff
@@ -55,31 +55,24 @@ ProjectEditor.defaultProps = {
...
-            {children}
-          </ProjectEditor>
-        )}
-      </CurrentUserConsumer>
-    )}
-  </ErrorHandlers>
-);
+const ProjectEditorContainer = ({ api, children, initialProject }) => {
+  const { currentUser } = useCurrentUser();
+  const wrapErrors = useErrorHandlers();
```
clottman: While we're here should we rename this variable to be `errorFuncs` or something more descriptive & consistent with the other places we use `useErrorHandlers`?

clottman:
Approved once this PR gets up to date with what's on Glitch. :D This is beautiful.

GregWeil:
> 
> 
> @GregWeil I think you're a few commits behind on Github - I see this line in current-user.jsx on Glitch but not here:
> 
> ```
> export const useCurrentUser = () => React.useContext(Context);
> ```

I actually snuck this in a pr a few days ago, but only used it in the one place
https://github.com/FogCreek/Glitch-Community/pull/261

PR #266, opened 2019-02-20
Name should not be null in an editable field, even if it is

clottman:
glitch.com/~lavish-team. Fixes issue where a user is on their user page and their name is null in the db.


clottman:
Editable field already has  `value: PropTypes.string.isRequired`, so if it happened during development we'd see the error that the prop it got was invalid.

PR #267, opened 2019-02-21
Shadow edger: don't console.log big objects, and other small fixes

clottman:
- breadcrumbs instead of console logging
- CollectionWide.props.avatarUrl should not be required 
- TrimmedValue makes extra sure that the value it's initialized with is a string (previous fix didn't solve the name-is-null problem :() https://sentry.io/organizations/fog-creek-software/issues/899779354/?project=1246508

glitch.com/~shadow-edger


src/presenters/current-user.jsx
```diff
@@ -35,10 +36,15 @@ const defaultUser = {
 function identifyUser(user) {
   const analytics = { window };
   if (user) {
-    console.log('👀 current user is', user);
-    console.log('🌈 login', user.login, user.id);
+    addBreadcrumb({
+      level: 'info',
+      message: `Current user is ${JSON.stringify(user)}`,
```
GregWeil: I think this will prevent their tokens being filtered out before they get to sentry

src/presenters/current-user.jsx
```diff
@@ -191,12 +197,14 @@ class CurrentUserManager extends React.Component {
...
-          newSharedUser,
-        );
+        addBreadcrumb({
+          level: 'info',
+          message: `Fixed shared cachedUser. Was ${JSON.stringify(sharedUser)}`,
+        });
+        addBreadcrumb({
+          level: 'info',
+          message: `New shared cachedUser: ${JSON.stringify(newSharedUser)}`,
+        });
```
GregWeil: Can you add a console.log back that just prints out id/login?
I think the persistentToken filtering applies here too

clottman: What's the goal of the console.logging? Do we expect that users might see errors in the console that wouldn't get to Sentry?

src/presenters/current-user.jsx
clottman: Oh right - I was thinking the `beforeBreadcrumb` filter would catch it, but that's only for `console`s.

src/presenters/current-user.jsx
GregWeil: Not sure what the goal is, leaving it just in sentry is fine.

PR #268, opened 2019-02-22
Bug fix for displaying collections from soft-deleted teams

ttseng:
Changes are in https://glitch.com/edit/#!/excellent-root?path=src/presenters/pop-overs/add-project-to-collection-pop.jsx:169:89

I check to see any instances where there is no corresponding user team for a collection\ (an indication that the team has been deleted) and prevent the collection from being displayed in the pop-over.

to replicate the bug on production - create a team, add a collection to that team, and then delete said team.  then, if you try to add a project to a collection, you get an error.

then go to https://excellent-root.glitch.me/ and check that the add-to-collection pop works for you.

(note, isaac is working on fixing the API endpoint to remove collections from deleted teams from the get collections response)

### DIFF
--- a/src/presenters/pop-overs/add-project-to-collection-pop.jsx
+++ b/src/presenters/pop-overs/add-project-to-collection-pop.jsx
@@ -151,8 +151,8 @@ class AddProjectToCollectionPop extends React.Component {
 
   async loadCollections() {
     try {
-      const{ data: allCollections } = await this.props.api.get(`collections/?userId=${this.props.currentUser.id}&includeTeams=true`);
-      let deletedCollections = []; //collections from deleted teams
+      const { data: allCollections } = await this.props.api.get(`collections/?userId=${this.props.currentUser.id}&includeTeams=true`);
+      const deletedCollectionIds = []; // collections from deleted teams
       // add user / team to each collection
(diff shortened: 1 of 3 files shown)


src/presenters/pop-overs/add-project-to-collection-pop.jsx
```diff
@@ -152,15 +152,22 @@ class AddProjectToCollectionPop extends React.Component {
   async loadCollections() {
     try {
       const { data: allCollections } = await this.props.api.get(`collections/?userId=${this.props.currentUser.id}&includeTeams=true`);
+      const deletedCollectionIds = []; // collections from deleted teams
       // add user / team to each collection
       allCollections.forEach((collection) => {
         if (collection.teamId === -1) {
           collection.user = this.props.currentUser;
         } else {
           collection.team = this.props.currentUser.teams.find(userTeam => userTeam.id === collection.teamId);
```
clottman: Is the assumption here that deleted teams are included in as a collection.teamId but not included in the currentUser.teams array?

ttseng: yeah, exactly.  the currentUser teams filters out deleted teams.

PR #269, opened 2019-02-22
.jsx -> .js

clottman:
https://glitch.com/~acute-swift

~I'm new here, should I name my file .js or .jsx?~

Oh, all the files use .js! Guess I'll name my file greatNewFeature.js


jennschiffer:
bye bye, ❌

PR #271, opened 2019-02-22
Quixotic rocket: more legible popovercontainer

GregWeil:
also upgraded a bunch of dependencies


src/presenters/pop-overs/popover-container.jsx
```diff
@@ -10,92 +10,59 @@ popover pans, which have straight-walled sides rather than angled.
...
     );
   }
-}
+  const before = outer ? outer(props) : null;
 
+  return (
+    <>
+      {before}
+      <MonitoredComponent
+        // disableOnClickOutside={!visible}
```
GregWeil: It doesn't seem to handle this changing at all. Always watching for clicks seems ok though, it doesn't do much.

clottman: Probably want to remove this?

src/presenters/pop-overs/popover-container.jsx
```diff
@@ -10,92 +10,59 @@ popover pans, which have straight-walled sides rather than angled.
...
-      handleClickOutside: () => handleClickOutside,
-      excludeScrollbar: true,
-    };
-    this.MonitoredComponent = onClickOutside(Wrapper, clickOutsideConfig);
-  }
 
+class UnmonitoredComponent extends React.Component {
   handleClickOutside(event) {
     // On keyup events, only hide the popup if it was the Escape key
     if (event.type === 'keyup' && !['Escape', 'Esc'].includes(event.key)) {
```
clottman: Maybe this is out of scope, but does this onkeyup stuff work at all? I don't think react-onclickoutside will ever pass in keyup events. https://github.com/Pomax/react-onclickoutside/issues/262
In any case it doesn't seem to work when I test it; maybe we should remove the misleading code that makes it seem like it does?

GregWeil: Hmm, it gets keyups if you click in the popover. I'll change it to use a normal listener

GregWeil:
Hmm, it seems like instead of merging I somehow ended up with all of the same changes copied into my branch

GregWeil:
Fixed, but not sure how that happened. I got the merge commit message and everything!

PR #273, opened 2019-02-23
Glitch OAuth Page

ant-glitch:
**The Remix Page**
https://gamy-raja.glitch.me/signin

**Disclaimer**
This page is meant to be a temporary stand-in for a, hopefully, less odd-looking page with much less copy/pasta. I plan to create a case for the Community team in Manuscript, but I need this particular page soon-ish to make Discourse OAuth work.

**What Does It Do**

- This page is called from the Glitch API server.
- When it is called, the URL includes some very important query params.
- This page, then...

1. Captures the query parameters in a state variable.
1. Checks to see if the user is logged in by checking the cached user in local storage.
1. If the user is logged in, send them back to the Glitch API server with (a) the original query params (b) an additional authorization query param which is needed to complete the OAuth dance.
1. If the user is not logged in, allow them to login via email.
1. Once the user inputs the correct sign-in code, send them back to the Glitch API server with all the needed query params.


cori:
Out of curiosity, will this OAuth flow support GitHub and Facebook logins as well eventually?

ant-glitch:
> Out of curiosity, will this OAuth flow support GitHub and Facebook logins as well eventually?

If a user has logged in via any method prior to attempting to sign in with Glitch from Discourse, then we just pick up their token and continue completing the OAuth process.

I didn't choose to support GitHub/Facebook OAuth because it seemed odd to implement something that basically says Glitch is verifying your identity to Discourse by verifying your identity on GitHub.

If not having GitHub/Facebook auth in this particular instance proves to be an issue though, it wouldn't require any changes for the Infra/Platform codebase, but it would require the Community team to add them to this page and make sure that the user is redirected correctly after authentication.

src/presenters/pages/sign-in.jsx
```diff
@@ -0,0 +1,254 @@
...
+            </>
+          }
+        </section>
+      </dialog>
+    );
+  }
+}
+
+const redirectToOauthDialog = (queryParams) => {
+  const cachedUser = JSON.parse(window.localStorage.cachedUser);
```
clottman: Add a TODO here about handling when window.localStorage is not available? (Some versions of Chrome for Android, for example)

clottman:
I'll probably have more comments when this is closer to being production-ready but for now it looks good enough.

clottman:
To fix your eslint errors (required before merge), you can run `eslint filename --fix` in the console or use the Prettier extension. (If you don't have it yet and don't expect to be doing a lot of development in the Glitch editor on ~community, just ping me and I'll run it on that file for you) 
Looks like there might be a few you'll have to fix manually, too. Lint errors show up in the logs after your webpack build completes. (Not in the editor gutter unfortunately.)

cori:
> I didn't choose to support GitHub/Facebook OAuth because it seemed odd to implement something that basically says Glitch is verifying your identity to Discourse by verifying your identity on GitHub.

I can understand that perspective. I think this is fine until we see how the community responds.

PR #275, opened 2019-02-25
Relieved paddleboat

ttseng:
https://relieved-paddleboat.glitch.me/

Two small CSS bug fixes
* For featured collections, show row of 3 projects on medium sized screens, rather than two rows. (fixed version on left)
image

* In search results, team and user info was getting pushed to a new line (fixed version on left)
image


GregWeil:
image

If you resize to the right width it'll still split them up

ttseng:
thanks for catching that @GregWeil! changed it the width to `31%` and it seems to be working fine.

PR #276, opened 2019-02-25
Succulent wolverine - some bug fixes

clottman:
https://succulent-wolverine.glitch.me/ 
https://glitch.com/~succulent-wolverine

- .babelrc changes: makes sure we get some polyfills for the browsers that error the most
  - Eventually it would be nice if we can load these conditionally based on browser/feature detects. For the short term I think making things work should be a priority over bundle size. 
- DefaultAvatar: clean this component up; hopefully avoid a timing issue we seemed to have had where the `ref` was still valid, but the SVG wasn't there. 
  - Plus: removes a package from our bundle! (react-svg-inline)
- collection.js: That prop is not actually required
- login.js: some browsers don't support window.origin; since we always know what window.origin is supposed to be we can just hard code it.


src/components/collection/defaultAvatar.js
```diff
@@ -0,0 +1,55 @@
...
+        stroke="#000"
+        strokeWidth={4}
+        fill="#FFF"
+      />
+    </g>
+  </svg>
+);
+
+DefaultAvatar.defaultProps = {
+  backgroundFillColor: '#45C1F7',
```
GregWeil: Does this need a proptypes, or is it implied by defaultprops?

src/presenters/includes/profile.js
```diff
@@ -95,28 +82,24 @@ CoverContainer.defaultProps = {
...
-      </div>
-    </InfoContainer>
-    {!!teams && !!teams.length && (
-      <div className="teams-information">
-        <TeamsList teams={teams} />
-      </div>
-    )}
-  </CoverContainer>
-);
+export class ProfileContainer extends React.PureComponent {
```
clottman: Hoping to address too much recursion errors when a user uploads a new cover photo or profile picture. PureComponent does a shallow comparison instead of deep comparison

src/presenters/pages/login.js
```diff
@@ -20,7 +20,7 @@ function notifyParent(message = {}) {
   // Specifically target our same origin;
   // we're only communicating between the editor and its corresponding ~community site,
   // not across other environments.
-  const sameOrigin = window.origin;
+  const sameOrigin = 'https://glitch.com';
```
GregWeil: This can use `APP_URL`

package.json
```diff
@@ -3,8 +3,8 @@
   "version": "0.0.1",
   "description": "Find cool projects and people using Glitch",
   "scripts": {
-    "prestart": "nodemon --watch server --watch shared --exec 'eslint --config server/.eslintrc.server.js server shared webpack.config.js' &",
-    "start": "nodemon --watch server --watch shared --watch views ./server/server.js",
+    "prestart": "node_modules/.bin/nodemon --watch server --watch shared --exec 'eslint --config server/.eslintrc.server.js server shared webpack.config.js' &",
+    "start": "node_modules/.bin/nodemon --watch server --watch shared --watch views ./server/server.js",
```
GregWeil: Why is this changed?

src/presenters/includes/profile.js
```diff
@@ -9,22 +9,17 @@ import Button from '../../components/buttons/button';
...
   <div className="upload-image-buttons">
-    { !!uploadImage && (
+    {!!uploadImage && (
       <TrackClick name={`Upload ${name}`}>
-
         <Button size="small" type="tertiary" onClick={uploadImage}>
-            Upload
-          {' '}
-          {name}
+          Upload {name}
```
GregWeil: Putting two expressions on the same line breaks an eslint rule we have, but I'd be in favor of disabling it

package.json
clottman: I assumed it was changed in ~community right before I remixed - I think one of the infrastructure team members tried it to fix node_modules. If we can safely change back, I'll do that.

src/presenters/includes/profile.js
clottman: Me too - that one seems to cause more less-readable code than it's worth.

src/components/collection/defaultAvatar.js
clottman: Eslint didn't seem to hate it as-is, but I can add it in.

PR #277, opened 2019-02-26
Superb guide: Allow many expressions per line, clean up line breaks

GregWeil:
I also made api props required in a bunch of places. The project item and project options pop do not require the api though, because we don't give the api on a few pages (project page related projects, index page)

### DIFF
--- a/src/.eslintrc.js
+++ b/src/.eslintrc.js
@@ -52,13 +52,10 @@ module.exports = {
     }],
     // React hooks config
     "react-hooks/rules-of-hooks": "error",
-    "react/jsx-one-expression-per-line": [ERROR, {"allow": "literal"}],
+    "react/jsx-one-expression-per-line": [OFF],
     "react/jsx-filename-extension": [ERROR, { "extensions": [".js"] }]
   },
   settings: {
-    react: {
-      version: "16.8.1" // Should match package.json
-    },
     "import/ignore": ['sentry'],
   },
 };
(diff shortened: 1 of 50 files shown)


src/.eslintrc.js
```diff
@@ -52,13 +52,10 @@ module.exports = {
     }],
     // React hooks config
     "react-hooks/rules-of-hooks": "error",
-    "react/jsx-one-expression-per-line": [ERROR, {"allow": "literal"}],
+    "react/jsx-one-expression-per-line": [OFF],
     "react/jsx-filename-extension": [ERROR, { "extensions": [".js"] }]
   },
   settings: {
-    react: {
```
clottman: Do we know why this was there originally?

src/presenters/pages/error.js
```diff
@@ -84,9 +80,7 @@ const oauthImageUrl = 'https://cdn.glitch.com/8ae9b195-ef39-406b-aee0-764888d156
 
 export const OauthErrorPage = ({ api, title, description }) => (
   <Layout api={api}>
-    <Helmet>
-      <title>🔑 {title}</title> {/* eslint-disable-line */}
-    </Helmet>
+    <Helmet title={`🔑 ${title}`} />
```
clottman: I think the eslint-disable-line here was about the emoji that doesn't have a `role` (which doesn't make sense in a <title>) - you might still need it.

src/.eslintrc.js
GregWeil: No, that value defaults to 'detect' which seems like it should work just fine.

src/presenters/pages/error.js
GregWeil: It only checks for roles within jsx content. That's an attribute which it's fine with being just a string

PR #282, opened 2019-02-27
Superb guide: Refactors to the current user manager

GregWeil:
This should fix the main cause of invalid cached user errors. It happens because when you log in with a second glitch.com window open the other window tries to reload the no longer valid anon user. While the reload request is in progress it gets updated props with the signed in user. The request fails because the anon user was merged, and the user fix gets applied to the updated signed in user, showing as a correction from the anon user to the signed in user.

### DIFF
--- a/src/presenters/current-user.js
+++ b/src/presenters/current-user.js
@@ -199,7 +199,7 @@ class CurrentUserManager extends React.Component {
       this.props.setSharedUser(sharedUser);
     }
 
-    // Check if we have to clear the cache
+    // Check if we have to clear the cached user
     if (!usersMatch(sharedUser, this.props.cachedUser)) {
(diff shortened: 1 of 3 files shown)


src/presenters/current-user.js
```diff
@@ -124,14 +124,20 @@ class CurrentUserManager extends React.Component {
       !usersMatch(cachedUser, sharedUser)
       || !usersMatch(sharedUser, prev.sharedUser)
     ) {
-      this.load();
+      // delay loading a moment so both items from storage have a chance to update
+      setTimeout(() => this.load(), 1);
```
GregWeil: This is the real fix, it gives a moment for both localstorage keys to get updated rather than running on partially outdated data

PR #285, opened 2019-02-28
Superb guide: Probably fix the remaining cachedUser errors

GregWeil:
We send the load user request on page load, even if you're on the signin page. Based on the remaining events, it looks like the anon user's load request sometimes gets queued after the sign in. When it goes through the anon user doesn't exist. That triggers a user fix even though the anon user isn't relevant any more.

Before we try to fix the user, check if the user was changed. If they changed try loading the new user instead of doing a fix.

### DIFF
--- a/src/presenters/current-user.js
+++ b/src/presenters/current-user.js
@@ -206,11 +206,11 @@ class CurrentUserManager extends React.Component {
 
     const newCachedUser = await this.getCachedUser();
     if (newCachedUser === 'error') {
-      // Looks like our sharedUser is bad, make sure it wasn't overwritten since we last checked
+      // Looks like our sharedUser is bad, make sure it wasn't changed since we read it
       // Anon users get their token and id deleted when they're merged into a user on sign in
       // If it did change then quit out and let componentDidUpdate sort it out
       if (usersMatch(sharedUser, this.props.sharedUser)) {
-        // The user wasn't changed externally, so we do need to fix it
+        // The user wasn't changed, so we need to fix it
         this.setState({ fetched: false });
         const newSharedUser = await this.getSharedUser();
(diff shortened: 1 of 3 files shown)


src/presenters/pages/login.js
```diff
@@ -51,7 +51,7 @@ class LoginPage extends React.Component {
         throw new Error(`Bad user id (${data.id}) after ${provider} login`);
       }
 
-      console.log('LOGGED IN', data);
+      console.log('LOGGED IN', data.id);
```
clottman: 👏

src/presenters/current-user.js
```diff
@@ -206,21 +206,25 @@ class CurrentUserManager extends React.Component {
...
+        console.log(`Fixed shared cachedUser from ${sharedUser.id} to ${newSharedUser && newSharedUser.id}`);
+        addBreadcrumb({
+          level: 'info',
+          message: `Fixed shared cachedUser. Was ${JSON.stringify(sharedUser)}`,
+        });
+        addBreadcrumb({
+          level: 'info',
+          message: `New shared cachedUser: ${JSON.stringify(newSharedUser)}`,
+        });
+        captureMessage('Invalid cachedUser');
```
clottman: Do we still need to send to Sentry if we're at this point?

GregWeil: I want it sending to sentry so we can catch those sudden spikes from strange race conditions. They aren't breaking the site, but they usually cause a bunch of extra requests that aren't needed.

GregWeil: As an aside, the new api should let us simplify this a lot. It lets us get a user by their token, which means we won't have to worry about tracking/recovering their id

PR #286, opened 2019-03-01
Tooltip component

taravancil:
This is a big one! It adds `src/components/tooltip-container.js` and `src/components/tooltip.styl`, and replaces all of the old tooltip instances with this new component.

My goal was to make our tooltips more accessible, and to make it difficult to compose them incorrectly. There are two different kinds of tooltips: "action" and "informational", and I focused on making sure that each type's content exists in the DOM at the right time, that it has the appropriate `aria-` attributes, etc. See [my research notes about how to do tooltips](https://www.notion.so/glitch/Tooltips-6564f2b7c6d24df18c05b72fcfaadf60).

This implementation is likely not perfectly a11y-friendly, but it's definitely better than what we had! I still have some questions about informational tooltips, but I think we should merge this before we answer those.

Here's what composing a tooltip looks like now:

```jsx
// An action tooltip, use with buttons, links, and other things that Do Stuff on click

<TooltipContainer
  id="a-unique-id"
  type="action"
  target={<Button>Click me</Button>}
  align={["top", "left"]}
  tooltip="I'm an action tooltip"
/>

// An information tooltip, use them with targets that don't Do Stuff on click
<TooltipContainer
  id="a-unique-id"
  type="information"
  target={<img src="/butts.png" />}
  tooltip="I'm an information tooltip"
/>
```
I've tested this in Firefox, Chrome, and Safari, but it would be great to get some extra eyes on it, especially if you use Edge or IE. You can see tooltips in action in [storybook](https://tooltips.glitch.me/storybook), or you can review the individual tooltips I ported over:

- Project visibility button/badge: open a project you own, one you don't own, like ~[discourse-poster](https://tooltips.glitch.me/~discourse-poster)
- New-stuff dog
- Avatars
- Verified badge: [@slack-blueprints](https://tooltips.glitch.me/@slack-blueprints)
- Add team members by domain controls: see a team you're an admin on or [@community](https://tooltips.glitch.me/@community)
- Team analytics remix avatars: open [@community](https://tooltips.glitch.me/@community), filter by ~community, see remix avatars
- User options in header


src/components/tooltip.styl
```diff
@@ -0,0 +1,101 @@
+@require '../../styles/globals'
+
+tooltip-background = rgba(0,0,0,0.95)
+background-color = white
```
ttseng: Instead of introducing another background-color variable here, we can just use `primary-background` from `global.styl`

src/components/tooltip.styl
```diff
@@ -0,0 +1,101 @@
...
+secondary = #9B9B9B
```
ttseng: I don't think this is used, so we can remove it

ttseng:
All the tooltips I tested are working well!

A few notes:
- suggest renaming the type `information` to `info` since we use `action` and `info` elsewhere (for example, to distinguish between pop-over section types)
- seems like `tooltip-container.js` and `tooltip.styl` should be in their own folder (`/src/components/tooltips`)

Looks like there's a regression in terms of pop-over font color on selection (for example:

`collection-result-item`
image

`project-result-item`
image

`user-result-item`
image

`time-frame-item`
image

Also collection avatar is not up to date - maybe pulling the latest from master will fix these issues?

src/components/tooltip-container.js
```diff
@@ -0,0 +1,92 @@
...
+    });
+  } else if (type === 'information') {
+    // information tooltips are visible on hover and focus, they provide supplementary info
+    // they should be empty when not "visible", and populated when they are
+
+    role = 'status';
+    extendedTarget = React.cloneElement(target, {
+      'aria-describedby': id,
+    });
+  }
```
sheridanvk: It's not really clear to me when I should use an action versus an information tooltip. e.g. ProjectRemixItem on the Team Analytics Page is an information tooltip, but it seems to act pretty identically to the Avatar tooltip, which is an action tooltip. Both provide additional information on hover (remix name versus user name) 
screenshot 2019-03-04 at 10 41 22 am
screenshot 2019-03-04 at 10 40 37 am

Could you clarify what the difference is?

src/components/tooltip.styl
```diff
@@ -0,0 +1,101 @@
...
+
+.touch .tooltip
```
sheridanvk: where does this touch class get added?

src/presenters/includes/team-users.js
```diff
@@ -88,14 +89,21 @@ export const WhitelistedDomain = ({ domain, setDomain }) => {
           open={visible}
           className="popover-container whitelisted-domain-container"
         >
-          <summary data-tooltip={!visible ? tooltip : null}>
-            <WhitelistedDomainIcon domain={domain} />
+          <summary>
+            <TooltipContainer
+              id="whitelisted-domain-tooltip"
+              type="action"
+              tooltip={visible ? null : tooltip}
```
sheridanvk: Do we still need this logic at this level, given you have a lot of this baked into the Tooltip Component itself? This particular situation seems like a weird special case, so perhaps it does still need to be here, it's just surprising to me that we need to pass logic into the component, as it seemed pretty well spec'ed otherwise.

stories/index.js
```diff
@@ -1,20 +1,63 @@
...
+storiesOf("TooltipContainer", module)
+  .add("action", () => (
+    <div style={{margin: '70px'}}>
+      <TooltipContainer
+        type="action"
+        id="a-unique-id"
+        target={<Button>Hover or focus me</Button>}
+        tooltip="I'm an action tooltip"
+      />
+    </div>
```
sheridanvk: just curious, do you know why the stories are showing the tooltip without needing to hover/focus? I would have expected this to display in the same way as it does when it's embedded into the community site

sheridanvk:
This looks great overall! Agree with Tiff's feedback, and then I just have some inline questions. 

One final overall question - what made you decide to show and hide tooltips with opacity: 0/1 versus display: none/whatever? I don't have a strong opinion either way, I was jut curious what the tradeoffs there were.

src/components/tooltip.styl
taravancil: I'm not sure; it's leftover from the old tooltip styles. I assume it's a top-level class on the page that allows us to target touch behavior of a bunch of different elements.

stories/index.js
taravancil: Hm, it is? I wasn't seeing that, and if you are, that's definitely not right!

taravancil:
> what made you decide to show and hide tooltips with opacity: 0/1 versus display: none/whatever?

That was an accessibility decision. So for example, action tooltips effectively serve as labels for the `button`, `link`s, etc. that they describe. Take the avatar component for example: the way you and I navigate, we can deduce that clicking on Tiff's avatar will take you to Tiff's profile. For low-sight people using screen readers, they need the tooltip nodes to always be "visible" within the DOM, even when they're hidden from sight. `display: none` removes nodes from the visual flow of the page and makes them inaccessible to screenreaders.

There are of course more subtleties but that's the gist of it 😄

src/components/tooltip-container.js
taravancil: Sure! The main difference is that an action tooltip has a `target` that Does A Thing when you click it (a button that opens a dropdown). Info tooltips have a `target` that **does not** Do A Thing (a project avatar with a tooltip that shows the project name).

So basically an action tooltip is a primary label for an action, whereas an info tooltip provides supplementary info, and clicking its `target` does not trigger an action.

Our action and info tooltips will probably always look the same and have a similar interface for composing them, but because they provide different functionality, in order to be accessible they need to be represented differently in the DOM.

I'm glad you mentioned it though, because the remix avatar *should* be an action tooltip! Clicking on its `target` takes you to the project page.

stories/index.js
taravancil: I *am* seeing this now, odd! Looking into it.

src/presenters/includes/team-users.js
taravancil: This is because the tooltip needs to be hidden when the "Add to team" popover is open. `visible` indicates whether the popover is open. If it's open, the `tooltip` prop will be `null` and `TooltipContainer` won't render the tooltip, even though you're hovering its `target`

taravancil:
I've pushed up some further changes:

- Removed the `background-color` global from our styles and switched over to `primary-background`, this fixes the regressions that @ttseng shared above
- Removed another unused global in `tooltip.styl`
- Converted team analytics remix tooltip from an `action` tooltip to an `info` tooltip
- Renamed `information` tooltips to `info`  tooltips to remain consistent with how we describe other component types

src/components/tooltip-container.js
sheridanvk: Aha that totally makes sense now - the inconsistency was the part that was confusing me. Thanks!

src/presenters/includes/team-users.js
sheridanvk: Ahh right, I see :( I think we should consider getting rid of this popover at some point and moving the action on the team to somewhere else, as it's really undiscoverable otherwise. Sounds good for now though :D

sheridanvk:
> > what made you decide to show and hide tooltips with opacity: 0/1 versus display: none/whatever?
> 
> That was an accessibility decision. So for example, action tooltips effectively serve as labels for the `button`, `link`s, etc. that they describe. Take the avatar component for example: the way you and I navigate, we can deduce that clicking on Tiff's avatar will take you to Tiff's profile. For low-sight people using screen readers, they need the tooltip nodes to always be "visible" within the DOM, even when they're hidden from sight. `display: none` removes nodes from the visual flow of the page and makes them inaccessible to screenreaders.
> 
> There are of course more subtleties but that's the gist of it 😄

that's great to know, thank you! makes sense :)

PR #287, opened 2019-03-04
Superb guide: query param fixes for the oauth page

GregWeil:
Clean up url param handling in the new oauth page. I also added a polyfill for UrlSearchParams, but I haven't been able to test it. (I'm deactivated in browserstack)

@ant-glitch


GregWeil:
Removed the polyfill since it breaks our editor embed support check

PR #289, opened 2019-03-04
Dented supernova - handle sorting of undeleted projects + sorting of user projects by updatedAt on profile pages

ttseng:
https://dented-supernova.glitch.me/

Currently, when the user undeletes a project, the undeleted project is shown as their most recent project.  However, upon refreshing the page, the undeleted project is no longer seen at the top of the list as user projects on the profile page are not automatically sorted by when they are last updated.

To test:
* open https://dented-supernova.glitch.me/ and try undeleting one of your projects.  Then refresh the page and check that the undeleted project still appears at the top of the list.

Manuscript case: https://glitch.manuscript.com/f/cases/3327976

### DIFF
--- a/src/presenters/user-editor.js
+++ b/src/presenters/user-editor.js
@@ -124,6 +124,7 @@ class UserEditor extends React.Component {
         console.warn('failed to rename project on undelete', e);
       }
     }
+    console.log('undeleted ', data);
     this.setState(({ projects, _deletedProjects }) => ({
       projects: [data, ...projects],
       _deletedProjects: _deletedProjects.filter(p => p.id !== id),
--- a/src/presenters/user-editor.js
+++ b/src/presenters/user-editor.js
@@ -124,7 +124,8 @@ class UserEditor extends React.Component {
         console.warn('failed to rename project on undelete', e);
       }
     }
-    console.log('undeleted ', data);
+    // temp set undeleted project updatedAt to now, while it's actually updating behind the scenes
+    data.updatedAt = Date.now();
     this.setState(({ projects, _deletedProjects }) => ({
       projects: [data, ...projects],
       _deletedProjects: _deletedProjects.filter(p => p.id !== id),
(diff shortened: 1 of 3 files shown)


src/presenters/user-editor.js
```diff
@@ -124,6 +124,8 @@ class UserEditor extends React.Component {
         console.warn('failed to rename project on undelete', e);
       }
     }
+    // temp set undeleted project updatedAt to now, while it's actually updating behind the scenes
+    data.updatedAt = Date.now();
```
clottman: This will return the date in seconds since the unix epoch - one of my projects' updatedAt is formatted like: 
```
"updatedAt":"2019-03-04T21:48:33.585Z"
```
I think you probably want `new Date()` ?

clottman: Also is this change necessary? We're doing the sort in ProjectsLoader and not updating ProjectsLoader.projects, so I'm not sure what component would be checking updatedAt on the newly added project.

ttseng: I think it's necessary to update the undeleted projects `updatedAt` because the ProjectsLoader in `user.js` is given the user's recent projects , now ordered by updatedAt date.  

If we don't update the date, `data` from `undeleteProject` uses the `updatedAt` date from the project at the time it was deleted, so the project doesn't appear at the top of the list.

clottman: It's really hard to follow the data flow between the two components, but it's working for me in ~dashing-caribou without this line.

ttseng: hm, it's not working for me. did you try undeleting something you deleted a long time ago?

clottman: Yup

clottman: Hmm - now it's not working for me either. Maybe I was testing on an old build. Never mind then!

ttseng:
Added the space to the pop-over title for adding projects to a collection: https://glitch.manuscript.com/f/cases/3327962/
image

PR #292, opened 2019-03-06
Superb guide: Use the rest api for teams pages

GregWeil:
https://superb-guide.glitch.me/@glitch ⚡

This is a big improvement for performance, but there's a few odd bits that I'll point out in comments.

There's also a new shared/api.js file. This has a few (but probably more as time goes on) utility functions for working with the api.

### DIFF
--- a/server/api.js
+++ b/server/api.js
@@ -23,7 +23,7 @@ async function getFromCacheOrApi(id, cache, api) {
     const value = await promise;
     return value;
   } catch (error) {
-    console.error(error);
+    console.error(error.message);
     return null;
   }
 }
--- a/server/api.js
+++ b/server/api.js
@@ -5,6 +5,7 @@ const {Cache} = require("memory-cache");
 const dayjs = require("dayjs");
 
 const {API_URL} = require("./constants").current;
+const {getFromApi} = require("../shared/api");
 
 const CACHE_TIMEOUT = dayjs.convert(15, 'minutes', 'ms');
 
@@ -35,8 +36,8 @@ const api = axios.create({
 
 async function getProjectFromApi(domain) {
   try {
-    const response = await api.get(`/projects/${domain}`);
-    return response.data;
+    const data = await getFromApi(api, `v1/projects/by/domain?domain=${domain}`);
+    return data[domain];
   } catch (error) {
     if (error.response && error.response.status === 404) {
       return null;
@@ -47,8 +48,8 @@ async function getProjectFromApi(domain) {
 
 async function getTeamFromApi(url) {
   try {
-    const response = await api.get(`/teams/byUrl/${url}`);
-    return response.data;
+    const data = await getFromApi(api, `v1/teams/by/url?url=${url}`);
+    return data[url];
   } catch (error) {
     if (error.response && error.response.status === 404) {
       return null;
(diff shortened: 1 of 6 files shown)


server/api.js
```diff
@@ -23,7 +24,7 @@ async function getFromCacheOrApi(id, cache, api) {
     const value = await promise;
     return value;
   } catch (error) {
-    console.error(error);
+    console.error(error.toString());
```
GregWeil: Way less noise when something times out, since it doesn't print the whole stack trace and response object

shared/api.js
```diff
@@ -0,0 +1,39 @@
...
+
+const getFromApi = async (api, url) => {
+  const { data } = await api.get(url);
+  return data;
+}
+
+const getSingleItem = async (api, url, key) => {
+  // The api is case insensitive when getting by url/login
+  // but it'll return an object keyed using the correct case
+  // so do a case insensitive lookup in the dict it returns
```
GregWeil: https://glitch.manuscript.com/f/cases/3328008/

src/presenters/pages/team-or-user.js
```diff
@@ -46,20 +47,29 @@ const getTeamById = async (api, id) => {
...
+      getAllPages(api, `v1/teams/by/id/users?id=${team.id}&orderKey=createdAt&orderDirection=ASC&limit=100`),
+      getAllPages(api, `v1/teams/by/id/pinnedProjects?id=${team.id}&orderKey=createdAt&orderDirection=DESC&limit=100`),
+      getAllPages(api, `v1/teams/by/id/projects?id=${team.id}&orderKey=createdAt&orderDirection=DESC&limit=100`),
+      getAllPages(api, `v1/teams/by/id/collections?id=${team.id}&orderKey=createdAt&orderDirection=DESC&limit=100`),
+    ]);
+
+    team.users = users.sort((a, b) => new Date(a.teamPermission.updatedAt) - new Date(b.teamPermission.updatedAt));
+    team.projects = projects.sort((a, b) => new Date(b.updatedAt) - new Date(a.updatedAt));
+    team.teamPins = pinnedProjects.map(project => ({ projectId: project.id }));
+    team.collections = collections;
```
GregWeil: https://glitch.manuscript.com/f/cases/3328004/
The api doesn't support sorting by updatedAt, so do the sort right here

This is stuffing the data into the layout of the old api, we can clean this up when every page is using the new api.

shared/api.js
```diff
@@ -0,0 +1,39 @@
+/// Api helper functions
```
clottman: Do edits in this file trigger a full webpack rebuild? If so I'd suggest moving this to src/shared or something so it doesn't get picked up by the server webpack.

src/presenters/pages/team-or-user.js
```diff
@@ -46,20 +47,29 @@ const getTeamById = async (api, id) => {
 };
 
 const getTeam = async (api, name) => {
-  const team = await getOrNull(api, `/teams/byUrl/${name}`);
+  const team = await getSingleItem(api, `v1/teams/by/url?url=${name}`, name);
+  if (team) {
+    const [users, pinnedProjects, projects, collections] = await Promise.all([
+      // load all users, need to handle pagination
+      getAllPages(api, `v1/teams/by/id/users?id=${team.id}&orderKey=createdAt&orderDirection=ASC&limit=100`),
```
clottman: Are any of these slower than others? For example maybe a team has a ton of projects that takes a while to return. Would it be helpful to do some of these calls in a separate async function so we can load a bit more progressively?

shared/api.js
clottman: Oh - I see it's used in server/api.js too. Never mind!

src/presenters/pages/team-or-user.js
GregWeil: They all load very quickly. We only actually need the first twelve projects on first page render, and we could wait on the rest until you click show more. That gets tricky since we need to make sure we're loading the first twelve non pinned, not featured projects, and the add project and remove team user menus both need the full list of projects.

PR #293, opened 2019-03-07
Relic Clutch: Use the rest api for index collections

keithk:
https://relic-clutch.glitch.me/@glitch ⚡

This builds on the PR that @GregWeil opened #292 using the same helper tools to replace the API calls for collections on the home page with the new API get requests.

There is one caveat which I'll call out in the comments.


src/presenters/featured-collections.js
```diff
@@ -72,10 +74,8 @@ CollectionWide.propTypes = {
 
 const loadCollection = async (api, { owner, name }) => {
   try {
-    const { data: collectionId } = await api.get(
-      `collections/${owner}/${name}`,
-    );
-    const { data: collection } = await api.get(`collections/${collectionId}`);
+    const collection = await getSingleItem(api, `/v1/collections/by/fullUrl?fullUrl=${owner}/${name}`, `${owner}/${name}`);
+    collection.projects = await getSingleItem(api, `/v1/collections/by/fullUrl/projects?limit=20&fullUrl=${owner}/${name}`, 'items');
     collection.projectCount = collection.projects.length;
```
keithk: Because the new API is paginated with a `limit` field, the `/collections/` endpoint doesn't (currently) return a `totalItems` or similar field.

I checked the past featured collections and the maximum number of projects they have, and set the limit to 20 so right now the counting will be correct, but in the future replacing this with a total number from `/collections/` would be preferable (so we can have MASSIVE collections 💪 )

clottman: Makes sense, is there a case for the infrastructure team about getting total count from /collections/ ?

GregWeil: If there are more thank 20 projects, could you have it just say 'View all projects' instead of giving a number

GregWeil: (I don't think there is a case, could you open one)

GregWeil:
Collections don't return the team name/avatar (just their id and url). It'll need an extra request to get that info

clottman:
@GregWeil 
> Collections don't return the team name/avatar (just their id and url). It'll need an extra request to get that info 
I see what you're saying but I can't see the visual difference between this remix and superb-guide - is the UI broken somewhere due to needing the extra request?

keithk:
Ahh you're right @GregWeil . @clottman it looks like there isn't a difference, but it's showing a default avatar for the glitch team on relic-clutch vs the glitch avatar
screen shot 2019-03-07 at 11 39 14 am
screen shot 2019-03-07 at 11 39 18 am

I'll fix this up!

src/presenters/featured-collections.js
keithk: @GregWeil I made a case to get a totalItems on collection, right now for the index page this is fine (because it only displays 3 random ones) but for the project pages yeah I think that'd be a good solution

src/presenters/featured-collections.js
```diff
@@ -72,10 +73,9 @@ CollectionWide.propTypes = {
 
 const loadCollection = async (api, { owner, name }) => {
   try {
-    const { data: collectionId } = await api.get(
-      `collections/${owner}/${name}`,
-    );
-    const { data: collection } = await api.get(`collections/${collectionId}`);
+    const collection = await getSingleItem(api, `/v1/collections/by/fullUrl?fullUrl=${owner}/${name}`, `${owner}/${name}`);
+    collection.projects = await getSingleItem(api, `/v1/collections/by/fullUrl/projects?limit=20&fullUrl=${owner}/${name}`, 'items');
+    collection.team = await getSingleItem(api, `/v1/teams/by/id?id=${collection.team.id}`, collection.team.id);
```
keithk: This gets the team details now, too!

clottman: Meaning the avatar that was missing before is there now? If so this looks good to me!

PR #295, opened 2019-03-07
Octagonal windshield: Use the REST API for user pages

modernserf:
- a user with little activity: https://octagonal-windshield.glitch.me/user/271884
- a user with some activity: https://octagonal-windshield.glitch.me/@modernserf
- a user with a lot of activity: https://octagonal-windshield.glitch.me/@jennschiffer

There's a note in the code about pushing some of the API calls into the components (so they can handle their own pagination, etc) but that seems like something we can iterate on in future PRs. Even with Jenn's profile page, fetching everything at once seems to work pretty well, so its not critical that we do this right away.

There's a substantial amount of change noise introduced by running prettier -- we should probably have a separate PR that just prettifies everything in one go.


src/.eslintrc.js
```diff
@@ -31,6 +31,11 @@ module.exports = {
     semi: ["error", "always"],
     "no-debugger": WARN,
     "jsx-a11y/label-has-for": OFF, // It's been deprecated. -- https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/label-has-for.md
+    // Disable style rules handled by `prettier`:
```
modernserf: prettier and eslint:recommended disagree here; if we're committing to prettier we should let that handle the style and use eslint for actual error handling

src/presenters/pages/team-or-user.js
```diff
@@ -21,23 +21,53 @@ const getOrNull = async (api, route) => {
...
+    projects: getAllPages(api, `v1/users/by/login/projects?login=${name}&limit=100&orderKey=createdAt&orderDirection=DESC`),
+    teams: getAllPages(api, `v1/users/by/login/teams?login=${name}&limit=100&orderKey=createdAt&orderDirection=DESC`),
+    collections: getAllPages(api, `v1/users/by/login/collections?login=${name}&limit=100&orderKey=createdAt&orderDirection=DESC`),
+  });
+  return mergeUserData(data);
 };
 
 const parseTeam = (team) => {
   const ADMIN_ACCESS_LEVEL = 30;
-  const adminIds = team.teamPermissions.filter(user => user.accessLevel === ADMIN_ACCESS_LEVEL);
```
modernserf: everything in the file after this point is just formatting changes

src/presenters/pages/user.js
```diff
@@ -251,6 +204,11 @@ UserPage.propTypes = {
     avatarUrl: PropTypes.string,
     color: PropTypes.string.isRequired,
     coverColor: PropTypes.string,
+    description: PropTypes.string.isRequired,
```
modernserf: Additional fields that the component was depending on, but not specified in the prop types

src/presenters/pages/user.js
```diff
@@ -24,53 +24,30 @@ function syncPageToLogin(login) {
   history.replaceState(null, null, getLink({ login }));
 }
 
-const NameAndLogin = ({
```
modernserf: Except for the added proptypes, this is just `prettier` changes. The style we were writing before more closely matches prettier with 80 char columns instead of 150.

src/presenters/project-item.js
```diff
@@ -53,7 +53,7 @@ ProjectItem.propTypes = {
     description: PropTypes.string.isRequired,
     domain: PropTypes.string.isRequired,
     id: PropTypes.string.isRequired,
-    private: PropTypes.bool.isRequired,
```
modernserf: `private` is not included if you're not logged in, but in that case you can't see private projects anyways.

src/presenters/pages/team-or-user.js
```diff
@@ -21,23 +21,53 @@ const getOrNull = async (api, route) => {
   }
 };
 
+// TODO: this can be put in a `utils` junk drawer
```
modernserf: This block is where the meat of the changes are made

src/presenters/pages/team-or-user.js
```diff
@@ -21,23 +21,53 @@ const getOrNull = async (api, route) => {
...
+};
+
+const mergeUserData = (data) => {
+  const { user, ...rest } = data;
+  return { ...user, ...rest };
+};
+
+// TODOs:
+// this works pretty well, even for users with lots of projects,
+// but we should consider pushing the API calls down to the individiual components,
```
keithk: Yeah, I agree with this TODO. Pushing them down to the components mean that they'll load individually and usually fast enough that the user won't perceive them loading separately

src/.eslintrc.js
clottman: We're not running prettier on builds, only manually, so the eslint rules make sure that you have actually run prettier on your code before you check it in.

src/.eslintrc.js
```diff
@@ -31,6 +31,12 @@ module.exports = {
...
+    "arrow-parens": [OFF],
```
clottman: Wouldn't we want `[ERROR, "as-needed", { "requireForBlockBody": false }]` here?

src/.eslintrc.js
clottman: Instead of turning these rules off, let's configure them to match what prettier does.

src/presenters/pages/team-or-user.js
```diff
@@ -21,23 +21,53 @@ const getOrNull = async (api, route) => {
   }
 };
 
+// TODO: this can be put in a `utils` junk drawer
+const allByKeys = async (objOfPromises) => {
```
clottman: This probably belongs in shared/api.js

src/presenters/pages/team-or-user.js
clottman: +1 to the todo, @modernserf perhaps make a case for it so we remember to go back and do it?

src/presenters/pages/user.js
```diff
@@ -24,53 +24,30 @@ function syncPageToLogin(login) {
...
-      return (
-        <h1 className="login">
-          @{login}
-        </h1>
-      );
+      return <h1 className="login">@{login}</h1>;
     }
     return (
       <>
-        <h1 className="username">
```
clottman: Changes like these were made when we had eslint configured to only allow one expression per line, which we quickly decided made things messier than they'd previously been. So this just gets things back to how they were until relatively recently.

modernserf:
@clottman changing eslint rules to match prettier will require us to change _every other file_, is that what we want to do in this PR? If we're gonna do that, I'd like to do that in a separate one.

clottman:
Fair point, we can do that in a separate PR. Can you make a case to fix it which notes the eslint lines that will be disabled by this PR, so we know which ones to work on when we get there?

src/presenters/pages/team-or-user.js
clottman: We also might want to name it something like `getAllByKeys` or `getAllItems` to make it more consistent with the `getSingleItem` that exists there.

shared/api.js
```diff
@@ -32,8 +32,20 @@ const getAllPages = async (api, url) => {
   return results;
 };
 
+// like Promise.all but with an object instead of an array, e.g.
+// `let { user, projects } = await allByKeys({ user: getUser(id), projects: getProjects(id) })`
+const allByKeys = async (objOfPromises) => {
```
modernserf: I'm calling this `allByKeys` instead of `getAllByKeys` because the `get` functions are all wrappers over `api.get`, while this operates over promises.

PR #296, opened 2019-03-07
Worried deposit: team page is missing invited users

modernserf:
WIP

The REST API for teams does not seem to include the tokens for invited users.

Causes Sentry error (but no user-visible error): https://sentry.io/organizations/fog-creek-software/issues/921852558/?project=1246508&referrer=slack&statsPeriod=14d

I don't know what the next step here is. Is this critical enough that we want to roll this back?


modernserf:
I created the team `@different-order` and added the user `@justin+test1`. The invited user is not visible on the page, and doesn't appear to be in any API responses.

jennschiffer:
make sure to make a manuscript case for infra to look at - this is a regression in the api

modernserf:
fixed on the back end

PR #297, opened 2019-03-07
Use the new api to load the members of multiple projects

oliviacpu:
Manuscript link: https://glitch.manuscript.com/f/cases/3328007/
Remix link: https://glitch.com/edit/#!/determined-mechanic

Summary of changes:

Aside from changing `eslint` rules (thank you @modernserf) and noise from `prettier` my changes are made in `src/presenters/projects-loader.js` and `shared/api.js`.

There is a helper function that will join IDs into a query string, which could be useful for other people.

Then we're using the new API to fetch a group of projects, and then gathering the user IDs listed in the permissions objects and fetching user info for each project (inside of `loadProjects` in `projects-loader.js`).

💁I think we are intending to gather _all_ of the users in a given set of projects and request them though, so that's next on this story maybe, but in another PR.


src/presenters/projects-loader.js
```diff
@@ -29,43 +31,47 @@ class ProjectsLoader extends React.Component {
...
+    projects = Object.values(projects);
+    projects = projects.map(async (project) => {
+      const userIds = project.permissions.map((permission) => permission.userId);
+      const users = await getFromApi(this.props.api, `v1/users/by/id?${joinIdsToQueryString(userIds)}`);
+      return {
+        ...project,
+        users: Object.values(users),
+      };
+    });
+    projects = await Promise.all(projects);
```
clottman: What is `projects` before line 48, and what is it after? Since the `getFromApi` calls above are `await`ed I don't understand what promises are contained in each `projects[i]`.

oliviacpu: That was weird to me too.

From the projects endpoint we get an object like this:
`{
  [project_id]: {
    id: project_id,
   ...[rest of the project],
  },
  [more projects]...
}`

I grab the values from that (now it's an array of project objects) to map over and fetch users, and since it's awaiting I would expect that the project object with users added would be returned. It's giving me promises those promises though, and I'm not sure if the `async`/`await` is having problems being in map for some reason?

clottman: I'm not sure why it's like that either, but it does look kind of confusing as written so I think we should refactor out the confusing bits somehow.

oliviacpu: @clottman I talked with @modernserf about why that was happening (async/await and Promises won't propagate through a map if I understood correctly) and their suggestion was to move the confusing parts out into an `asyncMap`, so that exists now.

oliviacpu:
💁 I think I like pulling the `asyncMap` out and am just cleaning up stuff that I broke related to imports and extractions.

src/presenters/projects-loader.js
oliviacpu: `asyncMap` doesn't exist anymore.

oliviacpu:
💁 Completing this as the ticket described avoided all of the promise weirdness and hopefully made things more clear. It also looks faster to me!

src/presenters/projects-loader.js
```diff
@@ -29,10 +27,28 @@ class ProjectsLoader extends React.Component {
...
+    // Gather unique user IDs for all of the projects being loaded, based on permissions
+    const uniqueUserIds = uniq(flatMap(projects, ({ permissions }) => permissions.map(({ userId }) => userId)));
+
+    // Load all of the users for this set of projects
+    const allUsers = await getFromApi(this.props.api, `v1/users/by/id?${joinIdsToQueryString(uniqueUserIds)}`);
+
+    // Go back over the projects and pick users out of the array by ID based on permissions
+    projects = projects.map((project) => ({
+      ...project,
+      users: flatMap(project.permissions, ({ userId }) => allUsers[userId]),
```
GregWeil: I think this can just use a normal map instead of flatMap

PR #298, opened 2019-03-08
Adding Annotations

sarahzinger:
This is still a draft, but I would appreciate some 👀 just to make sure I'm not doing anything too funky

TODO:
- [x] remove "Add note" button from the popover when there is already a note
- [x] add the right emoji to the add note button
- [x] 75 character limit
- [x] Remove projects-loader and load projects higher up in the foodchain
- [x] update component to be a css module
- [x] call everything a note instead of annotations since that's the word we're using in the ui
- [x] Ensure API call works, when it's pushed out on Monday
- [x] fix style bug where new notes don't have collectionCoverColor?
- [x] fix bug where we delete an annotation text and the annotation disappears before refresh
- [x] handle dark backgrounds to have light text
- [x] fix bug where you change the color of the collection and hit an error
- [x] figure out why state is not working in storybook
- [x] figure out what went wrong with my githistory lol
- [x] fix homepage css grid bug
- [x] arrow could probs be a bit bigger, and is not responsive
- [x] fix bug where note shows as authored by whoever is logged in rather than the actual author (🤦‍♀️ )
- [x] fix bug where non logged in users can edit note lol.
- [x] fix bug with content-editable on empty notes
- [x] fix bug related to rendering notes that are just spaces
- [x] test on mobile a bit (tested iphone safari and chrome on android, did find an existing android bug, but since I couldn't fix it quickly I decided to make a card for it [here](https://glitch.manuscript.com/f/cases/3328091/Collections-Titles-don-t-show-up-on-Chrome-on-Android))
- [x] tested voiceover on mac, was able to navigate to buttons and make a note
- [ ] test on different browsers (on mac latest versions of: firefox, safari, chrome, on windows: firefox..[.ie11 is currently borked](https://glitch.manuscript.com/f/cases/3328092/) so didn't get to test that yet) 

Things punted:
- [ ] css modules and storybook. I started down this path, but I believe it won't work properly in storybook until all the child components are css modules as well. We decided to hold off. 
- [ ] using the new api (it doesn't return the same stuff interestingly, would require a bigger refactor of what we have, decided we could do this in a new pr)

Manuscript: 
https://glitch.manuscript.com/f/cases/3327998/
https://glitch.manuscript.com/f/cases/3327999/

Apologies for this janky screencastgif I'm not sure why it's going so fast? lol if anyone has things they like better than gifbrewery or pro tips on how to use it would appreciate:
notes

View a collection live: https://iced-flag.glitch.me/@sarahzinger/my-great-collection


src/presenters/annotation.js
```diff
@@ -0,0 +1,44 @@
...
+
+const Annotation = ({
+  project, currentUser, update,
+}) => {
+  if (!project.isAddingANewAnnotation && !project.annotation) {
+    return null;
+  }
+
+  return (
+    <div className="annotation">
```
clottman: It would be neat if we could write these directly in CSS Modules, but I'm not sure if we're ready for that. @sheridanvk do you have a preference?

src/presenters/projects-loader.js
```diff
@@ -56,9 +56,12 @@ class ProjectsLoader extends React.Component {
...
     const { children, projects } = this.props;
-    const loadedProjects = projects.map(
-      project => this.state[project.id] || project,
-    );
+    // Feedback needed!
+    // I'm tempted to get rid of this "projects-loader" component altogether and load the projects wherever we define them in state.
+    // Otherwise we get into this weird state where we have in-progress updates (such as an annotation on a project in a collection) in props
+    // that are not getting properly passed down unless we merge like we're doing here. It seems funky to me to have props.projects of our parent element
+    // be totally different from the props.projects of the children elements here.
+    const loadedProjects = projects.map(project => ({ ...project, ...this.state[project.id] }));
```
clottman: That does seem weird and changing it as you suggested sounds pretty reasonable!

keithk: I agree with the conclusion that you came up with here!

src/presenters/annotation.js
sheridanvk: yeah that would be ideal at this point (all new components as CSS Modules). Looks like this one could be self-contained enough that this wouldn't be too problematic to do, so Sarah if you're up for it, then it would be great to add that in here. Happy to swing by and chat about what's involved if you'd like, and/or you can use this PR as a basis for building this: https://github.com/FogCreek/Glitch-Community/pull/244. I'll take a task to write up a "Adding a new component" guide as well to make this super clear.

sheridanvk:
One thing as I was reading through this - should we try to standardise on either calling these Annotations or Notes in the code? I guess it doesn't have to match what the user sees in the UI (although that would be ideal - and then the only inconsistency is with what the API calls it), but I found it a bit confusing that the component is called Annotation, but then elsewhere you have functions named things like addNoteField.

ttseng:
This is looking great so far!  

A few comments about the UI:

* The background color of the annotation should match the collection `coverColor` 
* Looks like we're not handling word wrap well for editable description fields more generally - if we add `word-break: break-all` then it wraps appropriately:

image

* To make the annotation feel more connected to the project, I would set a `margin-bottom: -30px` to the annotation div so it overlaps with the project card.
image
* at first, I was imagining that if a user clicks outside of an empty annotation, the annotation is removed.  but this doesn't account for if the page loses focus (for example, with switching tabs) so I think we'll need to add an explicit option to remove an added annotation.  I'll work on this and update the spec (and let you know when I do)

ttseng:
re sheridan's comment about naming - I agree it should be consistent. I talked to emanuele about it, and it sounds like they can rename `annotation` to `note` in the API.  For now, let's stick with referring to the new component as `note` on the community site.

ttseng:
also - you can check out the example of using `isDark`(in `/presenters/pages/collection.js`) to set the annotation text color to black or white depending on the collection coverColor:

image

src/presenters/includes/link.js
```diff
@@ -8,8 +8,7 @@ import { getLink as getProjectLink } from '../../models/project';
 import { getLink as getTeamLink } from '../../models/team';
 import { getLink as getUserLink } from '../../models/user';
 
-/* global EXTERNAL_ROUTES */
-const external = Array.from(EXTERNAL_ROUTES);
+const external = global.EXTERNAL_ROUTES ? Array.from(global.EXTERNAL_ROUTES) : [];
```
sarahzinger: this was causing some issues when I was running storybook (back before we decided not to make note a component for now) we don't really need it now, but it seemed chill to leave in, and might save a life someday?

CONTRIBUTING.md
```diff
@@ -50,14 +50,14 @@ _(Do this once per computer to create a local repository)_
 _Remember to [set up Git SSH](https://help.github.com/articles/connecting-to-github-with-ssh/) or the [Windows Client](https://desktop.github.com/) to get your auth in order._
 
   ```
-  # 1. On your local machine, clone our git repository from Github: 
+  # 1. On your local machine, clone our git repository from Github:
```
sarahzinger: I didn't actually make any edits here, it's just whitespace changes from atom from when I resolved some weird merge conflicts

src/presenters/pages/collection.js
```diff
@@ -74,130 +74,184 @@ const CollectionPageContents = ({
...
-          {collection.team && <TeamTile team={collection.team} />}
-          {collection.user && <UserTile user={collection.user} />}
-
-          <div className="collection-description">
-            <AuthDescription
-              authorized={isAuthorized}
-              description={collection.description}
-              update={updateDescription}
-              placeholder="Tell us about your collection"
+}) => {
```
sarahzinger: this git diff is hard to read so I'll highlight some changes I made here:
-  I find nested ternaries kind of hard to read (particularly with jsx) so I removed them. Now the structure is always if this -> render this. I also tried to make it a bit more readable with 2 more named variables. The upside is that it's hopefully a bit easier to read. The downside is that if we have freaky logic we might render more than one case. I guess another alternative could be a switch, or abstracting out the code into their own components or something like that. Hopefully this is at least a step in the right direction but if not feel free to speak your truth on that.
- I removed the projectsLoader from this file. I found it very strange that the props.projects of this component were totes different from the props.projects of it's children. Hopefully this is more clear. 
- I found isAuthorized a bit vague in this context (auth'd do what specifically?) and confused it with isAuthenticated at first pass (which... you know my bad really) but I thought the parent var name of userIsAuthor is super clear, so I changed that up, hopefully helps someone else (especially if we ever have multiple roles and owners of collections, this person isAuth'd to delete a note, this person is auth'd to edit a note whatever).

stories/index.js
```diff
@@ -61,6 +62,12 @@ storiesOf('Text Input', module)
   .add('with error', () => <TextInput placeholder="glitch" error="That team already exists" />)
   .add('text area', () => <TextArea placeholder="[Something here] doesn't seem appropriate for Glitch because..." error="Reason is required" />);
 
+storiesOf('Heading', module)
```
sarahzinger: so confused as to why this is showing up in my gitdiff... I tried to reset this file by copy- 🍝 -ing index.js from master here, and this showed up? lol. whyyy

shrinkwrap.yaml
```diff
@@ -1,27 +1,27 @@
 dependencies:
-  '@babel/core': 7.3.4
-  '@babel/plugin-proposal-class-properties': 7.3.4
+  '@babel/core': 7.2.2
```
sarahzinger: I do not understand why this happened? On master we're using 7.2.2 ?? https://raw.githubusercontent.com/FogCreek/Glitch-Community/master/shrinkwrap.yaml

sarahzinger:
@ttseng @sheridanvk I think this should be ready for a more formal review if you want to lend me your thoughts!

stories/index.js
ttseng: in the Glitch console, try `git checkout origin/master stories/index.js`?  same for shrinkwrap.yaml

src/presenters/note.js
```diff
@@ -0,0 +1,61 @@
...
+ */
+const Note = ({
+  collectionCoverColor, author, project, update,
+}) => {
+  if (!project.isAddingANewNote && !project.note) {
+    return null;
+  }
+
+  const className = classNames({
+    descriptionContainer: true,
```
ttseng: we usually hyphenate classnames so this should be `description-container` rather than `descriptionContainer`

ttseng:
I edited the css on the note item so it behaves better on resize:
```
.descriptionContainer
    width: calc(100% - 36px)
    &:after
        right: 26px
```
(made the change directly in the remix 🙂)
note-responsive

ttseng:
excited to go through this fully on Monday – thanks Sarah!

src/presenters/pages/collection.js
ttseng: to be even more explicit, we can rename `userIsAuthor` it `currentUserIsAuthor` similar to for `teams.js` we have the property `currentUserIsOnTeam` to determine whether or not they're authorized.

sarahzinger: great thought! will update :)

ttseng:
Some UI bugs I'm currently working on fixing:

- [ ] bottom of note changes based on the size of the description
positioning-bug

- [ ] shifting of description height based on whether user is editing / viewing
shifting

- [ ] handling of differently-sized notes in the same row
image


- [ ] update clearing of note UX - I'd like to modify it so that if the note goes from having a non-empty description to an empty one, we clear the note after the user toggles outside the note (by clicking outside of it)

ttseng:
Added UI so that note disappears if it becomes empty after editing.

disappearing-note

I think this is looking good from a UI perspective.  @sheridanvk @keithk would either of you be up for doing a more functional code review?

oliviacpu:
Screenshot 2019-03-19 at 10 15 09 AM

Some fun Markdown.

- Excessive returns won't get rendered in the Markdown view, but will expand the editable field beyond the background.
- The image is possible with a URL shortener and I think it is fun but I could see not wanting to allow certain tags in certain instances. [`markdown-it` has an option for this.](https://github.com/markdown-it/markdown-it#manage-rules)

oliviacpu:
I have this checked out locally to review and also look at this Markdown/editable field thing.

src/components/tooltips/tooltip.styl
```diff
@@ -94,6 +94,3 @@ tooltip-background = rgba(0,0,0,0.95)
   right: 32px
   top: 110%
 
-.tooltip.new-stuff::before
-  border-top-color: notification
-  border-bottom-color: transparent
```
sheridanvk: just checking, was this intentionally removed? doesn't seem related to the PR but maybe I'm missing it.

oliviacpu:
Screenshot 2019-03-19 at 3 40 58 PM

Setting `maxRows={4}` on the text area made it behave more like I would expect. I can still enter extra returns but the area is scrollable and doesn't overflow behind the project description. I don't know if where I set it is necessarily the right place to do that, but I just wanted to test it for now.

Exposing an `allowImages={false}` option on the `Markdown` component would also let us prevent the image thing I did.

sarahzinger:
@oliviacpu Great catch! would have never thought about images! I'll go on ahead and disable. Also yeah I'm def running into some weirdnesses with returns/spaces as notes. I'll try again with that.

@sheridanvk I did not! lol I don't understand how I keep getting stuff into my branch without intending too lol. I guess this is a lesson in small prs lol. Thanks for catching it :)

src/presenters/includes/link.js
sheridanvk: just curious, did this solve anything for you? I'm running into the same problem now because the Button now has a link, but adding this line did not help me.

sarahzinger: it did! I also had to install a decorator for it

sarahzinger: let me know if you want to pair on it :) my styles were super janky but I was able to get link components to load and stuff

ttseng:
re: markdown - we'll want to add a smart way for link markdown not to count towards your character limit.  but that's not a requirement for this PR - just wanted to make a note!

src/presenters/includes/link.js
sheridanvk: ohhhh wait I was definitely putting this in the wrong place :D got it to work now, thank you!

sheridanvk: although shouldn't this be window., instead of global., as we're in client-side code, not Node? 
https://javascript.info/global-object

sarahzinger: hmm potentially :) I'll remove from this pr and let you sort it out!

oliviacpu:
🎉 Tried it out on the remix and it looks good!

PR #299, opened 2019-03-08
Violet mailman: `prettier` all the things

modernserf:
- add prettier config (from chrome extension) to project
- add script to run prettier with this config
- update lint rules to work with prettier
- run prettier on everything

This also includes some manual changes of pathological cases around "double indentation" in ternaries:

```js
const onError = srcFallback
  ? (event) => {
      event.target.src = srcFallback;
    }
  : null
```

eslint's indent rule cannot be convinced that this is indented correctly. The prettier docs suggest disabling the indent rule (and deferring all indentation to prettier) but there were few enough of these cases that I just replaced them with `if` statements.


sh/prettify.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+# Runs `prettier` on our code
+npx prettier --config /app/.prettierrc --write "/app/{server,src}/**/*.js"
```
clottman: Does this have to be called manually?

src/curated/categories.js
```diff
@@ -6,63 +6,52 @@ export default [
     id: 4,
```
clottman: Files in `curated/` should probably be excluded from prettier - they're ignored by our eslint config, because they often get changed in the editor instead of through our pull request process.

sh/prettify.sh
modernserf: Where would a good place be for us to automate this? should this be called by the other scripts?

modernserf:
@clottman updated with .prettierignore, explicit lint rules

package.json
```diff
@@ -73,7 +74,9 @@
     "url-join": "^4.0.0",
     "webpack": "^4.29.3",
     "webpack-cli": "^3.2.2",
-    "webpack-dev-middleware": "^3.5.2"
+    "webpack-dev-middleware": "^3.5.2",
+    "eslint-plugin-prettier": "^3.0.1",
+    "eslint-config-prettier": "^4.1.0"
```
clottman: Looks like we're not extending this config any more, right? Also listed twice (line 38 and 79)

PR #302, opened 2019-03-11
Peridot soul: fix accidental event global

modernserf:
This block of code was using [window.event](https://developer.mozilla.org/en-US/docs/Web/API/Window/event), which is _usually_ what you want except when it isn't.

I found this while chasing down [this bug](https://sentry.io/organizations/fog-creek-software/issues/928442716) but it appears to be unrelated.


modernserf:
its ... not impossible that this could be related to the too much recursion on file upload bug?

GregWeil:
We only use progress promises in the s3 uploader. If you're feeling bold, removing them entirely would probably make this easier to reason about.

In the mean time this change looks sound on its own, the one place we use the notify function does give it an event object.

clottman:
I think this is good to merge/deploy. And perhaps removing the progress promise can be a future ticket/part of the infinite recursion investigation.

modernserf:
Updated this PR to:
- re-enable linting on this file (!)
- enable a lint rule that will catch usage of the `event` global

src/utils/progress-promise.js
```diff
@@ -21,10 +19,10 @@ if (Promise.prototype.finally == null) {
 }
 
 if (Promise.prototype._notify == null) {
-  Promise.prototype._notify = function(event) {
+  Promise.prototype._notify = function _notify(event) {
     return this._progressHandlers.forEach((handler) => {
       try {
-        return handler(event);
```
modernserf: for consistent return values; forEach doesn't use return values anyways

src/utils/progress-promise.js
```diff
@@ -1,11 +1,9 @@
-/* eslint-disable */
-
-export default ProgressPromise;
+/* eslint-disable no-extend-native, no-underscore-dangle */
 
 // Extend promises with `finally`
 // From: https://github.com/domenic/promises-unwrapping/issues/18
 if (Promise.prototype.finally == null) {
-  Promise.prototype.finally = function(callback) {
+  Promise.prototype.finally = function _finally(callback) {
```
modernserf: for func-names

src/utils/progress-promise.js
```diff
@@ -44,21 +42,23 @@ if (Promise.prototype.progress == null) {
...
-        ? p._progressHandlers.forEach((handler) => {
-            try {
-              return handler(event);
-            } catch (error) {
-              // empty
-            }
-          })
-        : undefined;
+  const p = new Promise((resolve, reject) => {
+    const notify = (event) => {
```
modernserf: add proper `event` variable

src/utils/progress-promise.js
```diff
@@ -44,21 +42,23 @@ if (Promise.prototype.progress == null) {
...
+      if (p._progressHandlers == null) {
```
modernserf: fix indentation (pathological case with prettier vs eslint rules)

src/utils/progress-promise.js
```diff
@@ -21,10 +19,10 @@ if (Promise.prototype.finally == null) {
 }
 
 if (Promise.prototype._notify == null) {
-  Promise.prototype._notify = function(event) {
+  Promise.prototype._notify = function _notify(event) {
```
sarahzinger: for my own brain, does this change do something or is it a stylistic thing?

modernserf: it potentially changes what stack traces look like, but its mostly just to please eslint

sarahzinger:
hmm seems legit to me! not sure why the runner has failed

PR #303, opened 2019-03-11
nice-bolt: Returns routing to /team for curated teams

keithk:
Prettier updates + making /team-name for teams in `curated/teams.js` work correctly

Example: https://nice-bolt.glitch.me/twilio


src/presenters/pages/router.js
```diff
@@ -140,130 +108,60 @@ const Router = ({ api }) => (
...
-        exact
-        render={({ location, match }) => (
-          <TeamOrUserPage
-            key={location.key}
-            api={api}
-            name={match.params.name}
-          />
-        )}
-      />
+      <Route path="/(@?):name" exact render={({ location, match }) => <TeamOrUserPage key={location.key} api={api} name={match.params.name} />} />
```
keithk: This is the only change -- wrapping the @ in a (?) which makes it optional even in an `exact` route

clottman:
Cross-posted from Slack: 

What's the expected behavior for when a team shares the same name as a collection?
I just created the team @learn-to-code, but https://glitch.com/learn-to-code is a collection.

keithk:
Just a note that we're talking about the collections-that-aren't-collections on the bottom of the home page (like More Ideas) - https://glitch.com/hello-worlds

Also the build failure was from a lack of space left on the device? D:

keithk:
Had a misunderstanding about this regression -- it's that teams in `curated/teams.js` are not correctly displaying, these are teams that should have a root URL, not *every* team.

Will be moving this to in progress to resolve.

src/presenters/pages/router.js
keithk: Reverted this - only curated teams (`curated/teams.js`) should have the root URL, not all teams.

src/presenters/pages/team-or-user.js
```diff
@@ -86,7 +69,7 @@ const getTeam = async (api, name) => {
 };
 
 const TeamPageLoader = ({ api, id, name, ...props }) => (
-  <DataLoader get={() => getTeamById(api, id)}>
+  <DataLoader get={() => getTeam(api, name)}>
```
keithk: We pass in the name, so now we use `getTeam`, which uses the new /v1 API above instead of the getTeamById

keithk:
This was due to the refactoring of the team pages to use the new API, the block in particular was `parseTeam` which filtered through params that did not exist in the old API calls.

Now using the `getTeam` to get the team, which returns from the v1 API.

src/presenters/pages/team-or-user.js
GregWeil: I'm not certain that all of the teams use the same curated url as their team url

keithk: I'll take a look 👀

keithk: If not then I'll just rewrite the getTeamByID to match the queries in getTeam

keithk: @GregWeil I just ran through the list and couldn't find any examples where there was a difference between the team presented at /name or /@name from the curated teams

PR #304, opened 2019-03-11
Quasar chicory: fix tag nesting in new project popup

modernserf:
The new project popup is _visually similar_ to the project lists in the Add Team/Add Collection popups, but it doesn't function the same, and using ProjectResultItem is [the wrong abstraction](https://www.sandimetz.com/blog/2016/1/20/the-wrong-abstraction).


src/presenters/pop-overs/new-project-pop.js
```diff
@@ -3,30 +3,40 @@ import PropTypes from 'prop-types';
...
-  );
+  const repoUrl = window.prompt('Paste the full URL of your repository', 'https://github.com/orgname/reponame.git');
   /* eslint-enable no-alert */
   if (!repoUrl) {
     return;
   }
   window.location.href = `/edit/#!/import/git?url=${repoUrl}`;
 };
 
+
```
modernserf: This was based on ProjectResultItem, but most of the complex stuff was either extracted into ProjectAvatar, or removed because its not used here -- neither a button nor a link are needed here, since the TrackedExternalLink handles that behavior.

src/presenters/pop-overs/new-project-pop.js
```diff
@@ -50,11 +55,7 @@ const NewProjectPop = ({ projects }) => (
       </div>
     </section>
     <section className="pop-over-actions last-section pop-over-info">
-      <button
```
modernserf: The remainder of this file is auto-formatting.

src/components/avatars/project-avatar.js
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
```
modernserf: The avatar component is (slightly) complex and self-contained, so it has been extracted into its own component and shared between the two project list item components.

src/presenters/pop-overs/new-project-pop.js
```diff
@@ -18,6 +18,19 @@ const importGitRepo = () => {
   window.location.href = `/edit/#!/import/git?url=${repoUrl}`;
 };
 
+
+const NewProjectResultItem = ({ id, domain, description }) => (
+  <div className="result result-project">
+    <ProjectAvatar domain={domain} id={id} cdnUrl="https://cdn.glitch.com" />
```
sheridanvk: does this cdnUrl ever change? Should we just have it be hardcoded into ProjectAvatar?

modernserf: even better -- the CDN url has a default argument value in `getAvatarUrl`. We can leave the prop here for completeness but make it optional, and remove it from the new project menu

sheridanvk: sounds great!

src/presenters/includes/project-avatar.js
```diff
@@ -0,0 +1,27 @@
...
+  />
+);
+
+ProjectAvatar.propTypes = {
+  domain: PropTypes.string.isRequired,
+  id: PropTypes.string.isRequired,
+  cdnUrl: PropTypes.string,
+};
+
+ProjectAvatar.defaultProps = {
```
modernserf: This seems somewhat superfluous, but the linter demands that you make this explicit

sheridanvk: why wouldn't the default for this be https://cdn.glitch.com?

modernserf: the defaulting is handled by `getAvatarUrl`

PR #305, opened 2019-03-11
Industrious-dolomite - replace 'null' in abuse reports of anon users with 'anonymous user' + id

jennschiffer:
closing it because i borked it up


jennschiffer:
i ran prettify on this and it made some formatting changes. could use a check on if they are the right changes!

PR #307, opened 2019-03-12
Superb guide: Componentized team name and url fields

GregWeil:

I like the more explicit debounce (vs having it hidden away in the constructor). The layout is pretty different from normal js, but I do feel a bit more confident reasoning about it.

I also moved everything that was in the fields folder into an inputs folder. Stuff in the inputs folder act like standard html input elements re value/onChange, vs stuff in the fields folder which expect async onChange handlers and handle displaying errors.

I'm not sure how viable these fields will be outside of the specific pages they're used. Do you think it'd be a good idea to move the TeamEditor api handling into the individual fields, so the fields would deal with making the PATCH request. Rather than taking an async onChange, they'd just only call onChange after successful requests, when the page would update state and set the url.


sheridanvk:
Your question about moving the api calls into the places where they're used makes sense to me - tagging @modernserf though, as I know he is thinking about our general approach to where we access the API and where to store data for easier access throughout the app.

src/components/fields/use-optimistic-text.js
```diff
@@ -0,0 +1,7 @@
+import useOptimisticValue from './use-optimistic-value';
+
+const useOptimisticText = (realValue, setRealValueAsync) => (
+  useOptimisticValue(realValue, (newValue) => setRealValueAsync(newValue.trim()))
+);
+
+export default useOptimisticText;
```
sheridanvk: We only use this component in optimistic-text-input, right? I'm wondering whether in those cases, we should just define these in the component that uses them. Seems like we have a lot more components than I expected in the fields folder, so while we shouldn't be exporting more than one component from them, perhaps we can define multiple components in there?

sheridanvk: or if these are hooks we think we'll use elsewhere, maybe we need a separate place to store them? They aren't components, so it's weird to me to have them here like this - I would only expect optimistic-text-input, team-name-input and team-url-input to be in this folder

src/presenters/pages/team.js
```diff
@@ -36,15 +37,10 @@ function syncPageToUrl(team) {
 const TeamNameUrlFields = ({ team, updateName, updateUrl }) => (
   <>
     <h1>
-      <EditableField value={team.name} update={updateName} placeholder="What's its name?" suffix={team.isVerified ? <VerifiedBadge /> : null} />
+      <TeamNameInput name={team.name} onChange={updateName} postfix={team.isVerified ? <VerifiedBadge /> : null} />
```
sheridanvk: Are there places where we would show the TeamNameInput without the verified badge (assuming the team is verified? If not I think we shouldn't be passing it in as a prop; it should just be part of the TeamNameInput component. Also would moving the api calls into the component mean we could remove the onChange prop here? If so I would be highly in favour :D

src/components/fields/use-optimistic-text.js
GregWeil: They'd be used for textarea/markdown fields and maybe the privacy toggle. I do agree this folder is getting pretty big. If we do move api calls and stuff into the fields then they would probably belong in a separate folder from fields.

useUniqueId and useDebouncedValue are both pretty generic. Maybe src/hooks?

src/presenters/pages/team.js
GregWeil: I'm not sure. I think we want to keep all of the team in a single state (for example, on the user page we use their name to sync up the project users with them)

It would mean the onChange could be replaced with an inline `name => setState({name})`

GregWeil: I don't think we have any other live edit team names.

Mind if I just import the current verified badge until the image component is ready?

GregWeil:
Updated with built in verified badge

GregWeil:
Are there any remaining concerns with this pr? I don't think it's worth refactoring the PATCH request with Justin's coming redux changes in mind.

src/components/fields/use-optimistic-text.js
sheridanvk: yeah I think that would be great!

src/presenters/pages/team.js
```diff
@@ -36,15 +37,10 @@ function syncPageToUrl(team) {
...
+      <TeamNameInput name={team.name} onChange={updateName} verified={team.isVerified} />
     </h1>
     <p className="team-url">
-      <EditableField
-        value={team.url}
-        update={(url) => updateUrl(url).then(() => syncPageToUrl({ ...team, url }))}
-        placeholder="Short url?"
-        prefix="@"
-      />
+      <TeamUrlInput url={team.url} onChange={(url) => updateUrl(url).then(() => syncPageToUrl({ ...team, url }))} />
```
sheridanvk: this is great! much cleaner than before

GregWeil:
Ok, will move to src/hooks while I fix merge conflicts

GregWeil:
Fixed merge conflicts, moved generic hooks to src/hooks

Will probably merge this later today

PR #308, opened 2019-03-12
Text / Markdown

oliviacpu:
Removing the `Markdown` and `TruncatedMarkdown` components and using the new `Markdown` component made in our new pattern.

- [x] Created new Markdown component
- [x] Removed old components
- [x] Added to Storybook

Areas to check:
- "Collection Item"
- "Featured Collections"
- "Description Field"
- "New Stuff"
- Project Pages
- Project Item / Team Item / User Item - Search Results?

Quick links:
- [Olivia's profile should have an image](https://joyous-wallaby.glitch.me/@oliviacpu)
- [Jenn's profile should have a link](https://joyous-wallaby.glitch.me/@jennschiffer)
- [Olivia's profile should not have an image in search](https://joyous-wallaby.glitch.me/search?q=oliviacpu)
- [Jenn's profile should have a link in search and also be truncated](https://joyous-wallaby.glitch.me/search?q=jennschiffer)


src/components/text/markdown.js
```diff
@@ -0,0 +1,52 @@
...
+  const className = cx({
+    'markdown-content': true,
+  });
+  let rendered = md.render(children || '');
+  if (length > 0) {
+    rendered = truncate(rendered, length, { ellipsis: '…' });
+  }
+  return (
+    <span
+      className={className}
```
sheridanvk: For this one, because it's so simple, you don't need to build a className variable - you can just delete it above and set className="markdown-content" here. You only need to use the classnames library if there's some complicated logic around what classnames you include in certain cases.

oliviacpu:
Not okay with this checkpoint emoji combo!

src/components/text/markdown.js
oliviacpu: Just resolved! Thank you!

oliviacpu:
🙅‍♂️ The font looks like it is not rendering properly.

oliviacpu:
💁 There are differences in the font between logged in and logged out profile pages. I think the behavior currently matches production.

sheridanvk:
> Not okay with this checkpoint emoji combo! 👼🔫 Checkpoint 

jeez! @jennschiffer FYI as I think you've been making changes here

src/components/text/markdown.js
```diff
@@ -0,0 +1,46 @@
...
+ * Markdown Component
+ */
+const Markdown = ({ children, length }) => {
+  let rendered = md.render(children || '');
+  if (length > 0) {
+    rendered = truncate(rendered, length, { ellipsis: '…' });
+  }
+  return (
+    <span
+      className={styles['markdown-content']}
```
sheridanvk: Just realised too we've been moving to the style of camel casing style attributes in CSS Modules, as that's easier to refer to in code (and enables e.g. styles.markdownContent). I'll add this information to the components guide so we can stay consistent there; sorry for the omission.

sheridanvk:
> 💁 There are differences in the font between logged in and logged out profile pages. I think the behavior currently matches production.

oh yeah I think this is by design to show it's editable. Agree that this should be left as-is!

PR #309, opened 2019-03-13
Fallback for tooltips within buttons on <= Firefox 66

taravancil:
On <= FF 66, mouse events on children of `<button>`s are swallowed, so `mouseEnter` and `mouseLeave` never fired for team member avatars in FF. Open [@testy-team](https://glitch.com/@testy-team) in FF for an example.

Case [3327992](https://glitch.manuscript.com/f/cases/3327992/Tooltips-don-t-show-up-for-team-members-on-firefox)

## Changes

This adds the `fallback` prop to the `TooltipContainer` component, which falls back to CSS tooltips when the tooltip target is the child of a `<button>`. Right now, this only happens on user avatars on team pages.

Open [like-pleasure.glitch.me/@testy-team](https://like-pleasure.glitch.me/@testy-team) in FF for a working example.

I've tested these changes in Chrome, Safari, and Firefox to check for regressions.


src/presenters/includes/team-users.js
```diff
@@ -29,8 +29,14 @@ export const TeamUsers = (props) => (
...
         <li key={user.id}>
           <PopoverWithButton
-            buttonClass="user button-unstyled"
-            buttonText={<UserAvatar user={user} suffix={adminStatusDisplay(props.adminIds, user)} />}
+            buttonClass="user button-unstyled tooltip-container-button"
+            buttonText={(
+              <UserAvatar
+                user={user}
+                suffix={adminStatusDisplay(props.adminIds, user)}
+                withinButton
```
oliviacpu: It's a little unclear to me when I should set `withinButton` or use a `fallback`.  I'm wondering if it could be set automatically at a lower level, maybe based on `hideTooltip` or something?

taravancil: `withinButton` is a prop on the `UserAvatar` component that indicates whether or not the `fallback` prop should be set on the `TooltipContainer` component. Since `UserAvatar` is not always within a `<button>`, we need `withinButton` to make sure we only use the fallback tooltips when we really need them.

Does that make sense? idk maybe not, but it's not immediately clear to me how we could use `hideTooltip` to make this simpler (also my brain is completely burnt out on tooltips lol)

styles/tooltips.styl
```diff
@@ -0,0 +1,97 @@
+tooltip-background = rgba(0,0,0,0.95)
+
+.touch [data-tooltip]::after
+  display: none
+
+.tooltip-container
+  position: relative
+  box-sizing: border-box
+  display: inline-block
```
GregWeil: Are these first two selectors missing `.fallback`?

taravancil: I suppose I could add them for clarity, but because we're using CSS modules (or whatever it's called), the non-fallback `.tooltip-container` class actually looks like `.tooltip__container{randomString}`

GregWeil: ah, that makes sense.

PR #310, opened 2019-03-13
Updated Search Result Cards & Filtering

ttseng:
Remix at https://pentagonal-streetcar.glitch.me/search?q=react

**Summary of changes:**
* Updated the search result page for smaller cards + 4 column layout on larger displays.
* Added filter buttons

filtering

Most changes are in `search.js`

Looking for feedback on:
* better handling of filter types - there's a comment at line 167 in `search.js` about the stuff I think can be improved.
* when all search results are shown, I show a total count of search results, mostly to distinguish between the "all results" and "team results filters".  that number changes as different API calls take different amounts of time to load.  Is there something else I can do here to make the count look less jumpy? or, if showing the total # of results is not so useful, I could put some other placeholder there altogether.
total-count


ttseng:
Added segmented buttons with badge styling for hit count
segmented-buttons-badges

Also fixed bug with handling no results.

oliviacpu:
> * when all search results are shown, I show a total count of search results, mostly to distinguish between the "all results" and "team results filters".  that number changes as different API calls take different amounts of time to load.  Is there something else I can do here to make the count look less jumpy? or, if showing the total # of results is not so useful, I could put some other placeholder there altogether.

Maybe just "All results for <query>" instead of the count with "8 results for <query> in Teams" when you change tabs?

ttseng:
> Maybe just "All results for " instead of the count with "8 results for in Teams" when you change tabs?

@oliviacpu hm yeah, that could work too.  here's what they look like compared to one another:


All results for react:
* note - this message can be displayed even as the results are loading
all-results-filters-loader


14 results for hello: 
* note - the message is only shown after all results are loaded and the count can be accurate
loading-with-count


My instinct is that If you see the All Results message while it's still loading, that's probably more helpful than seeing nothing until all the results are complete.  What do you think?

pketh:
Looks great, just a couple things:

# segmented buttons button styles
1. instead of having to explicitly pass in classes to buttons inside a .segmented-buttons you can just inherit them:

eg from the editor `buttons.styl`

```
button-small()
  border-radius: 5px
  font-weight: 600
  font-size: 12px
  height: 25px
  padding-left: 8px
  padding-right: 8px
  .loader
    vertical-align: -3px
  .emoji
    width: 16px
    height: 16px
    margin-top: 0

.segmented-buttons
  button
    button-small()
    &:first-child
      border-top-right-radius: 0
      border-bottom-right-radius: 0
    &:last-child
      border-left: 0
      border-top-left-radius: 0
      border-bottom-left-radius: 0
    &:not(:first-child):not(:last-child)
      border-radius: 0px
      border-left: 0

you could clean this code up a bit with the use of `button + button` instead of first/last-child 
```

2. the buttons in the segmented buttons are detached from each other. they should be connected
Screen Shot 2019-03-15 at 10 27 31 AM

3. the status badge color has incorrect metrics and background color.  
```
.status-badge
  font-weight: 600
  font-size: 12px
  line-height: 14px
  vertical-align: middle
  color: background-color
  .status
    padding-top: 2px
    padding-bottom: 0px
    border-radius: 5px
    padding-left: 5px
    padding-right: 5px
    display: inline-block
    margin-left: 5px
    background-color: line-on-secondary-background
```

4. the active style on a secondary button should not change the border width. (relatedly, clicking any of the buttons should not shift the position of the other buttons)
eg
Screen Shot 2019-03-15 at 10 36 53 AM

5. should the 'all' button display the total count in a badge too?

# search box
6. appears to have this double box thing going on
Screen Shot 2019-03-15 at 10 39 16 AM

# h1
7. "all results for X" seems unnecessary when 'all' is selected. The ui already implies this

# entity items

8. because most cover art doesn't change, prioritizing the cover art by putting it first/top doesn't seem to help differentiate results from each other very well. Because it's usually the least differentiated item, maybe it should go below the `.content` which are more commonly changed.
Screen Shot 2019-03-15 at 10 45 07 AM

9. when i hover over an entity, my mouse cursor changes but the button does not. when i hover over the button, the button changes. It seems like this makes the link between the button and the entity less strong/ 1:1 than it should be.  You could address this w a css override like 
```
.item
   &:hover
      button
        hoverstyles()
   &:active 
      etc
```


(tested in safari)

pketh:
10. the spacing between the segmented-buttons and the section below it is inconsistent for diff filters



👀 the space between `users` and `projects`

ttseng:
thanks @pketh !  looks like the styling issues might be safari-related (it looks different / fine to me on Chrome).  will clean it up.

package.json
```diff
@@ -91,5 +89,9 @@
     "./src/utils/progress-promise.js",
     "./src/polyfills.js"
   ],
-  "license": "Apache License Version 2.0"
+  "license": "Apache License Version 2.0",
+  "devDependencies": {
+    "eslint": "^5.15.1",
+    "eslint-plugin-react": "^7.12.4"
+  }
```
GregWeil: What's the reasoning for moving eslint here? (Same for adding the new config file)

oliviacpu:
> My instinct is that If you see the All Results message while it's still loading, that's probably more helpful than seeing nothing until all the results are complete. What do you think?

I agree. I feel like it gives a context to go back to maybe.

src/presenters/pages/search.js
```diff
@@ -86,42 +135,65 @@ class SearchResults extends React.Component {
...
   async addProjectToCollection(project, collection) {
     await this.props.api.patch(`collections/${collection.id}/add/${project.id}`);
   }
 
   render() {
-    const { teams, users, projects } = this.state;
+    const { teams, users, projects, activeFilter } = this.state;
     const noResults = [teams, users, projects].every((results) => !showResults(results));
+    // I'm sure there's a better way to do this
+    const showTeams = (activeFilter === 'all' || activeFilter === 'teams') && showResults(teams);
```
modernserf: you could do
```
const showTeams = ['all', teams'].includes(activeFilter) && showResults(teams);
```
but that's not really a huge improvement

src/presenters/pages/search.js
```diff
@@ -86,42 +135,65 @@ class SearchResults extends React.Component {
...
+    const showUsers = (activeFilter === 'all' || activeFilter === 'users') && showResults(users);
+    const showProjects = (activeFilter === 'all' || activeFilter === 'projects') && showResults(projects);
+
+    // store results per type
+    filters[1].hits = teams ? teams.length : 0;
+    filters[2].hits = users ? users.length : 0;
+    filters[3].hits = projects ? projects.length : 0;
```
GregWeil: Rather than referencing a global array shared with the FilterContainer, it could create the array here and pass it in as a prop.

src/presenters/team-item.js
```diff
@@ -14,7 +14,8 @@ export default function TeamItem({ team }) {
   const style = getProfileStyle({ ...team, size: 'medium' });
   const teamThanksCount = team.users.reduce((total, { thanksCount }) => total + thanksCount, 0);
   return (
-    <WrappingLink href={getLink(team)} className="item button-area" style={style}>
+    <WrappingLink href={getLink(team)} className="item button-area">
+      <div className="cover" style={style} />
       <div className="content">
```
GregWeil: WrappingLink is throwing a propTypes error, because react router's withRouter() doesn't like more than one child element.

ttseng:
@pketh re: 
3. the status badge color has incorrect metrics and background color.

can you point me to what the expected styling is?  I couldn't find any examples of status badges that weren't colored using `success. `warning`, or `error` styling (on the site or in the editor)

pketh:
to get the default styling , inspect the badge on a button (like tools -> container stats) and remove the `success` class from it

src/presenters/pages/search.js
ttseng: thanks! that does make it more concise.

src/presenters/pages/search.js
ttseng: thanks! done.

src/presenters/team-item.js
ttseng: thanks for catching this.

ttseng:
changes:
* fixed spacing between segmented buttons in safari
* fixed input type=search styling on safari
* added active & focus styling to segmented buttons
* updated team-item, project-item, and user-item to use wrappingLinks, fixed keyboard accessibility

responses to design feedback:
3. Looks like the editor's default for status-badge background color is `#B8B8B8`, which corresponds to `line-on-secondary-background` in global.jsx, which is what I'm using.
5. I didn't include the count # in all results since it's redundant (the sum of all the other badges), and I wanted the badges to mostly serve as a way to easily check the relative # of results for each type
7. I kept the h1 because 1) it gives you a sense of what search results you expect to see once all the loaders finish (before you see anything else on the page), and 2) it helps distinguish between showing all results and team results (w/o it, it seems like nothing happens when you click team results)
8. I kept the cover at the top to be consistent with how cover image usually implies that it appears above everything else.

ttseng:
will figure out a way to grab the latest shrinkwrap.yaml from master.  otherwise, this is ready for a review.  requesting re-review from @GregWeil but let me know if you'd prefer another reviewer.

src/presenters/pages/search.js
```diff
@@ -14,7 +19,51 @@ import ProjectsList from '../projects-list';
...
+    <>
+      <div className="search-filters segmented-buttons">
+        {filters.map(
+          (filter) =>
+            (filter.hits === null || filter.hits > 0) && (
+              <Button
+                key={filter.name}
+                size="small"
+                type="tertiary"
+                active={activeFilter === filter.name.toLowerCase() ? 'active' : false}
```
GregWeil: This passes in a string when the button expects only true/false

ttseng: oh yeah, thanks for catching this!  fixed now

ttseng:
I set the vertical alignment of the status badges so the buttons should be aligned better now.  @GregWeil what browser were you testing with?

GregWeil:
Firefox, looks good now

image

PR #311, opened 2019-03-13
Text / Heading

oliviacpu:
Adding a Heading component and using it in the user profile presenter.

- [x] Created new Heading component
- [x] Added to Storybook
- [x] Used in user profile presenter

Areas to check:
- Storybook
- User profiles

Quick links:
- [Olivia's profile with changes](https://kiwi-poinsettia.glitch.me/@oliviacpu)
- [Olivia's profile on production](https://glitch.com/@oliviacpu)

~💁 I decoupled presentation from the tag being used, but for now it can be used as a 1:1 replacement for h1-h4 by just using `tagName`. There's a comment where things are set to default for now.~

I'm going to continue to replace the rest of the header components in small PRs after this is merged to keep the potential for conflicts low.

### DIFF
--- a/src/components/text/heading.js
+++ b/src/components/text/heading.js
@@ -15,18 +15,17 @@ const TAGS_AND_SIZES = {
 /**
  * Heading Component
  */
-const Heading = ({ children, tagName, className }) => {
+const Heading = ({ children, tagName: TagName, className }) => {
   // In the future we might want tag names and sizes to be different
 
   // For now, we're keeping them 1-1,
   // and also specifically setting 'h1' and 'h2' headings to be bold
   // which should match the current styles on production for switching to this
 
-  const size = TAGS_AND_SIZES[tagName];
-  const bold = tagName === 'h1' || tagName === 'h2';
-
+  const size = TAGS_AND_SIZES[TagName];
+  const bold = TagName === 'h1' || TagName === 'h2';
   return (
-    <tagName
+    <TagName
       className={cx({
         [size]: true,
         bold,
@@ -34,7 +33,7 @@ const Heading = ({ children, tagName, className }) => {
       })}
     >
       {children}
-    </tagName>
+    </TagName>
   );
 };
 
--- a/src/components/text/heading.js
+++ b/src/components/text/heading.js
@@ -30,7 +30,7 @@ const Heading = ({ children, tagName, className }) => {
     [className]: className !== undefined,
   });
 
-  return <tagName className={className}>{children}</tagName>;
+  return <tagName className={classNames}>{children}</tagName>;
 };
 
 Heading.propTypes = {
--- a/src/components/text/heading.js
+++ b/src/components/text/heading.js
@@ -11,27 +11,20 @@ export const SIZES = ['small']; // I'm into sizes being different from tagNames 
 /**
  * Heading Component
  */
-const Heading = ({ children, length }) => {
+const Heading = ({ children, tagName }) => {
   const className = cx({
-    btn: true,
-    cta: type === 'cta',
-    small: size === 'small',
-    tertiary: ['tertiary', 'dangerZone'].includes(type),
-    dangerZone: type === 'dangerZone',
-    hover,
+    heading: true,
   });
-  return <TAG_NAMES className={className}> /;
+  const HeadingTag = TAG_NAMES[tagName];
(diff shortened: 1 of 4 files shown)


src/components/text/heading.js
```diff
@@ -0,0 +1,53 @@
...
+  h1: 'xlarge',
+  h2: 'large',
+  h3: 'medium',
+  h4: 'small',
+};
+
+/**
+ * Heading Component
+ */
+const Heading = ({ children, tagName: TagName, className }) => {
```
oliviacpu: 💁 This is an alias while destructuring props in order to make `tagName` become `TagName` without an extra line somewhere. `<tagName />` will render as `<tagName />` , but `<TagName />` would render as `<h1 />` if that's what you passed for `tagName`.

src/components/text/heading.js
```diff
@@ -0,0 +1,53 @@
...
+  // which should match the current styles on production for switching to this
+
+  const size = TAGS_AND_SIZES[TagName];
+  const bold = TagName === 'h1' || TagName === 'h2';
+  return (
+    <TagName
+      className={cx({
+        [size]: true,
+        bold,
+        [className]: className !== undefined,
```
oliviacpu: 💁 Supporting extra class names being added. User profile presenter and other places are doing this.

sheridanvk: We were hoping to not support passing classNames down into components as part of this process. I think we should look into why the heading in the User profile has this custom class and see if we can get rid of it for consistency's sake - also I can't actually see anywhere in the styles folder that this custom login class is being used, maybe it's being used in JS but I'll take a look tomorrow!

sheridanvk:
It feels a bit weird to me to have all these tag to size translations happening in the component, meaning that the CSS is now all defined in respect to sizes, not headings. Especially given we specify all these things by tag right now, and we're just potentially adding this extra ability to define size variants for each tag - should we consider sticking with defining base styles on the actual tags as we are right now (i.e. 
.h1
  font-weight: bold
  font-size: 22px
.h2
  font-weight: bold
  font-size: 18px
  a
    color: primary
.h3
  font-size: 16px
.h4
  font-size: 14px
)
and then adding the size modifiers on them later as needed (e.g. 
.h1
  &.extra-large
    font-size: 26px
, etc)?

Seems like potentially a big default paradigm shift otherwise for how we use these styles. Perhaps I'm just not used to having both of these concepts though?

Also it's not intuitive to me that we define the bold part for h1 and h2 in the component - I'd expect to see that in the styles file too.

oliviacpu:
I think it will just be 1:1 to how it was now with no weirdness (or extra class names).

src/components/text/heading.styl
```diff
@@ -0,0 +1,17 @@
...
+  font-size: 18px
+  font-weight: bold
+  a
+    color: primary
+
+h3
+  font-size: 16px
+
+h4
+  font-size: 14px
```
sheridanvk: so the only thing here is that to truly scope the styles to this component, CSS modules need classes on all elements that you want to style, because they can't modify the actual HTML element tags (you can't do <h1-6e4df1>Heading<h1-6e4df1> - instead, they need to be able to do something like this <h1 class="h1-6e4df1">Heading</h1> - and then the h1-6e4df1 class corresponds only to the styles specified here). To achieve that we should just add . to the front of these - but admittedly this is working fine right now; it's just not working as a true CSS module.

stories/index.js
```diff
@@ -61,6 +62,12 @@ storiesOf('Text Input', module)
   .add('with error', () => <TextInput placeholder="glitch" error="That team already exists" />)
   .add('text area', () => <TextArea placeholder="[Something here] doesn't seem appropriate for Glitch because..." error="Reason is required" />);
 
+storiesOf('Heading', module)
+  .add('h1 ', () => <Heading tagName="h1">H1, 22px bold</Heading>)
+  .add('h2', () => <Heading tagName="h2">H2, 18px bold</Heading>)
+  .add('h3', () => <Heading tagName="h3">H3, 16px</Heading>)
```
sheridanvk: Huh, what's interesting here is that the user agent stylesheet is still adding font-weight:bold to h3 and h4. If it looks how the design team wants it then I guess that's fine? just seems like maybe we can just remove bold from h1 and h2, but not a big deal.

oliviacpu: Removed `bold` from the styles.

PR #313, opened 2019-03-14
Tree odometer: put new collection items at beginning of collection

modernserf:
The API orders projects in a collection by the order they're added. Make the optimistic update match this.

Fixes https://glitch.manuscript.com/f/cases/3327898/


keithk:
lgtm 👍

PR #318, opened 2019-03-15
Cubic wren: Image component and replacements on community site

keithk:
`<Image>` component:
Params:
- [x] src: Required location of the image
- [x] srcSet: To make responsive image sets, example in story book loads the day Glitch fish on smaller screens, and the night Glitch fish on larger ones. [More details](https://css-tricks.com/responsive-images-youre-just-changing-resolutions-use-srcSet/)
- [x] Alt: Required for screen readers and accessibility 
- [x] Width & Height: Can modify the width and height of the element
- [x] Role: Presentation by default, change if the image is *required* for navigating the page
- [x] className: For passing through classes, shouldn't be required for most instances, though
- [x] backgroundImage: If set it will load the image in a div, there are a lot of places on the Glitch site that we do this now

Functionality:
- [ ] Set role to presentation when `alt` is not set, otherwise require an alt
- [ ] Replace `img` around the community site, making sure to set role where needed

- [ ] Handle merge conflicts 😭 

### classNames bubbling
We had a discussion in slack about not passing through classNames, and I agreed that is an anti-pattern that we should avoid. However, while trying to replace the  around the community site it became clear that there are _so many_ custom styles that this could be fixed on a case by case basis

### Open Questions
Are there any other props that should be passed through? I couldn't find any examples of images have an `onClick` handler, for example.


src/components/image/image.js
```diff
@@ -0,0 +1,70 @@
...
+        src={src}
+        srcSet={srcSet ? srcSet : undefined}
+        sizes={sizes}
+        width={width || undefined}
+        height={height || undefined}
+        className={className || undefined}
+        alt={alt}
+        role={role}
+      />}
+  {backgroundImage === true &&<div
```
keithk: I included a background image option just because this is something I've needed to do a lot in the past. It could be spun out into it's own component, but it uses a lot of the same props. 🤷‍♂️

src/components/image/image.js
```diff
@@ -0,0 +1,70 @@
...
+ * @param {number} width - Image width
+ * @param {number} height - Image height
+ * @param {string} role - Image role (typically presentation)
+ * @param {object} className - extra classes to be passed down  to the component
+ * @param {boolean} backgroundImage - If we want the image to be rendered as a background image
+ */
+
+const Image = ({ alt, backgroundImage, backgroundRatio, className, height, role, src, srcSet, sizes, width }) => (
+  <>
+    {!backgroundImage ? (
```
keithk: This _could_ be moved to a separate component (`<BackgroundImage>`) if opinions are strong enough!

src/components/image/image.js
```diff
@@ -0,0 +1,70 @@
...
+/**
+ * 🖼️ Image Component
+ *
+ * @param {string} src - Image source
+ * @param {array} srcSet - Responsive image source set
+ * @param {alt} alt - Alternative text
+ * @param {number} width - Image width
+ * @param {number} height - Image height
+ * @param {string} role - Image role (typically presentation)
+ * @param {object} className - extra classes to be passed down  to the component
```
sarahzinger: why is lassName an object and not a string. Also this can be filed under stupid questions but do we think height/weight could just be passed as part of a class? Or do we generally want to discourage that you think?

keithk: @sarahzinger In case className has been passed as an instance of `cx` from another component. It actually takes a string or an object, I should update that!

The goal is to get the least amount of className bubbling down as possible, so directly setting the height/width is more in line with that

src/presenters/pages/category.js
```diff
@@ -23,7 +24,7 @@ const CategoryPageWrap = ({ addProjectToCollection, api, category, currentUser,
         <header className="collection">
           <h1 className="collection-name">{category.name}</h1>
           <div className="collection-image-container">
-            <img src={category.avatarUrl} alt="" />
+            <Image src={category.avatarUrl} role="presentation" />
```
sarahzinger: it it's presentation by default maybe we don't need to send it?

keithk: Good catch, that was from before I made it default!

src/components/image/image.js
```diff
@@ -0,0 +1,70 @@
...
+  </>
+);
+
+Image.propTypes = {
+  alt: PropTypes.string.isRequired,
+  backgroundImage: PropTypes.bool,
+  backgroundRatio: PropTypes.number,
+  className: PropTypes.oneOfType([PropTypes.string, PropTypes.instanceOf(Object)]),
+  height: PropTypes.number,
+  role: PropTypes.string,
```
sarahzinger: so I had a thought, feel free to totally disagree with it. Do you think this should be required and we explain helpful defaults? I think this way you're reminded to consider if it's presentational or not? I could see myself forgetting this is a thing and then it defaults to presentation when it's actually something else

keithk: Hmmm, that's a good question. I'm not sure how often images _aren't_ just presentation, if we use them on the site to convey instructions or otherwise

src/components/image/image.js
sarahzinger: oh interesting about cx I thought it always returned a string in the end? But maybe I'm wrong about that

cool bout height/width!

keithk: will check 👀

sheridanvk: Just curious what the advantage is of specifying types here as well as in PropTypes? If you think it's useful additional information, should we update all the other components if we're going to do it here (for consistency)?

src/components/image/image.js
sheridanvk: yeah so far it seems we only have presentation images? But if you have good guidelines on this for when they shouldn't be, I'd be fine with making this required - perhaps we're missing cases where they should be something else.

keithk:
@sheridanvk Can you point out which ones I'm missing? I've got regular image, background image, both of those with width/height set, and a srcSet example!

src/presenters/pages/category.js
```diff
@@ -23,7 +24,7 @@ const CategoryPageWrap = ({ addProjectToCollection, api, category, currentUser,
         <header className="collection">
           <h1 className="collection-name">{category.name}</h1>
           <div className="collection-image-container">
-            <img src={category.avatarUrl} alt="" />
+            <Image src={category.avatarUrl} />
```
sheridanvk: I'm not sure what the answer is for this, but I wonder if we should be making height and width required on the component, and not rely on having these be passed down through the styles on the classname? I know this could cause problems for media queries and such though, so perhaps this is something to look at down the line when we have better componentisation of parent components - perhaps the image can be 100% height and width of the containing div, and the parent component sets the size and position of the div based on media queries?

src/presenters/pages/index.js
```diff
@@ -64,13 +64,12 @@ class WhatIsGlitch extends React.Component {
         <span>
           <figure>
             <h1>
-              <img className="witch large" src={witchLarge} alt={whatsGlitchAlt} />
-              <img className="witch small" src={witchSmall} alt={whatsGlitchAlt} />
+              <Image src={witch} alt={whatsGlitchAlt} sizes="(max-width: 320px) 280px, 650px" />
```
sheridanvk: oh wow, I didn't know what the sizes property was, this is great. I wonder, should we be supporting srcset too so we don't have to get rid of the small witch here (does that result in faster mobile loads due to the smaller image size)? https://bitsofco.de/the-srcset-and-sizes-attributes/

stories/index.js
```diff
@@ -54,18 +55,36 @@ storiesOf('TooltipContainer', module)
...
+      srcSet={[
+        'https://cdn.glitch.com/2bdfb3f8-05ef-4035-a06e-2043962a3a13%2Flogo-day.svg?x=2 1000w',
+        'https://cdn.glitch.com/2bdfb3f8-05ef-4035-a06e-2043962a3a13%2Flogo-night.svg?x=1 2000w',
+      ]}
+      alt="Glitch Logo"
+    />
   ))
-  .add('with error', () => (
-    <TextInput placeholder="glitch" error="That team already exists"/>
+  .add('width & height', () => (
```
keithk: You'll probably need to open this storybook file and look at these stories, I'm not sure how / why the formatting messed up in this way 🤦‍♂️

src/presenters/pages/index.js
sheridanvk: gah I just realised you do have srcSet :/ but curious why you didn't use it here?

stories/index.js
sheridanvk: all good, I can see them in the guide now, I must have somehow missed these before!

stories/index.js
```diff
@@ -54,18 +55,36 @@ storiesOf('TooltipContainer', module)
...
+    <Image backgroundImage={true} src="https://cdn.glitch.com/2bdfb3f8-05ef-4035-a06e-2043962a3a13%2Flogo-day.svg" alt="Glitch Logo" />
   ))
-  .add('search', () => (
-    <TextInput type="search" opaque={true} search={true} placeholder="bots, apps, users"/>
+  .add('srcSet', () => (
+    <Image
+      src="https://cdn.glitch.com/2bdfb3f8-05ef-4035-a06e-2043962a3a13%2Flogo-day.svg"
+      srcSet={[
+        'https://cdn.glitch.com/2bdfb3f8-05ef-4035-a06e-2043962a3a13%2Flogo-day.svg?x=2 1000w',
+        'https://cdn.glitch.com/2bdfb3f8-05ef-4035-a06e-2043962a3a13%2Flogo-night.svg?x=1 2000w',
```
sheridanvk: hah I like how this story changes the fish from day to night at the widest point, very cool!

src/presenters/pages/index.js
keithk: @sheridanvk Yeah, I missed that -- added them back in! (and you can actually rely on the browser to make the smart decision itself, without needing sizes in most cases)

sheridanvk: oh that's good to know, thanks!

src/presenters/pages/category.js
keithk: I made the defaults 100% for height / width, I think that addresses this, where you can size it with the parent div. I think making the default works better than a requirement, otherwise we'd have a ton of <Image width='100%' height='100%' src='etc> in the code

I think in most cases this will be the thing that we're looking for and so it can be overridden by individual components.

keithk:
I'm going to fix up the merge conflicts here and then merge this in, and then make another PR to do replacements on the community site 🎉

PR #320, opened 2019-03-15
Band caterpillar: emphasize "show more projects"

modernserf:
Screen Shot 2019-03-15 at 1 45 33 PM

Add a fade-out gradient to the projects list. See https://glitch.manuscript.com/f/cases/3327693/


src/components/containers/expander.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
```
modernserf: - Is this component built correctly?
- Is this a reasonable category for this component?
- Should the Readme expander be refactored to use this component?

keithk:
I like this a lot 👀

styles/profile.styl
```diff
@@ -77,9 +77,9 @@ anonAvatar = "https://cdn.gomix.com/01a72eba-647e-4abf-abe6-89b02e388a7f%2Fanon-
     padding: 0
     margin-bottom: 1rem
     display: inline-block
-    overflow: visible
+    //overflow: visible
```
modernserf: These were interfering with the container, and didnt appear to be doing anything useful. Am I missing something?

modernserf:
Other design questions:
- should we add an additional row of visible projects?
- should we remove the count from the button?

src/components/containers/expander.js
sheridanvk: Yes yes and yes! this all looks great :)

styles/profile.styl
sheridanvk: agree that this looks like it can be removed

PR #321, opened 2019-03-15
juicy-nebula 👾: Webpack caching 

keithk:
This adds two new packages to the webpack configuration:

# [Speed Measure Webpack](https://github.com/stephencookdev/speed-measure-webpack-plugin/)
This gives us greater insight into the webpack builds, so we can determine what loader modules are taking up the most time to build. This could be taken out and only put back in when we want to take a look at our webpack build, but it didn't add much time on top so seems like useful information to have at our fingertips!

The output:
```SMP  ⏱  
General output time took 82,106 ms
 SMP  ⏱  Plugins
LodashModuleReplacementPlugin took 2,080 ms
MiniCssExtractPlugin took 1,092 ms
StatsPlugin took 3 ms
 SMP  ⏱  Loaders

modules with no loaders took 66,315 ms
  median       = 307 ms
  mean         = 1,118 ms
  s.d.         = 2,371.888 ms
  range        = (0 ms --> 44,436 ms)
  module count = 1141

babel-loader, and eslint-loader took 61,097 ms
  median       = 3,568 ms
  mean         = 3,480 ms
  s.d.         = 2,496.432 ms
  range        = (178 ms --> 10,449 ms)
  module count = 143

mini-css-extract-plugin, and css-loader, and stylus-loader took 29,101 ms
  median       = 2,002 ms
  mean         = 4,504 ms
  s.d.         = 5,010.942 ms
  range        = (428 ms --> 14,026 ms)
  module count = 8

css-loader, and  stylus-loader took 26,878 ms
  median       = 1,895 ms
  mean         = 4,214 ms
  s.d.         = 4,570.133 ms
  range        = (413 ms --> 12,229 ms)
  module count = 8

babel-loader took 9,743 ms
  median       = 1,543 ms
  mean         = 3,248 ms
  s.d.         = 3,002.068 ms
  range        = (1,486 ms --> 6,714 ms)
  module count = 3
```

# [Cache Loader](https://github.com/webpack-contrib/cache-loader/)
This creates a `.cache-loader` directory and puts webpack build cache there. The general concern with this approach is that it can create a lot of disk space for the cache but over my testing the difference between the two containers was 2mb.

After a few tests of changing the same files in a fresh ~community build and this branch, these were the results:

### Changing a .styl file
fresh: 11690ms
with cache: 3987ms

### Changing a react component .js file
fresh: 3126ms
Time: 837ms

## Open Qs
1. I've added the `.cache-loader` directory to .gitignore, but I wonder if that would actually help with fresh builds if it were pushed up?


keithk:
@sheridanvk Can you take a look at this in the AM? It's working correctly (both on watch and build) on my juicy-nebula branch, but not when it goes to the test runner.

sheridanvk:
happy to! can you add me to the project when you get a sec?

PR #323, opened 2019-03-18
Better settee: sample collections _before_ fetching all of their data

modernserf:
We don't need to fetch the projects, users, etc of collections that won't even make it out of this function

### DIFF
--- a/src/presenters/includes/included-in-collections.js
+++ b/src/presenters/includes/included-in-collections.js
@@ -1,13 +1,14 @@
 import React, { useState, useEffect } from 'react';
 import { sampleSize } from 'lodash';
-import { getSingleItem, getAllPages, allByKeys } from '../../../shared/api';
+import { getSingleItem, getPages, getAllPages, allByKeys } from '../../../shared/api';
 
 import CollectionItem from '../collection-item';
 
 const getIncludedCollections = async (api, projectId) => {
   const collections = await getAllPages(api, `/v1/projects/by/id/collections?id=${projectId}&limit=100&orderKey=createdAt&orderDirection=DESC`);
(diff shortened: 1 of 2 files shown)


src/presenters/includes/included-in-collections.js
```diff
@@ -6,8 +6,9 @@ import CollectionItem from '../collection-item';
 
 const getIncludedCollections = async (api, projectId) => {
   const collections = await getAllPages(api, `/v1/projects/by/id/collections?id=${projectId}&limit=100&orderKey=createdAt&orderDirection=DESC`);
-  const withData = await Promise.all(
-    collections.map(async (collection) => {
+  const selectedCollections = sampleSize(collections, 3);
```
modernserf: the actual change: move `sampleSize` before fetching associated projects, user, team

GregWeil:
The change itself seems all good, but it seems like something is forcing redundant data loads. According to the network tabs, it'll sometimes load the collections and projects up to four times.

src/presenters/includes/included-in-collections.js
```diff
@@ -16,13 +17,12 @@ const getIncludedCollections = async (api, projectId) => {
...
   );
-  return sampleSize(withData, 3);
 };
 
-const useAsync = (asyncFunction, ...args) => {
+const useAsync = (asyncFunction, api, ...args) => {
   const [result, setResult] = useState(null);
   useEffect(() => {
-    asyncFunction(...args).then(setResult);
+    asyncFunction(api, ...args).then(setResult);
```
modernserf: @GregWeil i think this fixes the issue with the redundant fetching.

PR #325, opened 2019-03-18
Branched curio - adds word break so long URL overflow doesn't happen in chrome

jennschiffer:
to test, see the project description https://branched-curio.glitch.me/~eight-lamp in chrome. it was already word-break'in in firefox.

fixes case https://branched-curio.glitch.me/~eight-lamp


styles/profile.styl
```diff
@@ -44,6 +44,7 @@ anonAvatar = "https://cdn.gomix.com/01a72eba-647e-4abf-abe6-89b02e388a7f%2Fanon-
   > *
     margin: 0
     width: 100%
+    word-break: break-all
```
sarahzinger: maybe break-word instead of break-all?

Screen Shot 2019-03-18 at 2 46 54 PM
Screen Shot 2019-03-18 at 2 47 09 PM

sarahzinger: hmm jk maybe that's not actually usable everywhere https://stackoverflow.com/questions/47413677/alternative-for-word-break-break-word

src/curated/featured.js
```diff
@@ -14,18 +14,18 @@ Example:
 // make sure image urls use https
 export default [
   {
-    title: 'Shout out your peoples',
-    img: 'https://cdn.glitch.com/02863ac1-a499-4a41-ac9c-41792950000f%2Fglitch_shoutout.jpg?1552302001451',
-    link: 'https://glitch.com/~shout-out',
+    title: 'Reach Out and Touch',
```
sarahzinger: I'm guessing these things are on glitch.com live right now, and github just doesn't have the latest and greatest yeah?

taravancil: Yep! It's pretty normal to see this on files that the marketing team touches

PR #326, opened 2019-03-18
Tangible radar: prevent setting state after component has unmounted

modernserf:
Add a (somewhat inelegant) check in `add-project-to-collection-pop` to prevent setting state after component has unmounted.


src/presenters/pop-overs/add-project-to-collection-pop.js
```diff
@@ -144,6 +144,10 @@ class AddProjectToCollectionPop extends React.Component {
     this.loadCollections();
   }
 
+  componentWillUnmount() {
```
modernserf: Alternatively, we could [cancel the axios request](https://github.com/axios/axios#cancellation), though i'm not sure if the extra setup would be worthwhile.

PR #328, opened 2019-03-18
Butter onyx: don't warn users about leaving their teams projects

modernserf:
See https://glitch.manuscript.com/f/cases/3328083/


src/presenters/pop-overs/project-options-pop.js
```diff
@@ -24,7 +24,22 @@ const ProjectOptionsContent = ({ addToCollectionPopover, ...props }) => {
     props.togglePopover();
   }
 
+  function isTeamProject() {
+    const { currentUser, project } = props;
+    for (const team of currentUser.teams) {
```
sarahzinger: stupid q but teams is always defined and at least an empty [ ] right?

sarahzinger:
Looks good to me! 🚢

src/presenters/pop-overs/project-options-pop.js
modernserf: it _should_ be -- currentUser still uses the old API that has everything on it.
if this turns out to not be the case i'd rather this crash than fail silently.

PR #330, opened 2019-03-18
Wooded moose: improve flow when removing member from team

modernserf:
See https://glitch.manuscript.com/f/cases/3327687/

When removing a user from a team, there's a second popover that asks if you want to (also) remove them from the team's projects, or just the team itself. This is unnecessary and somewhat confusing if they aren't a member of any projects.

This PR moves much of the logic for removing team members from the `remove` popover and into the `info` popover, so that the remove one doesn't even need to be rendered in this particular case.

NB this is an excellent example of how putting our logic nearest to the UI that uses it makes refactoring kind of complicated.

### DIFF
--- a/src/presenters/pop-overs/team-user-info-pop.js
+++ b/src/presenters/pop-overs/team-user-info-pop.js
@@ -1,7 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
-import { getAvatarThumbnailUrl } from '../../models/user';
+import { getAvatarThumbnailUrl, getDisplayName } from '../../models/user';
 
 import { TrackClick } from '../analytics';
 import { NestedPopover } from './popover-nested';
@@ -75,7 +75,7 @@ const TeamUserInfo = ({ currentUser, currentUserIsTeamAdmin, showRemove, ...prop
   const currentUserHasRemovePriveleges = currentUserIsTeamAdmin || (currentUser && currentUser.id === props.user.id);
   const canRemoveUser = !(props.userIsTheOnlyMember || props.userIsTheOnlyAdmin);
   const canCurrentUserRemoveUser = canRemoveUser && currentUserHasRemovePriveleges;
-
+  
   return (
     <dialog className="pop-over team-user-info-pop">
       <section className="pop-over-info user-info">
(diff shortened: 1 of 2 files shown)


src/presenters/pop-overs/team-user-remove-pop.js
```diff
@@ -68,18 +53,19 @@ class TeamUserRemovePopBase extends React.Component {
...
 
   render() {
-    const allProjectsSelected = this.state.userTeamProjects.every((p) => this.state.selectedProjects.has(p.id));
+    const userTeamProjects = this.props.userTeamProjects.data || [];
+    const allProjectsSelected = userTeamProjects.every((p) => this.state.selectedProjects.has(p.id));
     const userAvatarStyle = { backgroundColor: this.props.user.color };
 
     let projects = null;
-    if (this.state.gettingUser) {
+    if (this.props.userTeamProjects.status === 'loading') {
```
modernserf: it might make more sense to have this loader in the `info` component now?

src/presenters/pop-overs/team-user-info-pop.js
```diff
@@ -69,13 +69,22 @@ const ThanksCount = ({ count }) => (
...
+const TeamUserInfo = ({ currentUser, currentUserIsTeamAdmin, showRemove, userTeamProjects, removeUser, ...props }) => {
   const userAvatarStyle = { backgroundColor: props.user.color };
 
   const currentUserHasRemovePriveleges = currentUserIsTeamAdmin || (currentUser && currentUser.id === props.user.id);
   const canRemoveUser = !(props.userIsTheOnlyMember || props.userIsTheOnlyAdmin);
   const canCurrentUserRemoveUser = canRemoveUser && currentUserHasRemovePriveleges;
 
+  // if user is a member of no projects, skip the confirm step
+  function onRemove() {
+    if (userTeamProjects.status === 'ready' && userTeamProjects.data.length === 0) {
```
sarahzinger: just for my know brain, what's this status thingy?

sarahzinger: oh jk I could just read lol plz ignore

src/presenters/pop-overs/team-user-info-pop.js
```diff
@@ -122,11 +131,31 @@ const TeamUserInfo = ({ currentUser, currentUserIsTeamAdmin, showRemove, ...prop
...
+  const { createNotification } = useNotifications();
+  const [userTeamProjects, setUserTeamProjects] = useState({ status: 'loading', data: null });
+  useEffect(() => {
+    props.api.get(`users/${props.user.id}`).then(({ data }) => {
+      setUserTeamProjects({
+        status: 'ready',
+        data: data.projects.filter((userProj) => props.team.projects.some((teamProj) => teamProj.id === userProj.id)),
+      });
+    });
+  }, [props.user.id]);
```
sarahzinger: does it make sense to make these data calls at a higher level? Would we make an api call every time we open a popover?

sarahzinger:
cool stuff I had one question, but I don't know if it really makes sense to work on at this point in time or not. Thanks for sharing I learned new things while reading this pr :) 

🚢 !

src/presenters/pop-overs/team-user-info-pop.js
taravancil: No useful feedback, just want to say I'm glad we're using hooks 😸

PR #333, opened 2019-03-19
Fixes to component styles

sheridanvk:
I think the tooltip component was build according to a bad implementation I'd done in the early days of CSS modules, and it was importing in the globals.styl file from the styles directory. This meant a bunch of styles that shouldn't have been available to the other components were available, so components looked fine in storybook when in fact they weren't self-composed.

Fixed a bunch of things but please let me know if you can see anything I missed! Everything is looking good at http://lumpy-organ.glitch.me/storybook

### DIFF
--- a/src/components/text/heading.styl
+++ b/src/components/text/heading.styl
@@ -3,16 +3,18 @@
 .heading
   font-family: sansserif
 
-.h1
-  font-size: 22px
-  
-.h2
-  font-size: 18px
-  a
-    color: primary
+  &.h1
+    font-size: 22px
+  font-weight: bold
 
-.h3
-  font-size: 16px
+  &.h2
+    font-size: 18px
+  font-weight: bold
+    a
+      color: primary
 
-.h4
-  font-size: 14px
+  &.h3
+    font-size: 16px
+
+  &.h4
+    font-size: 14px
--- a/src/components/text/heading.styl
+++ b/src/components/text/heading.styl
@@ -5,7 +5,7 @@
 
   &.h1
     font-size: 22px
-  font-weight: bold
+    font-weight: bold
 
   &.h2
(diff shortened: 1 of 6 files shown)


src/components/text/heading.styl
```diff
@@ -1,15 +1,20 @@
...
-
-.h4
-  font-size: 14px
\ No newline at end of file
+.heading
+  font-family: sansserif
+
+  &.h1
+    font-size: 22px
+  font-weight: bold
```
oliviacpu: I think this indentation is causing `font-weight: bold` to be applied to `.heading`

sheridanvk: good catch - fixed, thanks!

taravancil:
Good catch on this, thanks!

src/components/text/markdown.styl
```diff
@@ -2,6 +2,8 @@
 
 .markdownContent
   text-rendering: optimizeLegibility
+  font-family: sansserif
```
ttseng: hey!  just noticed there is a small regression here where markdownContent shouldn't default to sansserif (it should be shown with menlo font if it's editable):

image

I'm going to do a more thorough check on styling to see if I can find anything else and will put that in its own PR.

sheridanvk:
thanks Tiff, good catch! I think it is monospace after clicking in, but not
before the click. Is that the behaviour you're going to fix as well? I
guess we need to add the .content-editable class earlier.

sheridanvk:
Ahh that's actually the problem - there is no content-editable class on
Markdown. It was incorrectly inheriting its styles from global.styl acting
on its parent before - we should be passing the .content-editable class
from EditableDescription down into Markdown so we can change the font
accordingly.

PR #334, opened 2019-03-19
Marred driver: Update anonymous user login process

ttseng:
Testable at https://marred-driver.glitch.me/

**Summary of changes**
* removed anonymous user avatar from header
* added info section at top of sign-in-pop to let users know they can create an account by logging in
image
* updated styling of recent projects section for anonymous users with banner to let them know that anonymous projects expire after 2 weeks.
image
* added clear session button to recent project sections for anonymous users with updated alert message
image

to test a dummy account for an anonymous user with a single project, type this in your console window 
```
window.localStorage.cachedUser = `{\"persistentToken\":\"ce1b86a4-0d42-4b2c-865a-710af7d267c8\"}`
```

Note: I didn't clean `sign-in-pop` to get rid of anonymous user handling there.  Maybe I should?  I figured right now it's just hidden, so there's no reason we couldn't re-expose it in the future if we wanted to.


ttseng:
merged the latest from master to get rid of passing the api as props.

also updated the clear session message since *technically* projects aren't removed (you can still access an anonymous user profile even after they clear their session, e.g. https://marred-driver.glitch.me/user/3515235)

image

PR #335, opened 2019-03-19
Moor botany: [SPIKE] handle currentUser in redux

modernserf:
This PR is a first pass at handling currentUser in redux. it will be closed and replaced with many smaller PRs.


modernserf:
this is _probably_ how i will break this up:

- replace all uses of `<CurrentUserConsumer>` with hooks (lots of small changes)
- put API into context and remove plumbing (lots of small changes)
- move currentUser, localStorage etc into `/state` directory (lots of small changes)
- rewrite currentUser as redux state (a few big changes)

PR #338, opened 2019-03-20
Trail bail: move current-user, local-storage to 'state' directory

modernserf:
This is some initial work for handling current-user in redux, mostly to keep the "real work" PRs easier to follow.


clottman:
Is @sheridanvk's work on import aliases complete? Would be awesome if it could just be "state/userContext" instead of "../../..state/userContext" etc. Not that we need to change all these, just make sure that the new folder is included in the list of available aliases.

PR #341, opened 2019-03-20
Romantic barracuda: Bug Fixes to Collection Grid Styling

ttseng:
Removed grid css to fix styling issues on collection pages.

Also removed project heading with counter to make the layout a bit more readable with annotations.

Testable links:
https://romantic-barracuda.glitch.me/@scientiffic/glitch-things

vs 
https://glitch.com/@scientiffic/glitch-things

image
image

### DIFF
--- a/src/presenters/pages/collection.js
+++ b/src/presenters/pages/collection.js
@@ -179,35 +179,6 @@ const CollectionPageContents = ({
                   <ProjectsUL projects={collection.projects} collection={collection} api={api} projectOptions={{}} />
                 )}
               </div>
-              {currentUserIsAuthor && (
-                <ProjectsUL
-                  {...props}
-                  projects={collection.projects}
-                  collection={collection}
-                  api={api}
-                  hideNote={hideNote}
-                  projectOptions={{
-                    removeProjectFromCollection,
-                    addProjectToCollection,
-                    updateOrAddNote,
-                    addNoteField,
-                  }}
-                />
-              )}
-              {!currentUserIsAuthor && userIsLoggedIn && (
-                <ProjectsUL
-                  {...props}
-                  projects={collection.projects}
-                  collection={collection}
-                  api={api}
-                  projectOptions={{
-                    addProjectToCollection,
-                  }}
-                />
-              )}
-              {!currentUserIsAuthor && !userIsLoggedIn && (
-                <ProjectsUL projects={collection.projects} collection={collection} api={api} projectOptions={{}} />
-              )}
             </>
           )}
         </article>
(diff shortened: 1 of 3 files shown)


src/presenters/pages/collection.js
```diff
@@ -121,88 +119,64 @@ const CollectionPageContents = ({
               />
             </div>
 
+            <div className="collection-project-count">{collection.projects.length} Projects</div>
```
sarahzinger: I think there's like significance to heading vs div from accessibility perspectives yeah? maybe better to use the new Text component or something?

styles/pages/collection.styl
```diff
@@ -119,6 +115,8 @@
   .project-options
     display: block
     background: none
+  .project-options:hover
```
clottman: Should this also happen on :focus?

src/presenters/pages/collection.js
taravancil: There is, but the projects count strikes me more as supplementary info rather than the title for a section of the document, so yeah, I *think* the Text component makes sense here

src/presenters/pages/collection.js
```diff
@@ -121,88 +119,64 @@ const CollectionPageContents = ({
...
-                    !currentUserIsAuthor && !userIsLoggedIn && (
-                      <ProjectsUL
-                        projects={collection.projects}
-                        collection={collection}
-                        api={api}
-                        projectOptions={{}}
-                      />
-                    )
-                  }
+          {!collectionHasProjects && currentUserIsAuthor && (
```
sarahzinger: this is just like code styling diffs here right? hard to see what's changed

src/presenters/pages/collection.js
keithk: Agree with Tara, this doesn't feel like the most important (title) information

styles/pages/collection.styl
```diff
@@ -88,18 +92,10 @@
   .projects-container
     margin: 30px 30px 30px 16px
```
keithk: This isn't new, but does this look better as 30px 16px, so that it's equal margin on both sides of the container?

src/presenters/pages/collection.js
ttseng: cool, I updated it using the new text component

src/presenters/pages/collection.js
ttseng: yeah, I think it just shows up this way since some parts were rearranged (to move the header out)

styles/pages/collection.styl
ttseng: good catch!

PR #342, opened 2019-03-20
Wiggly snap: all current user handled via hooks

modernserf:
Replace all uses of `<CurrentUserConsumer>` in preparation for redux.

### DIFF
--- a/src/state/current-user.js
+++ b/src/state/current-user.js
@@ -281,13 +281,6 @@ CurrentUserProvider.propTypes = {
   children: PropTypes.func.isRequired,
 };
 
-export const CurrentUserConsumer = (props) => (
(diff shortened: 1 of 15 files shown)


src/state/current-user.js
```diff
@@ -281,13 +281,6 @@ CurrentUserProvider.propTypes = {
...
 };
 
-export const CurrentUserConsumer = (props) => (
-  <Context.Consumer>{({ currentUser, fetched, ...funcs }) => props.children(currentUser, fetched, funcs, props)}</Context.Consumer>
-);
-CurrentUserConsumer.propTypes = {
-  children: PropTypes.func.isRequired,
-};
-
 export const useCurrentUser = () => React.useContext(Context);
```
taravancil: ugh, so nice

PR #344, opened 2019-03-20
Leather reading: Display message to IE users

keithk:
This displays a message to users in IE 11, that gives them recommendations for other browsers to use while browsing glitch.

Screenshot:
image
(I love this image)


views/index.ejs
```diff
@@ -79,10 +79,26 @@
...
+            <div class="error-msg" id="fallback-ie" style="display: none">
+              <h1>Glitch doesn't support Internet Explorer</h1>
+              <p>Please upgrade to a browser that can support Glitch features. Please use 
+                <a href="https://www.mozilla.org/en-US/firefox/">Firefox</a>, <a href="https://www.google.com/chrome/">Chrome</a> or <a href="https://www.microsoft.com/en-us/windows/microsoft-edge">Edge</a>.</p>
+            </div>
           </main>
       </div>
     </div>
 
+    <script>
```
keithk: This is inline because it *needs* to come above the scripts being loaded (which kills the page). The other option is to create a small .js file for this and pull it in, but this seemed easier

views/index.ejs
```diff
@@ -79,10 +79,26 @@
               <h1>The site isn&apos;t quite ready yet</h1>
               <p>Try refreshing in a few moments.</p>
             </div>
+            <div class="error-msg" id="fallback-ie" style="display: none">
+              <h1>Glitch doesn't support Internet Explorer</h1>
```
keithk: Open to copy changes!

styles/pages/error.styl
```diff
@@ -42,7 +42,7 @@
     background: secondary-background 
     border: 3px solid black
     border-radius: 10px
-    *
+    h1, p
```
keithk: This was creating a margin 20px on any element in the div, which included the links 😭

sarahzinger: wowza, does this affect anything else? I see we use error-msg in a lot o' places

views/index.ejs
sarahzinger: yet? growth mindset? idk maybe not a good idea to say yet unless it's on the roadmap

sheridanvk: How about we remove the double "Please"? "Please upgrade to a web browser that includes the Javascript features we need - e.g. Firefox, Chrome or Edge"

views/index.ejs
```diff
@@ -79,10 +79,26 @@
...
+              <p>Please upgrade to a browser that can support Glitch features. Please use 
+                <a href="https://www.mozilla.org/en-US/firefox/">Firefox</a>, <a href="https://www.google.com/chrome/">Chrome</a> or <a href="https://www.microsoft.com/en-us/windows/microsoft-edge">Edge</a>.</p>
```
sarahzinger: just out of curiosity why these browsers and not others, although tbh this seems like a good list to me

views/index.ejs
```diff
@@ -79,10 +79,26 @@
...
+      var isIe = /Trident\/[7]{1}/i.test(navigator.userAgent);
+      if (isIe) {
+        var fallback = document.getElementById('fallback');
+        var fallbackIE = document.getElementById('fallback-ie');
+        var fallbackImg = document.getElementById('fallback-img');
+        fallback.style.display = 'block';
+        fallbackIE.style.display = 'block';
+        fallbackImg.style.display = 'block';
+      }
+    </script>
```
sheridanvk: We should be consistent with capitalisation between fallbackIE and isIe variable names - personally I like the IE variant but happy with whatever

views/index.ejs
sarahzinger: yeah I think it should be chill? I know inlines have some security concerns but I think that only matters if the user can like submit a form to alter the page they're looking at. I could see this fallback script getting bigger and including other browsers that we have trouble with in the future. If that day comes maybe then we can move it into a separate script?

views/index.ejs
sarahzinger: I could see an argument for keeping this browser text more generic, like "Glitch doesn't support this browser" then again I could some people being like "whats a browser" lol.

styles/pages/error.styl
keithk: Good call, I'm going to take a look! The ones on this page are all a single H1 and P tag (really they could just be wrapped and give that margin)

views/index.ejs
keithk: 🤷‍♂️ The most common + the MS official browser (since they're on windows)

PR #345, opened 2019-03-20
Supporting links in the Button component and adding the Emoji component

sheridanvk:
https://lumpy-organ.glitch.me

This PR:
1. Supports links in Buttons (and therefore allows us to get rid of ButtonLink in the presenters/includes/project-actions file)
2. Introduces the Emoji component
3. Supports Emoji in Buttons

Items to pay attention to are the Edit, Remix and Show buttons on user, team and project pages.

Screenshot 2019-03-20 at 2 50 00 PM

Screenshot 2019-03-20 at 2 56 11 PM

I'll follow up with replacing all buttons in a subsequent PR (I still need to modify PopoverWithButton before we fully support the new component).

Storybook has been updated here. 
http://lumpy-organ.glitch.me/storybook

Also per Tiff's request, have updated the "danger-zone" button to be red on hover, not off hover - but note that this isn't in use yet, due to PopoverWithButton still using old style buttons.


package.json
```diff
@@ -17,6 +17,7 @@
     "@sentry/browser": "^4.5.3",
     "@sentry/node": "^4.4.1",
     "@storybook/react": "^4.1.1",
+    "@storybook/addon-info": "^4.0.12",
     "autoprefixer-stylus": "^0.14.0",
```
sheridanvk: only this is an addition; just alphabetised the list otherwise

src/components/buttons/button.js
```diff
@@ -2,40 +2,67 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames/bind';
 import styles from './button.styl';
+import { Link } from '../../presenters/includes/link';
 
 const cx = classNames.bind(styles);
 
-export const TYPES = ['tertiary', 'cta', 'dangerZone'];
+export const TYPES = ['tertiary', 'cta', 'dangerZone', 'dropDown'];
```
sheridanvk: dropDown will be used for PopoverWithButton - it removes all styling.

src/components/buttons/button.js
```diff
@@ -2,40 +2,67 @@ import React from 'react';
...
   const className = cx({
     btn: true,
     cta: type === 'cta',
-    small: size === 'small',
+    small: size === 'small' || type === 'dangerZone', // we want to demphasize dangerous actions, so we make them small
     tertiary: ['tertiary', 'dangerZone'].includes(type),
     dangerZone: type === 'dangerZone',
+    unstyled: type === 'dropDown',
+    hasEmoji: React.Children.toArray(children).some((child) => child.type && (child.type.name === 'Emoji' || 'Image')),
+    transparent,
```
sheridanvk: Checks to see if any emoji are in this button; if so it adds the appropriate styles

src/components/buttons/button.js
```diff
@@ -2,40 +2,67 @@ import React from 'react';
...
+    }
+    return null;
+  },
+  /** OR link when button clicked */
+  href: (props, propName, componentName) => {
+    if (!props.onClick && !props.href) {
+      return new Error(`One of props 'href' or 'onClick' was not specified in '${componentName}'.`);
+    }
+    return null;
+  },
```
sheridanvk: ensures that either onClick OR href are specified.

src/components/buttons/button.js
```diff
@@ -44,16 +71,20 @@ Button.propTypes = {
   size: PropTypes.oneOf(SIZES),
   /** whether or not the button's hover state should be active */
   hover: PropTypes.bool,
+  /** make the button take on the background color */
+  transparent: PropTypes.bool,
```
sheridanvk: This property will be used for secondary buttons in popovers, which appear grey like their background.

src/components/images/emoji.js
```diff
@@ -0,0 +1,60 @@
+import React from 'react';
```
sheridanvk: Please let me know if this component isn't straightforward to understand!

src/presenters/includes/link.js
```diff
@@ -8,8 +8,8 @@ import { getLink as getProjectLink } from '../../models/project';
 import { getLink as getTeamLink } from '../../models/team';
 import { getLink as getUserLink } from '../../models/user';
 
-/* global EXTERNAL_ROUTES */
-const external = Array.from(EXTERNAL_ROUTES);
+/* global */
+const external = window.EXTERNAL_ROUTES ? Array.from(window.EXTERNAL_ROUTES) : [];
```
sheridanvk: adding this enables Links to be used in Storybook, where we don't have the EXTERNAL_ROUTES global variable

src/presenters/includes/project-actions.js
```diff
@@ -2,67 +2,44 @@ import React from 'react';
 import PropTypes from 'prop-types';
 
 import { getShowUrl, getEditorUrl, getRemixUrl } from '../../models/project';
-import { Link } from './link';
+import Button, { SIZES } from '../../components/buttons/button';
+import Emoji from '../../components/images/emoji';
 
-const showIcon = 'https://cdn.glitch.com/6ce807b5-7214-49d7-aadd-f11803bc35fd%2Fshow-app.svg';
-
-const ButtonLink = ({ href, children, className, ...props }) => (
```
sheridanvk: ButtonLink has been removed, as it's no longer needed

src/presenters/includes/project-actions.js
```diff
@@ -2,67 +2,44 @@ import React from 'react';
...
-  className: '',
 };
 
-export const EditButton = ({ name, isMember, ...props }) => (
-  <ButtonLink href={getEditorUrl(name)} {...props}>
+export const EditButton = ({ name, isMember, size }) => (
+  <Button href={getEditorUrl(name)} size={size}>
     {isMember ? 'Edit Project' : 'View Source'}
-  </ButtonLink>
+  </Button>
```
sheridanvk: EditButton displays in two different sizes - regular on the Project page, small on the featured embeds on users/teams

storybook-config/config.js
```diff
@@ -1,7 +1,19 @@
...
+import { MemoryRouter } from 'react-router-dom';
+
+const enableLinks = (story) => <MemoryRouter>{story()}</MemoryRouter>
+
+addDecorator(
+  withInfo({
+    header: false, // Global configuration for the info addon across all of your stories.
+  }),
+  enableLinks
+);
```
sheridanvk: This adds the MemoryRouter to storybook so we can use Links without errors

styles/pages/project.styl
```diff
@@ -32,8 +32,8 @@
       display: block
   
   .buttons.space-between
-    .margin
+    a
```
sheridanvk: Small tweak as I removed the class on the button. This will be tidied up when the embed becomes a better component.

package.json
```diff
@@ -17,6 +17,7 @@
     "@sentry/browser": "^4.5.3",
     "@sentry/node": "^4.4.1",
     "@storybook/react": "^4.1.1",
+    "@storybook/addon-info": "^4.0.12",
```
sheridanvk: only this is an addition; just alphabetised the list otherwise

shrinkwrap.yaml
```diff
@@ -1,24 +1,25 @@
 dependencies:
-  '@babel/core': 7.3.4
-  '@babel/plugin-proposal-class-properties': 7.3.4
+  '@babel/core': 7.2.2
```
sarahzinger: it's very odd to me that we're downgrading babel and some others, I know that's not the intention of this pr and I have run into this myself. I guess if it works that's chill? Such oddness!

src/components/buttons/button.js
```diff
@@ -2,40 +2,67 @@ import React from 'react';
...
+      );
+    }
+    return (
+      <button onClick={onClick} className={className} disabled={disabled}>
+        {children}
+      </button>
+    );
+  };
+
+  return linkOrButton();
```
sarahzinger: just to make sure I'm following, a link in this instance looks like a button stylistically but it's click function is to take us to a url yeah? I guess I could see a circumstance where we have a link that is more of a traditional link with an underline and stuff, but that would probs live elsewhere yeah?

src/components/buttons/button.js
sarahzinger: this is neat I didn't know this was a thing :) I don't know that this is any "better" (and maybe you already thought about it but ruled it out!) but another alternative would be to do something like:
```
onClick: PropTypes.oneOfType([PropTypes.string, PropTypes.func]).isRequired
```
but how you have it seems legit to me!

src/components/buttons/button.js
sarahzinger: Again this seems pretty legit to me, I leave a lot of comments and feel free to like ignore any and all of them! One thing I think about out is like is do we want to characterize the types of things that are transparent rather than passing the style? like "isSecondaryDropdownBtn" or "secondary" or something? Idk, then again maybe transparency is like a special case, at least I could see an argument for that!

src/components/buttons/button.js
```diff
@@ -44,16 +71,20 @@ Button.propTypes = {
...
 
 Button.defaultProps = {
-  onClick: () => {},
+  onClick: null,
+  href: null,
   disabled: false,
   type: null,
   size: null,
   hover: false,
+  transparent: false,
```
sarahzinger: are most of our buttons not transparent? I kind of wonder if the default should be they are transparent and only special ones have a color to them? I think I'm probs missing the context

src/components/images/emoji.js
```diff
@@ -0,0 +1,60 @@
...
+  wave: 'https://cdn.glitch.com/55f8497b-3334-43ca-851e-6c9780082244%2Fwave.png?1502123444938',
+};
+
+/**
+ * Emoji Component
+ */
+
+const Emoji = ({ name }) => {
+  const classNameObj = { emoji: true };
+  classNameObj[name] = true;
```
sarahzinger: maybe just a one liner? `const className = cx({ emoji: true, [name]: true });` feel free to yell at me if this is nitpicky or like not how we do things!

src/components/images/emoji.js
sarahzinger: seems pretty clear to me!

src/components/images/emoji.js
```diff
@@ -0,0 +1,60 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames/bind';
+import styles from './emoji.styl';
+
+const cx = classNames.bind(styles);
+
+const emojiMap = {
```
sarahzinger: maybe we should capitalize this just to so we can shout `EMOJIS` every time we read the code

stories/index.js
```diff
@@ -1,28 +1,43 @@
...
+  .add('link (click to a different page)', () => <Button href="https://support.glitch.com">Support</Button>)
+  .add('with emoji', () => (
+    <Button onClick={helloAlert}>
+      <Emoji name="sunglasses" /> Show
+    </Button>
   ));
 
+storiesOf('Emoji', module)
+  .add('standard', () => <Emoji name="herb" />)
+  .add('sunglasses', () => <Emoji name="sunglasses" />);
```
sarahzinger: lol I wish all of our emojis had a sunglasses mode that would be pretty sweet...maybe there's a glitch app idea 😎

src/components/buttons/button.js
sheridanvk: yeah that's exactly right. and then yeah for links that look like links, you use the Link component (currently in presenters/includes/links)

src/components/images/emoji.js
```diff
@@ -0,0 +1,60 @@
...
+  ambulance: 'https://cdn.glitch.com/cc880f8d-a84f-4909-b676-497522a8c625%2Fambulance.png?1522348095546',
```
taravancil: I assume this is all copied over from a previous implementation, but I don't think we need the cache buster on these, right?

src/components/buttons/button.js
sheridanvk: you are right, yeah - honestly I just kind of tripped up over myself on this one with naming and trying to have this make sense. I thought maybe it would be weird behaviour to have the non-special buttons (like CTA) all be transparent by default, because that seemed unexpected to me, but it is in fact the majority case.

The context is mostly for popovers - PopoverButtons in a secondary section in the popover match the background (light grey) - e.g. [here](https://github.com/FogCreek/Glitch-Community/blob/master/src/presenters/pop-overs/user-options-pop.js#L127)

I thought about making it a type (like secondary) but then we get into issues where the button can be both secondary and tertiary :/ so it seemed like it should be an orthogonal property. after realising that transparency would work too, I introduced that instead.

I guess the other option would be to call it "matchBackground" and have that be default true, and just override it in a couple of places (specifically the upload image buttons on the user profile). Will work on that now

src/components/images/emoji.js
sheridanvk: oh this is actually something I did not realise was possible, thank you! much cleaner.

views/index.ejs
```diff
@@ -1,4 +1,4 @@
-<!DOCTYPE html>
+ <!DOCTYPE html>
```
taravancil: +1 whitespace

src/components/images/emoji.js
sheridanvk: I like it :D

stories/index.js
sheridanvk: :D :D I love it

views/index.ejs
sheridanvk: 👍

src/components/images/emoji.js
sheridanvk: good point, fixed!

shrinkwrap.yaml
sheridanvk: I wonder if this is just something that happens outside of our codebase with the npm modules? Like they update their dependency lists, so shrinkwrap changes?

sarahzinger: yeah something like that? what's weird is that it's downgrading not upgrading lol

src/components/buttons/button.js
```diff
@@ -2,40 +2,67 @@ import React from 'react';
...
+const Button = ({ onClick, href, disabled, type, size, matchBackground, hover, active, children }) => {
   const className = cx({
     btn: true,
     cta: type === 'cta',
-    small: size === 'small',
+    small: size === 'small' || type === 'dangerZone', // we want to demphasize dangerous actions, so we make them small
     tertiary: ['tertiary', 'dangerZone'].includes(type),
     dangerZone: type === 'dangerZone',
+    unstyled: type === 'dropDown',
+    hasEmoji: React.Children.toArray(children).some((child) => child.type && (child.type.name === 'Emoji' || 'Image')),
```
GregWeil: `(child.type.name === 'Emoji' || 'Image')` looks like a mistake, it'll only compare against 'Emoji'

I also don't think type.name will be set correctly once this is minified. I think if you import those two components and checked against them directly it would work.

src/components/buttons/button.js
GregWeil: Why does this define then call a function rather than just doing the conditional inline?

src/components/buttons/button.js
keithk: oh wow yeah I also had never seen this before!

src/components/buttons/button.js
keithk:
```
{ href ? <Link to={href} className={className}>{children}</Link> : <button onClick={onClick} className={className} disabled={disabled}>{children}</button> }
```

Yeah, I think it feels nicer to do all of this in JSX

src/components/buttons/button.js
sheridanvk: ahh yes good catch, thank you. I have been importing everything as Emoji so didn't catch this - I actually think I'll just remove Image (it was there because of the sunglasses image, but I just made this an emoji.

I did build this as a production version and it seems to still work even though it's minimised - perhaps because it uses the SourceMap or similar? https://lumpy-organ.glitch.me/@sheridan

src/components/buttons/button.js
sheridanvk: I was a bit worried the ternary operator looked a bit overwhelming when doing it all inline, hence breaking it out with an if statement in a function. @clottman has had opinions on this in the past - I'm not wedded to one way or the other, but it would be nice to have some guidelines of when ternary operators become too much?

GregWeil: You can use the `linkOrButton` contents as-is

src/components/buttons/button.js
GregWeil: Hmm, I don't see a `hasEmoji` class on any buttons there.

sheridanvk: gah maybe it was cached on my machine :/ thanks for checking. OK, looking into it!

src/components/buttons/button.js
sheridanvk: hah that is a good point - but I actually decided the ternary looked good, so I went with it :D thanks all!

server/api.js
```diff
@@ -95,6 +95,6 @@ module.exports = {
   getProject: (domain) => getFromCacheOrApi(domain, projectCache, getProjectFromApi),
   getTeam: (url) => getFromCacheOrApi(url, teamCache, getTeamFromApi),
   getUser: (login) => getFromCacheOrApi(login, userCache, getUserFromApi),
-  getCollection: url => getFromCacheOrApi(url, collectionCache, getCollectionFromApi),
+  getCollection: (url) => getFromCacheOrApi(url, collectionCache, getCollectionFromApi),
```
sheridanvk: not a part of this PR, but linter was complaining at me so I added these parens

src/components/buttons/button.js
sheridanvk: @sarahzinger were you suggesting that I get rid of href and make onClick either a string or a function? If so the reason I didn't was because I render different types of HTML element based on the presence of an onClick or an href - <button> and <Link> (i.e. <a>) respectively. I guess I could just check type in the Component as well to make that distinction? For now I'll leave as-is but if we think the component is too complicated and want to simplify like this then happy to adjust in future!

src/components/buttons/button.js
```diff
@@ -44,16 +62,20 @@ Button.propTypes = {
...
 
 Button.defaultProps = {
-  onClick: () => {},
+  onClick: null,
+  href: null,
   disabled: false,
   type: null,
   size: null,
   hover: false,
+  matchBackground: false,
```
sheridanvk: @sarahzinger FYI I went back to default not matching background, because it seemed weird that buttons like the CTA button which definitely don't match background (they're green) would have that set to true by default. We can adjust if this doesn't feel right though as we start creating more buttons.

src/components/buttons/button.js
sheridanvk: @GregWeil I believe this should be working now in a production build here. Would you mind taking a quick look and approving if all ok? http://lumpy-organ.glitch.me/@sheridan

src/components/buttons/button.js
sarahzinger: yeah totally fine to leave as is! I was just sharing that there is an existing pattern I think for saying `hey we need either a string or a function for this component to work` and then based on what you get back you render different things. But your work with these nifty functions make sense to me

src/components/buttons/button.js
```diff
@@ -2,40 +2,58 @@ import React from 'react';
...
   const className = cx({
     btn: true,
     cta: type === 'cta',
-    small: size === 'small',
+    small: size === 'small' || type === 'dangerZone', // we want to demphasize dangerous actions, so we make them small
     tertiary: ['tertiary', 'dangerZone'].includes(type),
     dangerZone: type === 'dangerZone',
+    unstyled: type === 'dropDown',
+    hasEmoji: React.Children.toArray(children).some(
+      (child) => child.type && child.type.name === React.createElement(Emoji, { name: 'herb' }).type.name,
```
sarahzinger: if we ever got rid of 'herb' would this work? Would it be weird to import the emoji map and pick whatever the first one is? maybe this is a problem that isn't a problem lol

src/components/buttons/button.js
sheridanvk: great and much appreciated!

package.json
sarahzinger: this was true (I remember seeing it as true!) but is it still? it now looks like you're adding back cache-loader and 'speed-measure-webpack-plugin' and I think those were removed

src/components/images/emoji.js
```diff
@@ -0,0 +1,59 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames/bind';
+import styles from './emoji.styl';
+
+const cx = classNames.bind(styles);
+
+const EMOJIS = {
```
sarahzinger: 🙌

src/components/buttons/button.js
sheridanvk: oof yeah interesting - I like your idea, it's way more robust. Will fix!

GregWeil: I think you can just do `child.type === Emoji`

package.json
sheridanvk: they're still in package.json but annoyingly hidden in this diff (expand the left-hand side!) - they were at the bottom of the list before. We removed some of the places where we're using the caching (specifically the stylus files I think) but we still use the caching for the JS files

sheridanvk: ohhh they're not in the diff because apparently they're in the package.json twice now :| fixing

src/components/buttons/button.js
sheridanvk: you are right! thank you :)

PR #346, opened 2019-03-20
~markdown-cards

taravancil:
This converts the page descriptions from raw markdown into "plain text", so that social cards don't display raw markdown

Before:
Screen Shot 2019-03-22 at 9 26 13 AM

After:
Screen Shot 2019-03-22 at 9 26 01 AM

TODOs:
- [x] decide whether this should happen within individual routes or inside of `render`
- [x] make sure there aren't other markdown fields


server/api.js
```diff
@@ -95,6 +95,6 @@ module.exports = {
   getProject: (domain) => getFromCacheOrApi(domain, projectCache, getProjectFromApi),
   getTeam: (url) => getFromCacheOrApi(url, teamCache, getTeamFromApi),
   getUser: (login) => getFromCacheOrApi(login, userCache, getUserFromApi),
-  getCollection: url => getFromCacheOrApi(url, collectionCache, getCollectionFromApi),
+  getCollection: (url) => getFromCacheOrApi(url, collectionCache, getCollectionFromApi),
```
taravancil: linter fix

shrinkwrap.yaml
```diff
@@ -7525,13 +7549,21 @@ packages:
     dev: false
     resolution:
       integrity: sha1-dFz0GlNZexj2iImFREBe+isG2Q8=
+  /lodash.assignin/4.2.0:
```
sarahzinger: are these changes intentional?

taravancil: are changes to shrinkwrap ever intentional? lolll no they're not 🙃

taravancil:
@sarahzinger that's an interesting idea and a good point! I think to make that work we'd need to store all markdown fields twice: once as raw markdown, once as text content. That seems reasonable to me, and probably ultimately the Right Thing, but my gut feeling says that maybe this is an ok solution for now and maybe forever?

PR #347, opened 2019-03-21
Bow kite: fix sentry bug

sarahzinger:
Saw this in sentry: https://sentry.io/organizations/fog-creek-software/issues/947251341/?project=1246508&query=is%3Aunresolved&statsPeriod=14d&utc=false

Interesting that it hasn't more obviously broken things haha, anyway my bad. 

https://bow-kite.glitch.me/

### DIFF
--- a/src/curated/featured.js
+++ b/src/curated/featured.js
@@ -14,18 +14,18 @@ Example:
 // make sure image urls use https
 export default [
   {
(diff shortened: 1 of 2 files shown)


ttseng:
oh weird, I didn't think it would work without it!  thanks for finding this.

PR #349, opened 2019-03-21
Slow flock 1: use api through hooks

modernserf:
- extract api base from `current-user`
- create `useAPI` hook and use everywhere api calls are made
- remove all api plumbing (heuristic: api should never be passed as a prop to a component in another file)
- update corresponding propTypes (if api is a prop, its always required)
- run it all through `prettier`


server/server.js
```diff
@@ -50,7 +50,7 @@ const proxy = require('./proxy');
 const proxied = proxy(app);
 
 const logger = require('./logger');
-app.use(logger)
+app.use(logger);
```
GregWeil: We run eslint against the server code, does it use fewer rules than the client (beyond just react stuff)?

modernserf: the server rules are much slimmer; also they don't seem to be enforced by CI.

PR #351, opened 2019-03-21
Glacier fiction: Componentized Segmented Buttons & Badges

ttseng:
Given an array of button contents, should generate a segmented button component.
Also added badge component 

Initial integration with search page filters:

image

- [x] Added segmented buttons to storybook
- [x] Moved current segmented button styling over to segmented-button.styl
- [x] Properly pass button children to segmented-button component


src/components/segmented-buttons/segmented-buttons.js
```diff
@@ -0,0 +1,53 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames/bind';
+import styles from './segmented-buttons.styl';
+import Button from '../buttons/button';
+
+const cx = classNames.bind(styles);
+
+const className = cx({
```
clottman: You don't need to do this since you're not doing anything conditional - get rid of line 7 (classnames.bind) and pass the className in as styles.segmentedButtons (you'll want to rename to camelcase so it works in jsx)

src/components/segmented-buttons/segmented-buttons.js
```diff
@@ -0,0 +1,53 @@
...
+    if (this.props.onClick) this.props.onClick(index);
+  }
+
+  render() {
+    const { buttons } = this.props;
+
+    return (
+      <div className={className}>
+        {buttons.map((button, index) => (
+          <Button key={button.content} size="small" type="tertiary" active={index === this.state.activeFilter} onClick={() => this.setFilter(index)}>
```
clottman: I don't think `button.content` is defined

src/presenters/pages/search.js
```diff
@@ -19,6 +19,19 @@ import ProjectsList from '../projects-list';
 import TeamItem from '../team-item';
 import UserItem from '../user-item';
 
+function generateFilterButtons(filters) {
+  const filterButtons = [];
+  filters.forEach((filter) => {
+    const button = {};
+    button.contents = capitalize(filter.name);
```
clottman: button.contents would end up like `Group1 <div>1</div>` which react doesn't know how to render. It will if you make the button.contents valid jsx though, like this: 

```
  filters.forEach((filter) => {
    const button = {};
    button.contents = <> { capitalize(filter.name) } {filter.hits > 0 && <div className="status-badge">{filter.hits}</div>} </>;
    filterButtons.push(button);
  });
```

clottman: That now yields a UI like this on https://battle-eucalyptus-1.glitch.me/search?q=jenn. 
image

src/components/segmented-buttons/segmented-buttons.js
```diff
@@ -0,0 +1,53 @@
...
+          </Button>
+        ))}
+      </div>
+    );
+  }
+}
+
+SegmentedButtons.propTypes = {
+  /** Expected button format: [content: ButtonName] */
+  buttons: PropTypes.array.isRequired,
```
clottman: You can define that in code instead of with a comment by making the proptype like this: 
```
   buttons: PropTypes.arrayOf(PropTypes.shape({
     content: PropTypes.string.isRequired,
   })).isRequired,
}
```

src/components/segmented-buttons/segmented-buttons.js
```diff
@@ -0,0 +1,53 @@
...
+  onClick: PropTypes.func,
+};
+
+SegmentedButtons.defaultProps = {
+  onClick: () => {},
```
clottman: All buttons should do something on click, right? So this should be required?

src/components/segmented-buttons/segmented-buttons.styl
```diff
@@ -0,0 +1,23 @@
...
+      border-top-left-radius: 0
+      border-bottom-left-radius: 0
+    &:not(:first-child):not(:last-child)
+      border-radius: 0px
+      border-left: 0
+    &.active
+      background: secondary-background-active
+    &.active:hover
+      background-color: secondary-background-hover
+    &:focus
```
clottman: Are we missing cases for `.active:focus` and `:hover` when not `.active`?

src/presenters/pages/search.js
```diff
@@ -178,23 +174,25 @@ class SearchResults extends React.Component {
...
-    const showUsers = ['all', 'users'].includes(activeFilter) && showResults(users);
-    const showProjects = ['all', 'projects'].includes(activeFilter) && showResults(projects);
-
+    const { teams, users, projects, activeFilterIndex } = this.state;
     const teamHits = teams ? teams.length : 0;
     const userHits = users ? users.length : 0;
     const projectHits = projects ? projects.length : 0;
+
     const filters = [
       { name: 'all', hits: null },
```
clottman: Could hits be teamHits + userHits + projectHits? (total hits)?

stories/index.js
```diff
@@ -75,3 +76,6 @@ storiesOf('Text', module)
 storiesOf('Markdown', module)
   .add('regular', () => <Markdown>Some __Markdown__</Markdown>)
   .add('truncated', () => <Markdown length={35}>35 characters of rendered __Markdown__ (and a little **more**)</Markdown>);
+
+storiesOf('Segmented-Buttons', module)
+  .add('regular', () => <SegmentedButtons buttons={[{contents: 1}, {contents: 2}, {contents: 3}]} ></SegmentedButtons>);
```
clottman: Another story that shows buttons where contents is jsx would probably be nice!

src/components/segmented-buttons/segmented-buttons.js
ttseng: ah ok!  I modified it so it's not switching between class names based on type.  

```
onst cx = classNames.bind(styles);

/**
 * Badge Component
 */
  const className = cx({
    badge: true,
    success: type === 'success',
    warning: type === 'warning',
    error: type === 'error',
  });
  return <div className={className}>{children}</div>;
};
```

Maybe there's some way for me to pass type directly though (without having to use the `classnames` package)?

src/components/segmented-buttons/segmented-buttons.js
ttseng: added content to proptypes

src/presenters/pages/search.js
ttseng: added badge component and imported styles from there

ttseng: also, thank you (!!) for explaining how passing nodes as props work!

src/components/segmented-buttons/segmented-buttons.js
ttseng: 👍

src/components/segmented-buttons/segmented-buttons.styl
ttseng: added!

src/presenters/pages/search.js
ttseng: I don't show the hits count for the all segmented button, so I never populate the hits count here.

stories/index.js
ttseng: good idea - added that along with badges

src/components/segmented-buttons/segmented-buttons.js
ttseng: sure - changed it to required

stories/index.js
```diff
@@ -75,3 +77,17 @@ storiesOf('Text', module)
 storiesOf('Markdown', module)
   .add('regular', () => <Markdown>Some __Markdown__</Markdown>)
   .add('truncated', () => <Markdown length={35}>35 characters of rendered __Markdown__ (and a little **more**)</Markdown>);
+
+storiesOf('Badge', module)
+  .add('regular', () => <Badge>Regular</Badge>)
+  .add('error', () => <Badge type="error">Error</Badge>);
```
oliviacpu: 💜 Could we also add `success` and `warning` examples?

ttseng: done 👍

src/components/segmented-buttons/segmented-buttons.js
clottman: You can do 
```
className={`badge ${styles[type]}`}
```

clottman: `styles` that you imported at the top is basically an object where each key is a class you defined in the css file. So in the snippet above, you're getting the `type` key from the styles object, which resolves to `styles.success` or `styles.warning` etc based on the value of `type`.

PR #353, opened 2019-03-22
Amazing slug: use title prop for helmet

modernserf:
image
[pictured: an amazing slug]

The infinite recursion errors we're getting are coming from react-helmet's use of [deep equality comparison on circular structures](https://github.com/nfl/react-helmet/issues/441). This is supposedly fixed in the current beta, but I think that using a title prop instead of children will also fix this (since its now comparing strings, not react elements).


jennschiffer:
wow i would love to be friends with that slug

keithk:
that is a great slug

clottman:
Oh wow I'm so excited for this error to get fixed! Good find, good slug.

PR #355, opened 2019-03-22
~tin-beach More From Section in collections

sarahzinger:
See demo here: https://tin-beach.glitch.me/@sarahzinger/my-collection

Screen Recording
http://recordit.co/pvKJkrFzne

Things To Do:
- [ ] ~turn stuff into it's own components~ (punting on this till storybook is fixed)
- [x] code clean up
- [x] qa/browser test

Things I've tested:
- [x] general look over in chrome
- [x] general look over in firefox
- [x] general look over in edge
- [x] general look over in safari
- [x] resizing screen
- [x] collections without projects
- [x] users with only one collection
- [x] a team collection page
- [x] a user collection page
- [x] collections with long descriptions (it shows ellipses)
- [x] collections with long title names (https://tin-beach.glitch.me/@sarahzinger/my-collection)
- [x] collections with only one project
- [x] collections with a black background
- [x] signed in
- [x] not signed in


ttseng:
Cool, exciting to see this coming along!

A few things:

1. it looks like the text size for descriptions & the project count might be slightly off - are you using the <Text> component here?
image

2. need to do the contrast isDarkColor checking to ensure the color of the description is readable:
image

3. the background of the cover-container should match the user's / team's cover image

4. a few changes from the spec (sorry - my bad!): let's change the h2 header to `More Collections by UserName  →` (from `More from UserName  →`)
5. also, let's add the arrow to the right of the project count (`2 projects →`)
6. we can hide the section if the user doesn't have any other collections, e.g. https://tin-beach.glitch.me/@scientiffic-test/hi

re: componetizing, I was imagining we might have a small-card and normal-card styling for each type of object, but I think we should probably catch up on creating other atoms first before doing it for this use case.

data-loader also makes sense to me.

src/presenters/collection-editor.js
```diff
@@ -66,7 +66,7 @@ class CollectionEditor extends React.Component {
     await this.props.api.patch(`collections/${this.state.id}/project/${projectId}`, { annotation: note });
   }
 
-  addNoteField(projectId) {
+  displayNewNote(projectId) {
```
sarahzinger: this has nothing to do with this pr, but I had intended to make these name changes before, and then forgot to push up my changes to github before merging 😭

src/presenters/collection-item.js
```diff
@@ -12,23 +12,21 @@ import { TeamAvatar, UserAvatar } from './includes/avatar';
...
-    <Text>
-      {'This collection is empty – add some projects '}
-      <span role="img" aria-label="">
-        ☝️
-      </span>
-    </Text>
-  ) : (
-    <Text>No projects to see in this collection just yet.</Text>
-  );
+const ProjectsPreview = ({ collection, isAuthorized }) => {
```
sarahzinger: the git diff for this component might be a little unpleasant to read since I moved stuff around a bit. Here's some of the changes

Previously:
- this component assumed projects had loaded and showed either an empty state (collections had no projects) or a filled state (collections had at least one project)
- if projects hadn't loaded yet it would render a loader (outside of this component on what was line 111)

Now:
- loader renders in this component
- moved the empty state variable down where it is used (below the filled state)
- I moved the loader in here as sometimes we won't want to show this preview at all

src/presenters/more-collections.js
```diff
@@ -0,0 +1,107 @@
...
+  let moreCollections;
+
+  // get all collections from the author
+  moreCollections = await getSingleItem(api, `v1/${authorType}s/${authorId}/collections`, 'items');
+
+  // filter out the current collection
+  moreCollections = moreCollections.filter((c) => c.id !== collection.id);
+
+  // get project details for each collection
+  moreCollections = await Promise.all(moreCollections.map(async (c) => {
```
sarahzinger: I'm not sure if we need to worry about this yet from a load perspective but I do wonder if this is a bad idea. 

We really only need to render 3 collections at a time, but we don't want to render projects that don't have projects in them yet. So my solution for this is to populate the projects first and then filter out the ones that don't have projects, and then randomly pick 3 to show. 

But one could see this potentially being super time consuming if a team (like glitch for example!) has hundreds or thousands of collections... yikes! 

I'm not sure I have a great solution for it. We could ask the backend team to return back a new property like `numOfProjects` when we request collections, or we could do some funky logic here being like pick 3 random collections, get their projects, if they don't have any projects, swap it for the next collection and so forth, until you run through the whole list. 

Maybe there's another idea here I'm not considering.

I do also think generally when we request collections we're requesting the projects that come with them almost always? Kind of makes me think it should always come back in full? idk.

styles/note.styl
```diff
@@ -37,6 +37,14 @@
 
   .dark
     color: white
-
     textarea
-      color: white
\ No newline at end of file
+      color: white
```
sarahzinger: unrelated to this pr, I just sort of squeezed these fixes in here. Notes where a little hard to read on dark backgrounds

Before
Screen Shot 2019-03-25 at 3 02 21 PM

After
Screen Shot 2019-03-25 at 3 02 45 PM
Screen Shot 2019-03-25 at 3 02 50 PM

src/presenters/more-collections.js
taravancil: Hm, I suspect this will be fine until it isn't lol, but I think something like this could work:

```js
let moreCollections = [];

while (moreCollections.length < 3) {
  // get 3 (or maybe 5?) collections from the author...idk if this is actually our limit syntax
  let fetchedCollections = await getSingleItem(api, `v1/${authorType}s/${authorId}/collections?limit=3`, 'items');

  // filter out the current collection
  fetchedCollections = fetchedCollections.filter((c) => c.id !== collection.id);

  // get project details for each collection
  fetchedCollections = await Promise.all(moreCollections.map(async (c) => {
    c.projects = await getSingleItem(api, `/v1/collections/by/id/projects?id=${c.id}`, 'items');
    return c;
  }));

  // filter out collections that don't have projects
  fetchedCollections = fetchedCollections.filter((c) => c.projects && c.projects.length > 0);

   // then append/push/whatever fetchedCollections to moreCollections
}

// take a sample size of 3 from moreCollections
```

taravancil: Not sure if we care about randomness with regard to which "More collections" we show, but if we do, this approach might not be the best choice

ttseng: ^ I agree with the approach of grabbing a limited set of collections - maybe 10 to be safe?

similarly, we can limit the number of projects you grab from the collection, e.g. `/v1/collections/by/id/projects?id=${c.id}?limit=1` since you don't care how many exactly it has - just that it has at least one.

src/presenters/more-collections.js
```diff
@@ -0,0 +1,105 @@
...
+    c.projects = await getSingleItem(api, `/v1/collections/by/id/projects?id=${c.id}`, 'items');
+    return c;
+  }));
+
+  // filter out collections that don't have projects
+  moreCollections = moreCollections.filter((c) => c.projects && c.projects.length > 0);
```
ttseng: I would create a new variable here to be clear that we're filtering on `moreCollections` - maybe `collectionsWithProjects`?

src/presenters/more-collections.js
```diff
@@ -0,0 +1,105 @@
...
+import { getProfileStyle as getTeamStyle } from '../models/team';
+
+import { DataLoader } from './includes/loader';
+import { CoverContainer } from './includes/profile';
+import { UserLink, TeamLink } from './includes/link';
+
+import CollectionItem from './collection-item';
+
+const loadMoreCollectionsFromAuthor = async ({ api, collection }) => {
+  const authorType = collection.teamId === -1 ? 'user' : 'team';
```
ttseng: rather than pluralize the authorType later (e.g. ``v1/${authorType}s/${authorId}`), maybe you can rename the variable to something like `const authorEndpoint`

src/presenters/more-collections.js
sarahzinger: @ttseng I think we need to know how many projects there are so we can display the little bit on the bottom yeah? 

Screen Shot 2019-03-26 at 11 46 09 AM

Also what would we think about me picking 10 random collections, loading the projects for all ten, filtering out the ones that don't have projects, and showing up to 3 random ones (and if a user weirdly has a ton of random collections without projects there is a possibility we might hide that section or show 1-3 collections in the more collections box) that we never accidentally loop through like thousands of empty collections lol.

ttseng: ^ that strategy makes sense to me

PR #356, opened 2019-03-22
Ruddy quasar: refactor collections-list.js using function components & hooks

modernserf:


src/presenters/collections-list.js
```diff
@@ -1,62 +1,49 @@
...
-  constructor(props) {
-    super(props);
+function CollectionsList({ collections: rawCollections, title, api, isAuthorized, maybeCurrentUser, maybeTeam }) {
+  const [deletedCollectionIds, setDeletedCollectionIds] = useState([]);
 
-    this.state = {
-      deletedCollectionIds: [],
-    };
-    this.deleteCollection = this.deleteCollection.bind(this);
+  function deleteCollection(id) {
```
modernserf: no `await`ing is necessary here, so not an async function

src/presenters/collections-list.js
```diff
@@ -81,66 +68,55 @@ const CreateFirstCollection = () => (
...
-  constructor(props) {
-    super(props);
-    this.state = {
-      shouldRedirect: false,
-      loading: false,
-      newCollectionUrl: '',
-    };
-    this.createCollectionOnClick = this.createCollectionOnClick.bind(this);
-  }
+const collectionStates = {
```
modernserf: make transitions an explicit state machine

modernserf:
parking this to fix conflict

PR #358, opened 2019-03-22
Grandiose pony: use hooks for notifications

modernserf:
`useNotifications` everywhere (except collections-list, to avoid conflict with https://github.com/FogCreek/Glitch-Community/pull/356)


src/presenters/includes/name-conflict.js
```diff
@@ -1,9 +1,9 @@
-import React from 'react';
+import React, { useEffect } from 'react';
```
oliviacpu: :100: Looking into hooks more now because of this. Thank you for requesting my review!

PR #361, opened 2019-03-22
Coordinated justice: hookify AddCollectionProject

modernserf:
- use hooks/function component for AddCollectionProject
- use new REST API for team project lookup
- move `useAsync` hook to state/api

### DIFF
--- a/src/presenters/includes/add-collection-project.js
+++ b/src/presenters/includes/add-collection-project.js
@@ -16,16 +16,18 @@ const useTeamProjects = createAPIHook(async (api, teamId) => {
 });
 
 function AddCollectionProject({ collection, addProjectToCollection }) {
-  const maybeTeamProjects = useTeamProjects(collection.teamId);
+  const teamResponse = useTeamProjects(collection.teamId);
   const { currentUser } = useCurrentUser();
 
   let initialProjects = [];
-  if (maybeTeamProjects) {
-    initialProjects = maybeTeamProjects;
-  } else if (currentUser) {
-    initialProjects = currentUser.projects;
+  if (teamResponse.status === 'ready') {
+    if (teamResponse.value) {
+      initialProjects = teamResponse.value;
+    } else {
+      initialProjects = currentUser.projects;
+    }
   }
-
+  
   return (
     <PopoverWithButton buttonClass="add-project" buttonText="Add Project">
       {({ togglePopover }) => (
--- a/src/presenters/includes/add-collection-project.js
+++ b/src/presenters/includes/add-collection-project.js
@@ -3,47 +3,42 @@ import PropTypes from 'prop-types';
 
 import AddCollectionProjectPop from '../pop-overs/add-collection-project-pop';
 import PopoverWithButton from '../pop-overs/popover-with-button';
-import { useAPI } from '../../state/api';
+import { useAsync } from '../../state/api';
+import { useCurrentUser } from '../../state/current-user';
 
-class AddCollectionProject extends React.Component {
-  constructor(props) {
-    super(props);
-    this.state = { projects: [] };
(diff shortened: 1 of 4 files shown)


src/presenters/includes/add-collection-project.js
```diff
@@ -3,47 +3,46 @@ import PropTypes from 'prop-types';
...
+  let initialProjects = [];
+  if (maybeTeamProjects) {
+    initialProjects = maybeTeamProjects;
+  } else if (currentUser) {
+    initialProjects = currentUser.projects;
   }
+
+  return (
+    <PopoverWithButton buttonClass="add-project" buttonText="Add Project">
+      {({ togglePopover }) => (
```
clottman: Is this consistent with how other PopoverWithButtons are written now?

src/state/api.js
```diff
@@ -22,3 +22,12 @@ export function useAPI() {
   const { persistentToken } = useCurrentUser();
   return getAPIForToken(persistentToken);
 }
+
+export const useAsync = (asyncFunction, ...args) => {
+  const api = useAPI();
+  const [result, setResult] = useState(null);
+  useEffect(() => {
+    asyncFunction(api, ...args).then(setResult);
+  }, args);
```
clottman: `args` won't include `asyncFunction` - do we need to add `asyncFunction` to the dependency array in case it's set by a prop that can change?

src/presenters/includes/add-collection-project.js
```diff
@@ -3,47 +3,46 @@ import PropTypes from 'prop-types';
...
+import { useAsync } from '../../state/api';
+import { useCurrentUser } from '../../state/current-user';
+import { getAllPages } from '../../../shared/api';
 
-class AddCollectionProject extends React.Component {
-  constructor(props) {
-    super(props);
-    this.state = { projects: [] };
+async function getTeamProjects(api, teamId) {
+  if (teamId > 0) {
```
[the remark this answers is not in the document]
modernserf: Not sure -- this is just copied from the original logic. @GregWeil any ideas?

src/state/api.js
modernserf: redesigned this to ensure that this will only be used with a static function, and fix a handful of potential async bugs.

src/presenters/includes/add-collection-project.js
modernserf: yup

src/presenters/includes/add-collection-project.js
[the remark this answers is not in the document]
clottman: Ah I see that now. I guess never mind?

src/state/api.js
clottman: Where does args get used other than on line 66? Is `args` supposed to be the list of dependencies for `handler`? If so then a more specific variable name might help.

src/state/api.js
```diff
@@ -22,3 +22,66 @@ export function useAPI() {
...
+
+  // ... render the team ...
+}
+
+*/
+
+// we don't want to set "stale" state, e.g. if the user clicks over to a different team's page
+// while the first team's data is still loading, we don't want to show the first team's data when it loads.
+// this should also avoid errors from setting state on an unmounted component.
+function useAsyncEffectState(initState, handler, args) {
```
clottman: minor issue - `initState` sounds like a verb to me (run `init` on `state`) but it should be a noun. Maybe change to `initialState` for clarity?

src/presenters/includes/add-collection-project.js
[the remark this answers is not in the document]
GregWeil: That's the collection's teamId, so it'll be -1 when the collection is owned by a user. (and the opposite for collection.userId)

PR #363, opened 2019-03-22
Available purple: hookify editable wrapping field

modernserf:
image

### DIFF
--- a/src/presenters/includes/editable-wrapping-field.js
+++ b/src/presenters/includes/editable-wrapping-field.js
@@ -11,7 +11,7 @@ function useUniqueId(prefix) {
   return ref.current;
 }
 
-function PureEditableWrappingField({ value, placeholder, update, autoFocus, error }) {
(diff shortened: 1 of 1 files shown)


src/presenters/includes/editable-wrapping-field.js
```diff
@@ -1,42 +1,41 @@
...
-export class PureEditableWrappingField extends React.Component {
-  constructor(props) {
-    super(props);
-    this.state = { id: uniqueId('editable-field-') };
-    this.onChange = this.onChange.bind(this);
-  }
+function useUniqueId(prefix) {
+  const ref = useRef(uniqueId(prefix));
+  return ref.current;
+}
```
sheridanvk: FWIW we also have this: https://github.com/FogCreek/Glitch-Community/blob/master/src/components/fields/use-unique-id.js

do we want to be consistent here and maybe expose whatever we use as a global hook? @GregWeil implemented this.

PR #364, opened 2019-03-25
Glacier-fiction: fixed bug with showing 0 project count in search filters

ttseng:
Testable by entering a search query that doesn't have a result in a particular category, e.g. 
https://glacier-fiction.glitch.me/search?q=scientiffic

On glacier-fiction:
image

On production (bug):

image

### DIFF
--- a/src/components/badges/badge.js
+++ b/src/components/badges/badge.js
@@ -1,6 +1,10 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import classNames from 'classnames/bind';
+
 import styles from './badge.styl';
+
+const cx = classNames.bind(styles);
 
 export const TYPES = ['success', 'warning', 'error'];
 
@@ -8,7 +12,7 @@ export const TYPES = ['success', 'warning', 'error'];
  * Badge Component
  */
 const Badge = ({ type, children }) => {
-  const className = styles.badge + (type ? styles.type : '');
+  const className = cx({ badge: true, type });
   return <div className={className}>{children}</div>;
 };
 
--- a/src/components/badges/badge.js
+++ b/src/components/badges/badge.js
@@ -1,22 +1,14 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import classNames from 'classnames/bind';
-
 import styles from './badge.styl';
 
 export const TYPES = ['success', 'warning', 'error'];
-const cx = classNames.bind(styles);
 
 /**
  * Badge Component
  */
 const Badge = ({ type, children }) => {
-  const className = cx({
-    badge: true,
-    success: type === 'success',
-    warning: type === 'warning',
-    error: type === 'error',
-  });
+  const className = styles.badge + (type ? styles.type : '');
   return <div className={className}>{children}</div>;
 };
 
--- a/src/presenters/pages/search.js
+++ b/src/presenters/pages/search.js
@@ -24,16 +24,17 @@ import UserItem from '../user-item';
 function generateFilterButtons(filters) {
   const filterButtons = [];
   filters.forEach((filter) => {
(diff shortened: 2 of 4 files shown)


src/components/badges/badge.js
```diff
@@ -1,22 +1,14 @@
...
  * Badge Component
  */
 const Badge = ({ type, children }) => {
-  const className = cx({
-    badge: true,
-    success: type === 'success',
-    warning: type === 'warning',
-    error: type === 'error',
-  });
+  const className = styles.badge + (type ? styles.type : '');
```
clottman: I think you want to use string templating here - this could result in, essentially, `badgesuccess` which I don't think is what you want.

You can also use the classnames library without the bind syntax - remove `/bind` from your import, and call it like `classNames(styles.badge, {styles.type: type})`. 

The classnames lib without bind just combines any standalone args you give it (badge) with anything in an object, conditionally based on its key. 

So it's equivalent to basically
```
cx({
badge: true,	
   'type': type
})
```

src/presenters/pages/search.js
```diff
@@ -24,15 +24,17 @@ import UserItem from '../user-item';
...
-    const button = {};
-    button.id = filter.name;
-    button.contents = (
-      <>
-        {capitalize(filter.name)}
-        {filter.hits && <Badge>{filter.hits}</Badge>}
-      </>
-    );
-    filterButtons.push(button);
+    if (filter.hits > 0 || filter.hits === null) {
```
clottman: When will filter.hits be null? Is that shorthand for the All filter?

src/components/badges/badge.js
ttseng: I see - I ran into syntax errors using `classNames(styles.badge, {styles.type: type})` so decided to just use `classnames/bind` and set ClassName with 
```
const cx = classNames.bind(styles);
const className = cx({ badge: true, type });
```

src/presenters/pages/search.js
ttseng: yeah - made it clearer by making that explicit. 
`filter.hits > 0 || filter.name === 'all';`

ttseng:
I had to fix a bug where the activeFilter was not being properly set.  

Basically, I was passing a subset of all buttons over to the `segmentedButtons` component based on which ones I actually wanted to display.  However, when setting the activeFilter back in `search.js`, it was indexing into an array containing all filters, regardless of whether or not there were hits.

So now I pass all the buttons over to `segmentedButtons` along with a prop `display` for whether or not the button should be displayed.  

I guess the way I could fix this is to have `setFilter` take an input of the button name, rather than its index, but it seems brittle to be passing a string around like that.

Any thoughts about this @clottman?

src/components/badges/badge.js
clottman: That was probably from the . in `{styles.type:`. `{[styles.type]: type}` would fix it. The way you chose works too though!

src/presenters/pages/search.js
clottman: Awesome.

clottman:
Having setFilter take a string name makes sense to me. Taking just an index is also brittle but in a sorta hidden way, because it relies on the order of the button elements never changing. This way you're being explicit, so even if we add a new button (for, I don't know, some new entity type that we can search by), the `setFilter(index)` calls won't have to change.

ttseng:
yeah, that makes sense to me - I updated it to use the button name instead.  I didn't like using display because I would imagine in other cases, we would want that to default to true, and I'm not sure if you can set defaultProps for nested shape props.

ttseng:
also, thanks @clottman for your feedback - I always learn so much from your reviews!

PR #366, opened 2019-03-25
Classy cobalt: use redux for current user

modernserf:
Notes:
- this diff is a lot easier to follow in side-by-side mode


src/state/current-user.js
```diff
@@ -86,195 +83,178 @@ function usersMatch(a, b) {
...
 
-  async login(user) {
-    this.props.setSharedUser(user);
-    this.props.setCachedUser(undefined);
-  }
+  const newCachedUser = await getCachedUser(nextState.sharedUser);
+  if (newCachedUser === 'error') {
+    // Looks like our sharedUser is bad, make sure it wasn't changed since we read it
+    // Anon users get their token and id deleted when they're merged into a user on sign in
+    // If it did change then quit out and let componentDidUpdate sort it out
```
modernserf: This part is what was confusing me the most. What is the desired behavior here?

src/state/current-user.js
```diff
@@ -86,195 +84,185 @@ function usersMatch(a, b) {
...
+      ...state,
+      sharedUser: null,
+      cachedUser: null,
+    }),
+  },
+});
+
+const selectCurrentUser = (state) => state.currentUser;
+
+const handleLoadRequest = afterReducer(matchTypes(actions.requestedLoad), async (store, action, prevState) => {
```
modernserf: There are essentially two schools of thought for working with side effects with redux.

The first approach is to put the side effects in the actions -- e.g. instead of dispatching a simple value, dispatch a promise, or a function that calls side effects. Middleware like redux-promise and redux-thunk intercept these non-value actions and run them.

The second approach is to put the side effects directly into the middleware -- the actions are always simple values, but the middleware listens to these actions, and can run side effects in response to them. This is the approach used by redux-saga and redux-loop; it's also how logging and metrics-reporting middleware work.

I prefer the second approach, but the tools associated with it (particularly redux-saga) tend to have a pretty steep learning curve. The approach I'm using here is less expressive, but much closer to the complexity of redux-thunk.

farskid: My opinion here probably doesn't matter at all but just dropped by to say that I'd also vote for the 2nd approach because, in case of refactoring and adopting paradigms such as FRP, this approach would need less rewriting. As proof, this is how Redux Observable works as an integration of RXJS with Redux.

src/state/current-user.js
GregWeil: The goal is to not call fixSharedUser if something else made a change to the shared user while we were loading. `this.props` would update while the async function runs, but that wouldn't happen here.

It's meant to prevent a race condition on login. On initial page load we fire off a load user request, but then the login page makes a post request that might delete/invalidate the user if they were anon and get merged. If the load finishes first that's fine, once the login finishes it'll set the user as expected. If the login finishes first and causes the load to fail, calling fix will overwrite the logged in user with undefined since the anon user doesn't exist.

src/state/current-user.js
```diff
@@ -86,195 +84,185 @@ function usersMatch(a, b) {
...
-  async logout() {
-    this.props.setSharedUser(undefined);
-    this.props.setCachedUser(undefined);
+const { slice, reducer, actions } = createSlice({
+  slice: 'currentUser',
+  initialState: {
+    // sharedUser syncs with the editor and is authoritative on id and persistentToken
+    sharedUser: readFromStorage('cachedUser') || null,
+    // cachedUser mirrors GET /users/{id} and is what we actually display
+    cachedUser: readFromStorage('community-cachedUser') || null,
```
GregWeil: This lost the storage event handling, so if you sign in on one tab you'll have to refresh the others to get them signed in.

src/state/current-user.js
```diff
@@ -86,195 +84,185 @@ function usersMatch(a, b) {
...
+    loaded: (state, { payload }) => ({
+      ...state,
+      ...payload,
+      loadStatus: 'ready',
+    }),
+    loggedIn: (state, { payload }) => ({
+      ...state,
+      sharedUser: payload,
+      cachedUser: undefined,
+    }),
```
GregWeil: We need an extra afterReducer to load the cachedUser after login. The login response doesn't have any teams/projects in it.

modernserf:
@GregWeil updated to handle storage events & loading cached user after login

src/state/current-user.js
GregWeil: I think this is still an issue, this async either needs to get cancelled when something else calls load or it needs to get the current state's sharedUser for the check rather than using what it is going to set the state to.

src/state/current-user.js
```diff
@@ -86,195 +84,204 @@ function usersMatch(a, b) {
...
+    // cachedUser mirrors GET /users/{id} and is what we actually display
+    cachedUser: readFromStorage('community-cachedUser') || null,
+    // states: init | loading | ready
+    loadStatus: 'init',
+  },
+  reducers: {
+    requestedLoad: (state) => ({
+      ...state,
+      loadStatus: 'loading',
+    }),
```
GregWeil: requested load used to get called every time sharedUser or cachedUser changed, but now it only gets called on page change. That means after you log in it doesn't have you teams/projects right away, and when you sign out it doesn't create an anon user.

modernserf: loading the user's teams/projects after login happens in `loadUserAfterLogin`; I'll add another handler for logout.

modernserf:
- added handler to get anon user after logout
- handle bail out of load with errors

src/state/current-user.js
```diff
@@ -86,195 +84,218 @@ function usersMatch(a, b) {
...
-      {(props) => <Context.Provider value={props}>{children}</Context.Provider>}
-    </CurrentUserManager>
-  );
-};
-CurrentUserProvider.propTypes = {
-  children: PropTypes.node.isRequired,
+const loadUserAfterLogin = afterReducer(matchTypes(actions.loggedIn, actions.storageChanged), async (store, action) => {
+  const sharedUser = action.payload;
+  const cachedUser = await getCachedUser(sharedUser);
+  store.dispatch(actions.updated({ cachedUser }));
```
GregWeil: `updated()` merges the argument with `state.cachedUser`, so this sets `state.cachedUser = { cachedUser: {} }`

src/state/current-user.js
```diff
@@ -86,195 +84,218 @@ function usersMatch(a, b) {
...
+});
+
+const loadAnonUserAfterLogout = afterReducer(matchTypes(actions.loggedOut), async (store) => {
+  const anonUser = await getAnonUser();
+  store.dispatch(actions.loadedAnonUserAfterLoggedOut(anonUser));
+});
```
GregWeil: I'm not sure exactly why, but this doesn't make it back to the state. On logout the POST /users/anon is made but the store never gets the anon user and localStorage.cachedUser stays null.

src/state/current-user.js
```diff
@@ -86,195 +84,218 @@ function usersMatch(a, b) {
...
+    // Anon users get their token and id deleted when they're merged into a user on sign in
+    // If it did change then quit out and let componentDidUpdate sort it out
 
-    // Check if we have to clear the cached user
-    if (!usersMatch(sharedUser, this.props.cachedUser)) {
-      this.props.setCachedUser(undefined);
+    if (usersMatch(prevUser.sharedUser, nextUser.sharedUser)) {
+      // The user wasn't changed, so we need to fix it
+      nextUser.sharedUser = await fixSharedUser(prevUser.sharedUser);
     }
```
GregWeil: I think this check has different behavior in its redux form. It was intended to guard against changes from outside of the load function, in case the user it failed to load had already been fixed by a login or something else.

The comment above mentioning componentDidUpdate is also outdated.

PR #367, opened 2019-03-25
Upbeat wildcat: fix image proptypes

modernserf:
image

fix the proptypes errors on Image components -- Image width/height are also being used with units, eg. `width="25%"`.


keithk:
ooo good catch 👏

PR #368, opened 2019-03-26
Nova chance: Add sign in using google auth behind a dev toggle

GregWeil:
This also changes how we pick which api to use. Before the `RUNNING_ON` env var determined the server api, while the current url determined the client api. Now the env variable will determine both.

@davidhodgson


src/presenters/includes/embed.js
```diff
@@ -17,18 +17,17 @@ class Embed extends React.Component {
   // TODO(sheridan) make this more robust in future
   browserSatisfiesRequirements = () => {
     try {
-      /* eslint-disable */
-      new URLSearchParams();
-      /* eslint-enable */
-
+      if (URLSearchParams.name !== 'URLSearchParams') {
```
clottman: Was this an old change that git got confused about, or have we not deployed it yet?

src/presenters/includes/embed.js
```diff
@@ -17,18 +17,17 @@ class Embed extends React.Component {
...
+        throw new Error('URLSearchParams is minified, so it must have been polyfilled');
```
clottman: How's the behavior of `throw`ing here different from just returning false?

src/presenters/includes/embed.js
GregWeil: It's an old change git got confused about, because I got confused about my remix names.

src/presenters/includes/embed.js
clottman: Ahh never mind, I see the catch now

GregWeil: It means all 'not supported' cases go through the catch, so we don't have to duplicate the log message.

PR #369, opened 2019-03-26
Pool musician: Image replacements

keithk:
Using the image component to update images around the entire community site.

Checklist

*QA*
- [x] Works in Edge / looks the same
- [x] Works in Firefox / looks the same
- [x] Works in Chrome / looks the same
- [x] Voiceover works the same

(Maybe I will set up screenshot tests after this..)


src/presenters/collection-item.js
```diff
@@ -11,6 +11,7 @@ import CollectionAvatar from './includes/collection-avatar';
 import { TeamAvatar, UserAvatar } from './includes/avatar';
 import { getAvatarUrl } from '../models/project';
 import { isDarkColor } from '../models/collection';
+import Image from '../components/image/image';
```
oliviacpu: I don't know if the aliases were done when this was started but we might be able to do `@components/image/image` now?

keithk: @oliviacpu I couldn't find documentation for the aliases and when I tried that it errored 💀

oliviacpu: My bad that is _almost_ ready!

keithk:
Merging master in, fixing a regression and adding aliases!

PR #371, opened 2019-03-26
Efficacious seer: remove state from Embed component

modernserf:
the check for native URLSearchParams doesn't need to be in component state, as it never changes and can be done well before the component renders.

### DIFF
--- a/src/presenters/includes/embed.js
+++ b/src/presenters/includes/embed.js
@@ -5,61 +5,46 @@ import PropTypes from 'prop-types';
 
 const telescopeImageUrl = 'https://cdn.glitch.com/7138972f-76e1-43f4-8ede-84c3cdd4b40a%2Ftelescope_404.svg?1543258683849';
 
-class Embed extends React.Component {
-  constructor(props) {
-    super(props);
-    this.state = {
-      browserSupported: this.browserSatisfiesRequirements(),
-    };
-  }
+// Babel does not transpile URLSearchParams, so using this as a compatibility check for showing embeds now.
+// TODO(sheridan) make this more robust in future
+const browserSatisfiesRequirements = (() => {
+  if ('URLSearchParams' in window && window.URLSearchParams.name === 'URLSearchParams') {
+    return true
+  } 
+  console.log(
+    "Sorry, you don't have the necessary JavaScript features to run Glitch code editors. Try applying your latest system updates, or try again with a different web browser.",
+    error,
+  );
+  return false;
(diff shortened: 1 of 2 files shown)


src/presenters/includes/embed.js
```diff
@@ -5,61 +5,50 @@ import PropTypes from 'prop-types';
...
-    } catch (error) {
-      console.log(
-        "Sorry, you don't have the necessary JavaScript features to run Glitch code editors. Try applying your latest system updates, or try again with a different web browser.",
-        error,
-      );
+// Babel does not transpile URLSearchParams, so using this as a compatibility check for showing embeds now.
+// TODO(sheridan) make this more robust in future
+const browserSatisfiesRequirements = (() => {
+  try {
+    if (URLSearchParams.name !== 'URLSearchParams') {
```
modernserf: Not relevant to this PR but -- why does it matter if its polyfilled?

sheridanvk: This check in the embed has a history so we shouldn't make these changes without coordinating - we use this check as a definitely too-brittle way of ensuring that the Google Crawler can crawl our site. 

glitch.com wasn't showing up in Google Search results because we weren't transpiling to ES2015, and we were instead checking for a bunch of JS functions before bootstrapping the app in client.js - if they didn't exist, we loaded the "you don't have the necessary JS" version of our site. This had no mention of the word Glitch on it, so to the crawler we didn't look like a relevant result for the word glitch.

So we added in the proper transpilation targets so it could render the majority of the site properly, but we still needed to make sure we didn't render embeds (as the user's app might not be transpiled and therefore would still break the whole site), so we implemented the check for embed support in this kind of janky way. Babel doesn't actually transpile/polyfill URLSearchParams out of the box, and we kept it this way so that this check will fail without ES6 support and we'll just not render the embeds instead of not rendering the whole site like we did before.

We had been talking about moving this check back into client.js and then passing some ES6 support Boolean value into the whole app, which could determine whether or not to show these embeds, but we never got around it. Could we maybe talk in person about it tomorrow? I think with the move to Redux this could be made much easier, as the whole app would have easier access to the flag determining whether or not to show embeds.

sheridanvk: One addition which I forgot that Greg made - we do actually polyfill URLSearchParams now, but the check for minification allows us to keep this compatibility check. So I'll just test that this works on Chromium 41 (will remix and build in production mode) and if it works then we're good to go.

PR #373, opened 2019-03-26
Storybook fixes and upgrades

sheridanvk:
This PR fixes the issues @sarahzinger was seeing on Production. Most issues were solved by upgrading to v5 of storybook per this recommendation: https://github.com/storybooks/storybook/issues/6248 

It also adds a much nicer UI :)


src/presenters/recent-projects.js
```diff
@@ -88,5 +87,4 @@ const RecentProjects = () => {
   );
 };
 
-
-export default RecentProjects;
\ No newline at end of file
+export default RecentProjects;
```
sheridanvk: this is just from running prettier due to a complaining build

stories/index.js
```diff
@@ -142,11 +142,24 @@ storiesOf('Badge', module)
...
+        {
+          contents: (
+            <>
+              <Badge type="error">Error</Badge> Badge
+            </>
+          ),
+        },
+      ]}
+    />
+  ));
```
sheridanvk: Again just prettier formatting (which I think adds too much space :( but not sure it's avoidable with their current configuration options

storybook-config/webpack.config.js
```diff
@@ -1,26 +1,33 @@
-const path = require("path");
+const path = require('path');
```
sheridanvk: Modified this config file to use the "full-control" rather than "extends" version of adding to the webpack configuration, because they're going to deprecate the latter

webpack.config.js
```diff
@@ -1,25 +1,25 @@
-const path = require("path");
-const LodashModuleReplacementPlugin = require("lodash-webpack-plugin");
-const TerserPlugin = require("terser-webpack-plugin");
-const MiniCssExtractPlugin = require("mini-css-extract-plugin");
-const AutoprefixerStylus = require("autoprefixer-stylus");
-const StatsPlugin = require("stats-webpack-plugin");
-const SpeedMeasurePlugin = require("speed-measure-webpack-plugin");
+const path = require('path');
```
sheridanvk: Ran prettier on this file

PR #374, opened 2019-03-26
Instinctive diagram: miscellaneous hookifications

modernserf:
a couple of simple hookifications all rolled up together so i'm not spamming the PRs


oliviacpu:
This all looks okay to me but I can't get this remix to load. I remixed your remix and was able to load that, but I want to find out what is up with this maybe?

oliviacpu:
The remix is loading now I'm going to test these changes.

oliviacpu:
Checked the secret page and the projects list expanding. :100:

PR #375, opened 2019-03-26
Olive team: Show Collection Notes on the Homepage

sarahzinger:
Gif: http://g.recordit.co/dKfDOQgF86.gif

Changes:
- if a collection's featured projects have at least one note it will show the note and hide the description
- I combined our two `loadCollection` functions, feel free to take an extra close eye there to make sure I'm not somehow breaking something by doing that. 
- I removed the projectLoader here as I don't think we need it...but again feel free to take a close 👁 to make sure that doesn't have a bad side effect I'm not thinking about
- some flexbox-y css stuff to make sure things are vertically aligned properly when we have notes

How to test:
- go here: https://glitch.com/edit/#!/olive-team?path=src/curated/collections.js
- add the collections you want to test with. For my testing I used these:
```
  { owner: 'glitch', name: 'probot-your-bff-on-github' },
  { owner: 'glitch', name: 'make-me-multilingual' },
  { owner: 'glitch', name: 'code-with-comics' },
  { owner: 'glitch', name: 'sketch-with-p-5' },
  { owner: 'clever-assembly-test', name: 'forest-melange' },
  { owner: 'empty-team', name: 'markdown-overflow-testing' }
```
but feel free to add your own... I have noticed we don't seem to support user collections (only team collections) on the homepage at this time which is interesting, but that's a current bug on prod I think, so I decided not to tackle that yet. 
- **make sure to reset the curated collections to whatever they were before**! Going to try my very hardest to make sure I don't put a test collection on the homepage when I deploy this lol

Things left to do before deploying:
- [x] qa logged in user
- [x] qa logged out user
- [x] qa firefox
- [x] qa safari
- [x] qa edge


src/presenters/project-item.js
```diff
@@ -8,7 +8,7 @@ import UsersList from './users-list';
 import Note from './note';
 import WrappingLink from './includes/wrapping-link';
 
-const ProjectItem = ({ project, collection, ...props }) => (
+const ProjectItem = ({ project, collection, hideProjectDescriptions = false, ...props }) => (
```
clottman: This should probably go in default props instead of in the arg destructure @oliviacpu

src/presenters/project-item.js
```diff
@@ -19,16 +19,20 @@ const ProjectItem = ({ project, collection, ...props }) => (
     <UsersList glitchTeam={project.showAsGlitchTeam} users={project.users} extraClass="single-line" teams={project.teams} />
     <ProjectOptionsPop project={project} {...props} />
     <WrappingLink href={getLink(project)} className="button-area">
-      <div className={['project', project.private ? 'private-project' : ''].join(' ')} data-track="project" data-track-label={project.domain}>
+      <div className={['project', project.private ? 'private-project' : '', hideProjectDescriptions ? 'hide-description' : ''].join(' ')} data-track="project" data-track-label={project.domain}>
```
clottman: This is getting messy - good opportunities to use the `classnames` lib. (Hit me up if you haven't used it before and need help - it'll be something like 

```
classnames('project', {['private-project']: project.private, ['hide-description']: hideProjectDescriptions})
```

clottman: @oliviacpu

src/presenters/projects-list.js
```diff
@@ -77,10 +77,10 @@ ExpandyProjects.defaultProps = {
   maxCollapsedProjects: 12,
 };
 
-export const ProjectsUL = ({ ...props }) => (
+export const ProjectsUL = ({ showProjectDescriptions = true, ...props }) => (
```
clottman: Default should go in defaultProps

GregWeil:
I'd rather consolidate collection loading on the homepage version, not the collection page code, since the collection page doesn't use the rest api.

I'm ok with leaving that for later if using the old api is part of getting it working again.

oliviacpu:
@GregWeil I put the collection loading back to how it was on the featured collections, since it was using the rest api. I did not change the collection page to use the rest api yet though.

oliviacpu:
I'm going to try to finish loading things from the rest api (user bubbles aren't showing) or I can just put it back to how it was for now.

src/curated/collections.js
```diff
@@ -2,9 +2,9 @@
 
 // What collections to show in the expanded area on the homepage
 export const featuredCollections = [
-  { owner: 'glitch', name: 'probot-your-bff-on-github' },
-  { owner: 'glitch', name: 'make-me-multilingual' },
-  { owner: 'glitch', name: 'code-with-comics' },
+  { owner: 'glitch', name: 'draw-with-music' },
```
clottman: maybe check with marketing that these are correct before you merge.

src/presenters/featured-collections.js
```diff
@@ -38,9 +40,9 @@ const CollectionWide = ({ collection }) => {
         </div>
       </header>
       <div className="collection-contents">
-        <ProjectsLoader projects={collection.projects}>{(projects) => <ProjectsUL projects={projects} />}</ProjectsLoader>
+        <ProjectsUL projects={featuredProjects} collection={collection} hideProjectDescriptions={featuredProjectsHaveAtLeastOneNote} />
         <CollectionLink collection={collection} className="collection-view-all">
-          View all <Pluralize count={collection.projectCount} singular="project" /> <span aria-hidden>→</span>
+          View all <Pluralize count={collection.projects.length} singular="project" /> <span aria-hidden>→</span>
```
clottman: Will projects contain all projects (even if there is a lot)?

oliviacpu: When we're loading we're also setting `collection.projects` to a sample of 3 from what we get from the API, I think. So we could remove that on line 24 and use `collection.projectCount` (which is the `projects.length` before it is sampled).

I'm fine doing those wherever. Since we're also checking for `featuredProjectsHaveAtLeastOneNote` around here though, I think I would prefer to keep `loadCollections` just loading things from the API and do things like `sampleSize` and keeping the original count around here as well.

oliviacpu: This will not show all projects as it is though, because of the `sampleSize` in `loadCollections` and it calling `projects.length` which should be "3" always.

oliviacpu: (That would load all the users for all the projects though, even if it is a lot? Nevermind what I said.)

oliviacpu:
💁 The featured collections component is using all v1 now. I am using some code from projects-loader to load the users, but I think that's something we can revisit later to see if it fits into one of our new patterns.

PR #376, opened 2019-03-26
Added a 1-second cache header to all community site routes

sheridanvk:
Hoping to reduce occurrences of this bug: https://glitch.manuscript.com/f/cases/edit/3328085/Improve-edge-caching-understanding-and-processes

This PR adds a 1s cache header to all non-static routes (it's placed after the static routes for this reason). This reduces the likelihood that a user will get a version of the page from cloudfront with old build URLs that don't exist on that app any more.

How to test:
- Go to https://wakeful-galleon.glitch.me/, and observe a 1s response header on the page
Screenshot 2019-03-26 at 6 16 54 PM
- Check that JS bundles still have the old caching on them (e.g. react.js)
Screenshot 2019-03-26 at 6 17 48 PM

### DIFF
--- a/server/routes.js
+++ b/server/routes.js
@@ -39,6 +39,7 @@ module.exports = function(external) {
   const ms = dayjs.convert(7, 'days', 'miliseconds');
   app.use(express.static('public', { index: false }));
   app.use(express.static('build', { index: false, maxAge: ms }));
+  app.use(express.static('views', { index: false, maxAge: 1000 }));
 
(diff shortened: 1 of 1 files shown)


oliviacpu:
👀 👍 - `This PR adds a 1s cache header to all non-static routes (it's placed after the static routes for this reason).`

Checked the headers and the routes and it looks :100:

sheridanvk:
thanks @oliviacpu! I'm going to roll this out then :)

PR #377, opened 2019-03-26
Circular snow: use algolia search (behind dev toggle)

modernserf:
First pass at algolia search


src/presenters/pages/search.js
```diff
@@ -104,132 +93,78 @@ const UserResults = ({ users }) => (
...
-    await this.props.api.patch(`collections/${collection.id}/add/${project.id}`);
-  }
-
-  render() {
-    const { teams, users, projects, activeFilter } = this.state;
-    const teamHits = teams ? teams.length : 0;
-    const userHits = users ? users.length : 0;
-    const projectHits = projects ? projects.length : 0;
+  const filters = [
+    { name: 'all', hits: searchResults.totalHits },
```
ttseng: we should leave totalHits out from the filter button, as it's redundant info, and the primary purpose of the badges is to help users quickly scan the relative difference in search results per type

PR #378, opened 2019-03-27
Ninth doom - fixed storybook for badges and segmented buttons

ttseng:
testable at https://ninth-doom.glitch.me/storybook

also added styling fixes to add-collection-project-pop 
image


src/components/badges/badge.js
```diff
@@ -12,7 +12,8 @@ export const TYPES = ['success', 'warning', 'error'];
  * Badge Component
  */
 const Badge = ({ type, children }) => {
-  const className = cx({ badge: true, type });
+  console.log(cx);
```
oliviacpu: ℹ️ `console.log` is still here, if this was for testing.

PR #379, opened 2019-03-27
Fix for the underscore in username collections bug in production

sheridanvk:
This PR fixes the production user issue by adopting the new v1 REST API, which handles the underscore in usernames correctly.
User who initially reported the issue: https://pickle-meteoroid.glitch.me/@shorty_jooce/neon-digest

~~One outstanding question: the ordering of projects seems to have been based on lastAccess, I believe, but this field no longer exists in the API (which makes sense, as that could be sensitive information). I ordered based on updatedAt, but I'm just confirming with the relevant stakeholders that this is ok (as it doesn't match what's on production right now).~~ confirmed with Tiff and Melissa that this is fine until projectOrder works.

Other tests done to confirm no regressions:
- Viewed collection including projects with a bunch of users: https://pickle-meteoroid.glitch.me/@taravancil/glitch
- Viewed collection with a lot of projects (20): https://pickle-meteoroid.glitch.me/@glitch/this-week-on-glitch-mar-27-2019 (this is a little slower now compared to production, but not sure if it can be avoided now that we have to get the users for every project)
- Viewed user collection: https://pickle-meteoroid.glitch.me/@shorty_jooce/neon-digest
- Viewed team collection: https://pickle-meteoroid.glitch.me/@glitch/this-week-on-glitch-mar-27-2019
- Created a new team collection
- Created a new user collection


src/presenters/pages/collection.js
```diff
@@ -196,21 +197,27 @@ CollectionPageContents.defaultProps = {
...
+      collection.user = await getSingleItem(api, `v1/users/by/id?id=${collection.user.id}`, collection.user.id);
+    } else {
+      collection.team = await getSingleItem(api, `v1/teams/by/id?id=${collection.team.id}`, collection.team.id);
+    }
+
+    // fetch users for each project
+    if (collectionProjects) {
+      const projectsWithUsers = await Promise.all(
+        collectionProjects.map(async (project) => {
+          project.users = await getAllPages(api, `v1/projects/by/id/users?id=${project.id}&orderKey=createdAt&orderDirection=ASC&limit=100`);
```
oliviacpu: This is neat! I'm going to try it in a couple of places maybe.

sheridanvk: thanks!

PR #381, opened 2019-03-27
Spurious attack: add collections to search (algolia only)

modernserf:
- handle collections in algolia and 'legacy' search APIs (but legacy collections search is disabled)
- show collections on the search pag
- refactor the handling of filter options
- fix some CSS on the search page

### DIFF
--- a/styles/collections.styl
+++ b/styles/collections.styl
@@ -31,179 +31,181 @@ collection-avatar-size = 55px
       display: inline-block
       vertical-align: bottom
 
-  .collections-container
-    margin: 10px 0px
-    padding: 0
-    list-style-type: none
-    // overflow: auto
-    overflow: hidden
-    
-    .collection-options-pop-btn
-      display: block
-      position: inherit
+.collections .collections-container,
+.search-results .collections-container
+  margin: 10px 0px
+  padding: 0
+  list-style-type: none
+  // overflow: auto
+  overflow: hidden
 
-    > li
-      float: left
-      display: block
-      margin-right: 10px
-      width: calc(33.3% - 10px)
-      position: relative
-      margin-bottom: 12px
-      @media (max-width: mediumViewport)
-        width: 100%
-      
-    .down-arrow
-      margin: 0
-    
-    .user-avatar
-      border-radius: 50%;
-      
-    .team-avatar
-      display: block
-      
-    .collection-curator a
-      display: flex
-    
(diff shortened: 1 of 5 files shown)


styles/collections.styl
```diff
@@ -31,179 +31,181 @@ collection-avatar-size = 55px
       display: inline-block
       vertical-align: bottom
 
-  .collections-container
-    margin: 10px 0px
-    padding: 0
-    list-style-type: none
-    // overflow: auto
-    overflow: hidden
+.collections .collections-container,
```
modernserf: this looks like a lot of changes, but its all just indentation

styles/pages/search.styl
```diff
@@ -15,23 +15,19 @@
...
   
-  .projects-container, .teams-container, .users-container
+  .projects-container, .teams-container, .users-container, .collections-container
     overflow: initial
     border-radius: 5px
     display: flex
     flex-direction: row
     flex-wrap: wrap
     align-items: flex-start
+    margin-left: -5px
```
modernserf: This fixes the oversized right margin

ttseng:
nice!  

for the css, if we use css-grid instead of flex, we can prevent the items in the last row from overstretching:

https://oasis-explanation.glitch.me/search?q=hello

here are the changes I made to `search.styl`
```
.projects-container, .teams-container, .users-container, .collections-container
    overflow: initial
    border-radius: 5px
    display: grid
    grid-template-columns: repeat(auto-fit, minmax(220px, 1fr))
    grid-gap: 5px
    margin-left: -5px
    margin-right: -5px
    >li
      margin: 0
      padding: 0 5px 10px
      width: auto
```

mainly:
```
.projects-container, .teams-container, .users-container, .collections-container
    display: grid
    grid-template-columns: repeat(auto-fit, minmax(220px, 1fr))
    grid-gap: 5px
```
and
``` 
>li 
      width: auto
```

I wasn't sure how to get a single search result not to stretch to the full width though:

image

ttseng:
Ultimately we'll want to have a more condensed collection card style for search results.  It's not necessary for this PR but just wanted to flag that as a TODO
image

ttseng:
I would say once CSS grid changes are made, this is good to go!

modernserf:
@ttseng  oh wow, that css is totally mangled, no idea how i ended up with that. check it now

PR #383, opened 2019-03-28
Stupendous fisher 1: Fix aria-labelledby errors in avatar tooltip container

meowwwls:
Currently where there is a user avatar with a tooltip, there is an invalid `aria-labelledby` value if a user's name has one or more spaces. The name is being used as the `id` value, which is then passed down to `aria-labelledby`. Having a space in that value is invalid. 

Accessibility errors during Lighthouse audit showing invalid `aria-labelledby` values

To solve this, I replaced spaces in the name at the site of the `id` prop with an underscore. I'm not sure if there is a better solution, but this seems to be the most simple and solved the issue.

### DIFF
--- a/src/components/tooltips/tooltip-container.js
+++ b/src/components/tooltips/tooltip-container.js
@@ -15,6 +15,9 @@ function TooltipContainer({ id, type, tooltip, target, align, persistent, childr
     'tooltip-container': true,
   });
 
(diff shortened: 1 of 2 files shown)


meowwwls:
@oliviacpu Thank you! I was having trouble locating where the changes should have been made. `tooltip-container` component makes perfect sense. 

I've gone ahead and made the suggested changes!

PR #384, opened 2019-03-28
Adding aliasing paths to the Community Site and Storybook

sheridanvk:
https://wax-snowstorm.glitch.me

This PR adds a bunch of absolute path aliases, updates all components to use the Components alias, modifies one path for each other alias to make sure it's working, and ensures Storybook still works with this change.

- Components', '/app/src/components'
- Utils', '/app/src/utils'
- Curated', '/app/src/curated'
- Models: '/app/src/models'
- Shared', '/app/shared'

Things to check:
[x] Storybook recompiled and works with no errors: https://wax-snowstorm.glitch.me/storybook
[x] Buttons and Images still rendering correctly (for Components alias)
[x] Avatars still rendering (for Model alias)
[x] Add Project on Collections page still working (for Shared alias)
[ ] As I'm now using an alias for the Utils directory, which contains Sentry, I will check that Sentry still receives errors when we get to community-staging (no errors are reported on other builds).


.babelrc
```diff
@@ -1,21 +1,20 @@
 {
-  "presets" : [
+  "presets": [
```
sheridanvk: just formatting

package.json
```diff
@@ -38,6 +39,7 @@
     "email-addresses": "^3.0.3",
     "eslint": "^5.15.1",
     "eslint-config-airbnb": "^17.1.0",
+    "eslint-import-resolver-alias": "^1.1.2",
```
sheridanvk: only this is new; the rest is alphabetising only

server/webpack.js
```diff
@@ -32,4 +32,4 @@ module.exports = function(app) {
   } else {
     app.use(webpackExpressMiddleware());
   }
-};
+};
```
sheridanvk: formatting only

stories/index.js
```diff
@@ -1,24 +1,27 @@
...
+import SegmentedButtons from 'Components/buttons/segmented-buttons';
 
 const helloAlert = () => {
   alert('hello');
 };
 
 const withState = (initState, Component) => {
-  const [state, setState] = useState(initState);
-  return <Component state={state} setState={setState} />;
+  const WrappedComponent = () => {
```
sheridanvk: This fix is from Justin so we render the components that receive state correctly (right now just Segmented Buttons)

sheridanvk:
> I haven't tried this yet but I am excited.
> 
> Should we add something about this to STYLE.md or something so people are aware they can use it?

Great call, added. Also I refactored this into a shared/aliases file so that it can also be used on the server side (and pulled in the module-alias package to enable that)

STYLE.md
```diff
@@ -29,6 +29,10 @@ We now have [Storybook for React](https://www.npmjs.com/package/@storybook/react
 1. Run `npm run storybook` from the terminal console to build the static storybook files.
 1. Go to `https://<remix-name>/storybook`. All the files are served there (from the build folder in the app).
 
+### Make use of the absolute path aliases
+
+We have aliases available on the server and client side, defined in the [shared aliases file](https://glitch.com/edit/#!/wax-snowstorm?path=shared/aliases.js:8:28). You can make use of those to remove the need for relative path imports.
```
oliviacpu: Should be `community` instead of `wax-snowstorm` in the link.

sheridanvk: ooh thanks, fixed!

PR #385, opened 2019-03-29
Nova chance: new embed support check that should be more reliable

GregWeil:
Use `new Function()` to test the features we care about without transpilation getting in the way.

Embeds now work on latest ff/chrome/edge, and get hidden in chrome 41

### DIFF
--- a/src/presenters/includes/embed.js
+++ b/src/presenters/includes/embed.js
@@ -9,8 +9,9 @@ const telescopeImageUrl = 'https://cdn.glitch.com/7138972f-76e1-43f4-8ede-84c3cd
 // TODO(sheridan) make this more robust in future
 const browserSatisfiesRequirements = (() => {
   try {
-    if (URLSearchParams.name !== 'URLSearchParams') {
-      throw new Error('URLSearchParams is minified, so it must have been polyfilled');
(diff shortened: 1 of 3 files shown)


src/presenters/includes/embed.js
```diff
@@ -9,8 +9,9 @@ const telescopeImageUrl = 'https://cdn.glitch.com/7138972f-76e1-43f4-8ede-84c3cd
 // TODO(sheridan) make this more robust in future
 const browserSatisfiesRequirements = (() => {
   try {
-    if (URLSearchParams.name !== 'URLSearchParams') {
-      throw new Error('URLSearchParams is minified, so it must have been polyfilled');
+    const test = new Function('return () => true'); // eslint-disable-line no-new-func
+    if (test()() !== true) {
+      throw new Error('Arrow functions are not supported, so the editor will not work');
```
sheridanvk: Not blocking rolling this out, but just curious: do we ever expect it to get to this line? At least on Chromium 41, the previous line makes it error so it drops into the catch before getting here.

GregWeil: MDN doesn't really specify what happens when the function body has a syntax error. I included that to handle if it crashes on call or does something silly like noop return undefined.

PR #387, opened 2019-03-29
Walnut cup: autocomplete in search (with algolia)

modernserf:
Notes:
- This isn't pixel-accurate to the mockup, but I figure refinements can be done in a subsequent PR.
- After #381 is merged, I will enable collection items.
- In the mockup, team and collection results show the avatars of their constituent members. We can't efficiently store that in Algolia, though. Can we limit those to the search results page?
- a few other important fields that we _can_ store still need to be backfilled, but that can be independent of this PR.


src/components/search/form.js
```diff
@@ -0,0 +1,47 @@
...
+
+  return (
+    <PopoverContainer>
+      {({ visible, setVisible }) => (
+        <form
+          action="/search"
+          method="get"
+          role="search"
+          onSubmit={onSubmit}
+          autoComplete={algoliaFlag ? 'off' : 'on'}
```
ttseng: is this prop needed?

modernserf: yeah -- this disables the browser's built-in autocomplete

clottman: Some people using iOS ask for auto-capitalize off as well - should we add that in here, too?

clottman:
Left some accessibility notes in Slack. :)

PR #388, opened 2019-03-29
Petal rock: resource manager [WIP]

modernserf:
image


keithk:
👀  I think a community team preso on how immer works would be cool 👏

PR #389, opened 2019-03-30
Arrow bloom: Add label text to TextInput component

meowwwls:
Currently, wherever the `TextInput` component is used, there is an accessibility failure as none of these inputs are provided with any sort of label text. They have placeholders only, which are meant to be hints and from an accessibility standpoint, do not serve as a replacement for labels.

To avoid changing the design, I added the label text but used a `visually-hidden` class to hide it while keeping it accessible to assistive technology. 

An alternative would be to use `aria-label` on the input, but I went with this method because the value of `aria-label` is not translated by services such as Google Translate:
http://www.heydonworks.com/article/aria-label-is-a-xenophobe

This change brings the Glitch landing page's Lighthouse Accessibility audit score from 73 to 82. 

I tried to find all instances where the `TextInput` component was used and pass in an applicable label. Apologies if I missed any or should use different text for any of them. If that's the case, I can make those changes.


src/components/inputs/text-input.js
```diff
@@ -12,7 +12,7 @@ const TYPES = ['email', 'password', 'search', 'text'];
 
 const InputPart = ({ children }) => <span className={styles.inputPart}>{children}</span>;
 
-const TextInput = ({ autoFocus, className, disabled, error, maxLength, name, onChange, opaque, placeholder, postfix, prefix, type, value }) => {
+const TextInput = ({ labelText, autoFocus, className, disabled, error, maxLength, name, onChange, opaque, placeholder, postfix, prefix, type, value }) => {
```
keithk: Small nit, can this prop be added in alphabetical order?

meowwwls: Done!

styles/globals.styl
```diff
@@ -99,6 +99,20 @@ hr
 
 .hidden
   display: none !important
+  
+// visually hides content while allowing it to be accessible to assistive technology (e.g., when you don't want an input label in the design even though all inputs must be labeled)  
+.visually-hidden
+	border: 0 !important
```
oliviacpu: I think we want these to be indented with two spaces for `styl` files?

```// visually hides content while allowing it to be accessible to assistive technology (e.g., when you don't want an input label in the design even though all inputs must be labeled)  
.visually-hidden
  border: 0 !important
  height: 1px !important
  width: 1px !important
  margin: -1px !important
  overflow: hidden !important
  padding: 0 !important
  position: absolute !important
  white-space: nowrap !important
  clip: rect(1px, 1px, 1px, 1px) !important
  clip-path: inset(50%) !important
  -webkit-clip-path: inset(50%) !important```

**Comment by oliviacpu** [comment]
I accidentally restarted the test runner while this was starting. It should run on the next set of changes with the right status links and everything.

File: src/components/inputs/text-input.js [hunk 3]
```diff
@@ -71,6 +73,7 @@ TextInput.defaultProps = {
   className: '',
   disabled: false,
   error: null,
+  labelText: undefined,
```
clottman: Since this is a `required` prop, it shouldn't have a `defaultProps` value. (So you can just remove this line)

src/presenters/header.js
```diff
@@ -19,6 +19,30 @@ const ResumeCoding = () => (
   </TrackedExternalLink>
 );
 
+function SearchForm({ defaultValue }) {
```
clottman: I believe this is a change that got stuck in here from when you remixed, even though it's no longer in the current version of the site. Perhaps you need to update your fork with what's on FogCreek:master to clear it out?

styles/globals.styl
```diff
@@ -99,6 +99,20 @@ hr
...
+	height: 1px !important
+	width: 1px !important
+	margin: -1px !important
```
clottman: Can you tell us a bit about what the `margin: -1px` is doing? It's not part of the recommended CSS from Web AIM for making invisible content that screenreaders can see. I'm a bit concerned that it might make the content fall under the '0 pixel sizing techniques' shortcomings and not be picked up by some screenreader/browser combinations.

clottman: Eep, I'm sorry! I see now that you're using the CSS clip technique, not the absolute positioning one, and this makes so much sense. My bad!

src/components/inputs/text-input.js
meowwwls: Sorry about that! Changes made.

styles/globals.styl
meowwwls: Sorry! Hopefully it's good to go now.

src/presenters/header.js
meowwwls: Hopefully everything's up-to-date now!

meowwwls: Apologies. Off to a rough start today. Now the `header.js` file should be up-to-date without conflicts.

clottman:
Thanks so much for improving this, @meowwwls! We have some more work to do on these inputs still - 

In particular: 
 - the equivalent fields on the user pages are not using the new OptimisiticTextInput components yet, so they don't get the improvements even though they look like they should behave the same way, and 
 - the 'prefix' prop sometimes contains text that doesn't make a lot of sense (like the @ before a team name - Voiceover for me is just inserting '@' in random places in the announcement and it's not very clear why). 

That work should be coming soon, along with a full accessibility audit and some dedicated time from our team to fix a lot of the bugs like this.

 In the mean time, I will go ahead and merge this and get it deployed ASAP! 

Thanks again, and happy Glitching!

clottman:
@meowwwls Just an update - your changes will be live as soon as the cache is busted! (so quite soon, or if you clear your cache, now!)

keithk:
Thanks so much for this contribution ❤️

PR #390, opened 2019-03-30
Superficial park: show all items of type when filter applied

modernserf:
Currently, we limit search results to 20 items per type.
Instead, when showing all search result types, limit 20 items per type, but show all results when showing only a single type.

---

Indicate that there are more results than are visible in the filter badge. (Using an imprecise number here signals to the user that they should refine their search, and allows us to make a smaller initial search request.)
Screen Shot 2019-03-30 at 9 54 42 AM

---

Add a "show all" button to the bottom of the results for a type that sets the search filter.
Screen Shot 2019-03-30 at 9 52 34 AM

### DIFF
--- a/src/presenters/pages/search.js
+++ b/src/presenters/pages/search.js
@@ -65,10 +65,8 @@ const groups = [
   { id: 'collection', label: 'Collections', ResultComponent: ({ result }) => <CollectionItem collection={result} /> },
 ];
 
-
-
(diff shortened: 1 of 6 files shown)


ttseng:
nice!  I'd suggest increase the bottom margin for each article on the search page from 1rem to 3rem `margin-bottom: 3rem` and placing the ShowAllButton within the article.

PR #394, opened 2019-04-01
Sunset stitch: Super User Banner and Toggle Button 

sarahzinger:
## Links
* https://sunset-stitch.glitch.me
* https://glitch.manuscript.com/f/cases/3326596/
* https://glitch.manuscript.com/f/cases/3327754/
* https://glitch.com/edit/#!/support-toggle

## GIF/Screenshots:
Screen Shot 2019-04-03 at 1 53 55 PM
Screen Shot 2019-04-03 at 1 54 04 PM

## Changes:
* normal users
    * should have no impact
* users invited to the supah secret project (ask cori)
    * all super users should see a big red banner with a button to turn off their super user status
    * to see the banner all the time users can run `window.localStorage.setItem('showSupportBanner', true)` and they will see the banner all the time until they clear their localStorage
    * when they have their super user feature turned off, they are able to hide the banner with a button click (which clears that part of localStorage)

## How To Test:
* log in as a user with the super_user feature, verify you see the banner
* or if you're not a super user such as myself you can mock it by adding the following code to line 74 on router.js:
`currentUser.features = [{ name: "super_user", expiresAt: "2019-11-01T14:11:01.233Z" }] `
* OR you can also put `window.localStorage.setItem("showSupportBanner", true)` in your console to see the banner

## Feedback I'm looking for:
* would love some super users to verify that they see the banner and that this meets their needs
* make sure this isn't too funky code wise. I did use inline styles just because it seemed easy/simple but happy to make this a real component or use css if we think that's best! 

- [x] fix CORS issue
- [x] get a super_user to qa it
- [x] verify I can't enable super user status if I'm not a member of the secret project
- [x] verify once more this doesn't do anything funky on other browsers for some reason.


src/presenters/pages/router.js
```diff
@@ -59,8 +60,40 @@ const PageChangeHandler = withRouter(({ location }) => {
   return null;
 });
 
+const SuperUserBanner = () => {
+  const { currentUser, persistentToken } = useCurrentUser();
```
modernserf: `useAPI` will add the current user's auth header automatically

src/presenters/pages/router.js
```diff
@@ -59,8 +60,40 @@ const PageChangeHandler = withRouter(({ location }) => {
...
+  if (currentUser && persistentToken) {
+    const supportAPI = axios.create({
```
modernserf: not sure, but you might be able to just do
```
const api  = useAPI()
```
and
```
api.post('https://support-toggle.glitch.me/support/enable')
```

since useAPI already handles the auth headers

src/presenters/pages/router.js
```diff
@@ -59,8 +60,40 @@ const PageChangeHandler = withRouter(({ location }) => {
...
+    });
+
+    const superUser = currentUser.features && currentUser.features.find((feature) => feature.name === 'super_user');
+    const toggleSuperUser = () => {
+      supportAPI.post(superUser ? 'disable' : 'enable');
+      window.location.reload();
+    };
+    const expirationDate = superUser && new Date(superUser.expiresAt).toUTCString();
+    const displayText = `SUPER USER MODE: ${superUser ? `ENABLED UNTIL: ${expirationDate}` : 'DISABLED'}`;
+    const showSupportBanner = window.localStorage.getItem('showSupportBanner');
```
modernserf: you might want to use the `useLocalStorage` hook for this

src/presenters/pages/router.js
sarahzinger: I think that useAPI has a base url that's different from this new endpoint, https://glitch.com/edit/#!/sunset-stitch?path=src/state/api.js:10:23

I thought about potentially editing the useAPI method to have 2 different base urls but that seemed overkill maybe?

src/presenters/pages/router.js
sarahzinger: yeah started down that path, then it seemed like more than we needed, but maybe better to use it haha. Let me try adding that back in :)

src/presenters/pages/router.js
modernserf: yeah -- i think if you pass a full URL (with protocol, domain etc) it will ignore the baseURL https://github.com/axios/axios#request-config

sarahzinger: Cool!

cori:
Sarah and I had a Slack convo about this iteration and here's what we came up with:
1. there will be a persistent toggle in `localStorage` that will force the banner to display whether you're `super_user` or not, with appropriate actions to enable / disable `super_user`. The banner will only be toggleable for members of the @Glitch team for now
    1. in this state the banner can be persistently hidden by closing it from the banner itself
1. the banner will show always when a `super_user` is logged in
    1. banner can then disable `super_user`. If you have the localStorage flag set then you'll see the _enable_ version of the banner unless you dismiss it.

keithk:
I wonder if moving it out of router makes sense, because it's now the only non-routing logic in that file, I'm not sure if that's the first place that I'd go to look for it.

sarahzinger:
@keithk would it be chill to put it in a separate file that's imported in Router? Or do you see this being added a different place in the tree maybe? I could import it in App maybe? https://github.com/FogCreek/Glitch-Community/blob/master/src/app.js

like maybe 
```
<CurrentUserProvider>
    <APIContextProvider>
        <SuperUserBanner>
            <Router><Router />
        </SuperUserBanner>
    </APIContextProvider>
</CurrentUserProvider>
```

Feel free if you think of a better place I keep moving it around lol. I want it to show up on a every page so I think it has to be pretty high up.

keithk:
@sarahzinger yeah, it makes more sense for me for it to be in app.js! Why do you have it wrapping <Router> in that case?

sarahzinger:
@keithk good point no need :P cool I'll go update :)

cori:
LGTM! Thanks!!

PR #396, opened 2019-04-02
projects navigation + filtering

taravancil:
## Links
* [projects-nav.glitch.me](https://projects-nav.glitch.me)
* [spec](https://www.notion.so/glitch/Show-more-filtering-updates-for-projects-listing-9cb5e72492bf4047ba02345776b91c7a)

## Screenshots
Screen Shot 2019-04-02 at 9 48 56 AM
Screen Shot 2019-04-02 at 9 58 03 AM
Screen Shot 2019-04-02 at 9 54 50 AM

## Changes:
- Pagination and filtering for projects lists (`src/projects-list.js`)
- When positioning the filter input, I realized we were using margins to create the projects grid, which means it didn't span the full width of the container. I updated `.collections-container` and `.projects-containers` to use CSS grid and consolidated their styles, since we had been duplicating them in a few places.

## How To Test:
**Filtering**:
- [ ] Filter a team's projects [@glitch](https://projects-nav.glitch.me/@glitch)
- [ ] Filter a user's projects [@potch](https://projects-nav.glitch.me/@potch)
- [ ] Does filtering feel fast? If not, how might we make it snappier?

**Pagination**
- [ ] Pagination controls visible when user has many projects? [@potch ](https://projects-nav.glitch.me/@potch)
- [ ] Pagination controls not visible when user has < 6 projects [@beyonce](https://projects-nav.glitch.me@beyonce)

**Project + collection containers** 
The CSS changes I made are used in a quite a few places, so we should QA them carefully, specifically with projects that have many users (lots of avatars tends to reveal grid issues)

- @modernserf can you double check the search results? [/search?q=wishlist](https://projects-nav.glitch.me/search?q=wishlist)
- @sarahzinger can you double check collections + narrative collections? [@glitch/marvelous-maps](https://projects-nav.glitch.me/@glitch/marvelous-maps)
- @ttseng do you want to take a glance at these changes to double check for regressions?
- Collections on the homepage
- Deleted projects grid
- Collections, pinned/recent projects on user + team pages

## Feedback I'm looking for:
- Am I debouncing correctly? Are there other filtering perf improvements that could be made?
- Am I using hooks effectively?

## Notes
The pagination and filtering controls are not a11y friendly, I'm sure of it. I'll be focusing on some other a11y fixes later today, so after this is merged + deployed, I will circle back and do some research on how to address that.


sarahzinger:
one random thought that you can feel free to disagree with, when I start typing the screen jumps around a bit as it resizes, maybe we should make it like a fixed height? idk maybe that would look awkward tho? http://g.recordit.co/p1FSMU8eb1.gif

src/presenters/projects-list.js
```diff
@@ -1,67 +1,136 @@
-import React, { useState } from 'react';
+import React, { useEffect, useState } from 'react';
 import PropTypes from 'prop-types';
-import Heading from 'Components/text/heading';
-import ExpanderContainer from 'Components/containers/expander';
+import { debounce } from 'lodash';
+
 import ProjectItem from './project-item';
+import Text from '../components/text/text';
```
sarahzinger: if you want I think you can do this from a non-relative path now which is neat, `'Components/buttons/button';`

src/presenters/projects-list.js
```diff
@@ -1,67 +1,136 @@
...
+  let { projects } = props;
+
+  function filterProjects() {
+    setIsDoneFiltering(false);
+
+    if (validFilter) {
+      setFilteredProjects(props.projects.filter((p) => p.domain.includes(filter) || p.description.toLowerCase().includes(filter)));
+      setIsDoneFiltering(true);
+    } else {
+      setFilteredProjects([]);
```
sarahzinger: should you setIsDoneFiltering(true) here as well?

modernserf:
search results are fine for projects, but teams & users are now the wrong size
Screen Shot 2019-04-02 at 11 47 29 AM

src/presenters/projects-list.js
```diff
@@ -1,67 +1,136 @@
...
-        </button>
-      }
-    >
-      <ProjectsUL {...props} projects={visibleProjects} />
-    </ExpanderContainer>
+    <>
+      <ProjectsUL {...props} projects={projects} />
+
+      {canPaginate ? (
+        <div className="view-controls">
```
sarahzinger: should all of this live in PaginationControls? should we pass the functions as props? feel free to disagree! just questions that are occurring to me while reading, not necessarily a "stance"/opinion.

src/presenters/projects-list.js
```diff
@@ -1,67 +1,136 @@
...
+    <div className="filter-results-placeholder">
+      <img alt="" src="https://cdn.glitch.com/c117d5df-3b8d-4389-9e6b-eb049bcefcd6%2Fcompass-not-found.svg?1554146070630" />
+      <Text>No projects found</Text>
+    </div>
+  ) : (
+    props.placeholder
+  );
+
+  return (
+    <article className={`projects ${extraClasses}`}>
```
sarahzinger: could this whole component live in /Components you think?

sarahzinger:
narrative collections look right to me, Checked on the home page, on a user page and on a team page!

src/presenters/projects-list.js
```diff
@@ -1,67 +1,136 @@
...
-      <ProjectsUL {...props} projects={visibleProjects} />
-    </ExpanderContainer>
+    <>
+      <ProjectsUL {...props} projects={projects} />
+
+      {canPaginate ? (
+        <div className="view-controls">
+          <PaginationControls />
+          <Button type="tertiary" onClick={() => setExpanded(true)}>
+            Show all<Badge>{hiddenProjects}</Badge>
```
ttseng: This should show `numProjects` rather than `'hiddenProjects` (should display the total count rather than what's not visible).

ttseng:
^ as justin mentioned, the search page was updated to show a grid of 4 columns for teams and users.  with the css changes here, it looks like it's returned to a 3 column grid.
For reference: https://glitch.com/search?q=hello

taravancil:
> when I start typing the screen jumps around a bit as it resizes, maybe we should make it like a fixed height?

@sarahzinger yeah I agree, I think that's the right thing to do. I'll make a case to add that as an enhancement, I think maybe it's ok to deploy this first version without it?

src/presenters/projects-list.js
taravancil: Yeah, now is probably a a better time than ever to move it there!

src/presenters/projects-list.js
taravancil: Oh hm, I see what you're saying. I think that would be ideal, but since the "Show all" button is also visible based on `canPaginate`, we'd need to put the button in `PaginationControls`, and that feels a bit weird to me?

src/presenters/projects-list.js
taravancil: We don't set it here, otherwise the "no projects found" placeholder shows up prematurely

src/presenters/projects-list.js
sarahzinger: @taravancil I think my comment is mostly stylistic so you can def ignore :) 

I was suggesting instead of doing 
```
      {canPaginate ? (
        <div className={styles.viewControls}>
          <PaginationControls />
          <Button type="tertiary" onClick={() => setExpanded(true)}>
            Show all<Badge>{numProjects}</Badge>
          </Button>
        </div>
      ) : null}
```
you could do 
```
{ canPaginate && (
  <PaginationControls 
       setExpanded={setExpanded}
       numProjects={numProjects}
        ...whatever other props ya need
  />
)}
```

taravancil:
Ok! I made a few more changes:

1. moved `projects-list.js` to `/components`
2. Fixed users and team grid layout on /search

Screen Shot 2019-04-02 at 2 54 22 PM

src/components/containers/projects-list.js
```diff
@@ -0,0 +1,159 @@
...
+
+  if (!expanded && canPaginate) {
+    const startIdx = (page - 1) * projectsPerPage;
+    projects = projects.slice(startIdx, startIdx + projectsPerPage);
+  }
+
+  const PaginationControls = () => (
+    <div className={styles.paginationControls}>
+      <Button aria-label="Previous" type="tertiary" disabled={page === 1} onClick={() => setPage(page - 1)}>
+        <img alt="" className={styles.paginationArrow} src="https://cdn.glitch.com/11efcb07-3386-43b6-bab0-b8dc7372cba8%2Fleft-arrow.svg?1553883919269" />
```
sarahzinger: should we use the Image component?

src/components/containers/projects-list.js
```diff
@@ -0,0 +1,159 @@
...
+  let projectsEl;
+  if (enablePagination) {
+    projectsEl = <PaginatedProjects {...props} projects={projects} />;
+  } else {
+    projectsEl = <ProjectsUL {...props} projects={projects} />;
+  }
+
+  const placeholderEl = filtering ? (
+    <div className={styles.filterResultsPlaceholder}>
+      <img alt="" src="https://cdn.glitch.com/c117d5df-3b8d-4389-9e6b-eb049bcefcd6%2Fcompass-not-found.svg?1554146070630" />
```
sarahzinger: again should we use image component?

sarahzinger:
@taravancil I think you've got a small style bug, on small screens on projects with lots of contributors the width of the projects is a bit too big: http://recordit.co/Sh4gdD2eZT

taravancil:
@sarahzinger the overflow issue is fixed. Thank you for catching that 🙏

sarahzinger:
sorry I know these are nitpicky, finding stuff slowly, I think you need a little of margin on the search page before the "Show all projects"
Screen Shot 2019-04-02 at 5 10 23 PM

taravancil:
Don't be sorry, these are good! Thanks, fixing

src/components/containers/projects-list.js
```diff
@@ -0,0 +1,160 @@
...
+  projects: PropTypes.array.isRequired,
+  projectsPerPage: PropTypes.number,
+};
+
+PaginatedProjects.defaultProps = {
+  projectsPerPage: 6,
+};
+
+export const ProjectsUL = ({ showProjectDescriptions, ...props }) => (
+  <ul className="projects-container">
```
sarahzinger: do you want to put the styles for this in projects-list.styl ?

ttseng:
taking a final look at the UI for this now!

src/components/containers/projects-list.js
```diff
@@ -0,0 +1,160 @@
...
+
+  const validFilter = filter.length > 1;
+
+  let { projects } = props;
+
+  function filterProjects() {
+    setIsDoneFiltering(false);
+
+    if (validFilter) {
+      setFilteredProjects(props.projects.filter((p) => p.domain.includes(filter) || p.description.toLowerCase().includes(filter)));
```
ttseng: since project domains are always lowercase, we should use `p.domain.includes(filter.toLowerCase())`.  Mobile keyboards default to sentence case for input, so testing the filter on mobile, I'm not able to find projects that don't match in case (for example, a query of `Stencilfy` doesn't return the project `stencilfy`)

styles/includes/projects-collections-teams-users-container.styl
```diff
@@ -0,0 +1,18 @@
...
+  padding: 0
+
+  @media (min-width: mediumViewport)
+    grid-template-columns: 1fr 1fr 1fr
+    > li
+      min-width: 33%  
+
+  > li
+    margin-top: 2px
+    overflow: hidden
```
ttseng: Right now, it looks like the project-options-pop is getting cut off in the grid of recent projects:

image

Setting overflow to visible here seems to fix the issue without any regressions, since we handle overflow for everything inside the li well.

Check out https://projects-nav.glitch.me/@scientiffic where I have a project with a very long name and description
image

ttseng:
Just a few bug fixes, but otherwise this is looking really great.  The search feels really snappy.  Excited to have this on the site! 🌟

taravancil:
> would be good to add some stories to storybook now that we're adding 2 Components could help illustrate when to use projectsList vs projectsUL

👍 added a case

src/components/containers/projects-list.js
taravancil: Good catch, thanks 🔠

taravancil:
Ok, the overflow issue is fixed!

**Overflowed menus are visible:**
Screen Shot 2019-04-03 at 10 11 11 AM

**Projects with lots of users don't overflow on small screens:**

Screen Shot 2019-04-03 at 10 11 28 AM

PR #397, opened 2019-04-02
Judicious pruner: componentize project items in search results

modernserf:
## Links
* Remix link: https://judicious-pruner.glitch.me
* Manuscript link: https://glitch.manuscript.com/f/cases/3328196/

### Search results (logged in)
Screen Shot 2019-04-02 at 2 29 36 PM

### Project with many users
Screen Shot 2019-04-02 at 2 29 12 PM

### Private project
(Note that private projects don't actually come up in search, but this component is intended to eventually work in other contexts)
Screen Shot 2019-04-02 at 2 38 52 PM

## Changes
- make new ProjectItem component (like ProjectItem presenter, but without the overrides for collection views)
- use new component in search results
- added gentle fade on overflowing UsersList

## How To Test:
Explore search results, e.g. https://judicious-pruner.glitch.me/search?q=community-staging, and verify that the results look "normal".

## Feedback I'm looking for:
- does this look & work correctly?
- does this have unforseen side effects on other project views?

## Next steps (for future PRs)
-  use in other project item contexts (integrate into or after user page search PR)
-  create project item component that includes notes, exclude descriptions, for use in collection pages
-  componentize its non-component dependencies (e.g. ProjectOptionsPop, UsersList, etc) and remove the `:global` css overrides
-  delete the old project item presenter


sheridanvk:
Would you consider adding storybook stories now so we can evaluate the components in isolation?

Mostly just wanted to say this is awesome and I can't wait for it :D I would love to be able to e.g. take all the project actions I am authorised to take from anywhere that a project is displayed (e.g. collections on the homepage), and it sounds like that will be possible once you port the collections to start using this component?

In terms of specific feedback, I noticed you no longer have the onhover effect of adding the dropshadow for the buttons. This was actually the plan (as we didn't want the inconsistency any more) but we do still need a hover state for the buttons when hovering over the entire project area (not the avatars). It should just match the standard button hover state - a change the background colour of the button. I had created a hover prop for the Button component for exactly this reason (so you could pass it down when the parent element was hovered), but let me know if you have a better idea of how to implement!

taravancil:
> I noticed you no longer have the onhover effect of adding the dropshadow for the buttons. This was actually the plan I didn't know we're getting rid of the drop shadow

I didn't know we were getting rid of that, I love the drop shadow 😭

sheridanvk:
> > I noticed you no longer have the onhover effect of adding the dropshadow for the buttons. This was actually the plan I didn't know we're getting rid of the drop shadow
> 
> I didn't know we were getting rid of that, I love the drop shadow 😭

spoke to Tara off thread about this - we want to keep hover states consistent, but I will chat to Tiff about making the hover state more obvious than the very subtle grey that we have right now! we can absolutely pick a brighter colour or similar :D

modernserf:
- added storybooks for project item & small collection item (plus a lot of plumbing to work with context)
- added a `decorative` flag to the Button componnt for buttons (such as these) that are actually inside larger buttons/links, and should have hover effects applied when the outer link is being hovered

src/components/buttons/button.js
```diff
@@ -26,30 +26,51 @@ const Button = ({ onClick, href, disabled, type, size, matchBackground, hover, c
...
+    return (
+      <Link to={href} className={className}>
+        {children}
+      </Link>
+    );
+  }
+
+  if (decorative) {
+    return (
+      <button className={className} disabled={disabled} tabIndex={-1}>
```
modernserf: decorative buttons don't "click", so they shouldn't be tab selectable

src/components/blocks/project-item.js
```diff
@@ -0,0 +1,64 @@
...
+const getLinkBodyStyles = (project) =>
+  classnames(styles.linkBody, {
+    [styles.private]: project.private,
+  });
+
+const ProjectItem = ({ project, projectOptions }) => (
+  <div className={styles.container}>
+    <header className={styles.header}>
+      <div className={styles.userListContainer}>
+        <StaticUsersList extraClass="single-line" glitchTeam={project.showAsGlitchTeam} users={project.users} teams={project.teams} />
```
modernserf: I feel like its better to have these user avatars _not_ be links, at least in search results -- it makes tabbing through results a lot cleaner if they aren't

stories/index.js
```diff
@@ -11,6 +11,15 @@ import Heading from 'Components/text/heading';
 import Markdown from 'Components/text/markdown';
 import Badge from 'Components/badges/badge';
 import SegmentedButtons from 'Components/buttons/segmented-buttons';
+import ProjectItem from 'Components/blocks/project-item';
+import SmallCollectionItem from 'Components/blocks/small-collection-item';
+import { Context as CurrentUserContext } from '../src/state/current-user';
+import { Context as APIContext } from '../src/state/api';
+
+// load global styles
+import '../build/styles.css';
```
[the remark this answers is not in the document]
sheridanvk: We explicitly avoided doing this because if you pull this in, it has all the site's legacy css in it, and it's therefore harder to catch whether you missed some styles in the process of converting something to a component. If you need global styles on a component-by-component basis for now (especially here because you haven't fully componentised this particular component yet), would it work to just pull in the pieces you need from the global stylesheets into the component's stylesheet?

Fonts should already work (there's a constants.styl in the components directory, and we import the webfont in the header of the storybook page.

src/components/images/avatar.js
```diff
@@ -0,0 +1,104 @@
+import React from 'react';
```
modernserf: mostly copied over directly from avatar presenter (which is still being used elsewhere)

src/components/images/image.js
```diff
@@ -14,34 +14,38 @@ import PropTypes from 'prop-types';
...
-          paddingBottom: `${backgroundRatio}%`,
-          backgroundRepeat: 'no-repeat',
-          width,
-          height,
-        }}
-      />
-    )}
-  </>
-);
+const handleDefaultSrc = (defaultSrc) => (event) => {
```
modernserf: Handle loading errors if src 404s

src/components/user/users-list.js
```diff
@@ -0,0 +1,104 @@
+import React from 'react';
```
modernserf: mostly copied over from presenter, but with `layout` instead of `extraClass`

stories/index.js
```diff
@@ -1,5 +1,6 @@
 import React, { useState } from 'react';
 import { storiesOf } from '@storybook/react';
+import 'Components/global.styl';
```
modernserf: just `*` and `body` styles here

src/components/buttons/button.styl
```diff
@@ -78,3 +78,8 @@
       background-color: #eaeaea;
       border-radius: 5px;
       border: none;
+
+// apply hover effects if decorative button is inside a link/button that's being hovered
+a:hover .btn.decorative,
+button:hover .btn.decorative
+  background-color: secondary-background
```
sheridanvk: this looks great. We might not need the hover prop now on the button, but I can remove that in a separate PR 👍

src/components/user/users-list.js
sheridanvk: we should break these out into a single component per file (we only want to export one component per file so they're easy to find), or we actually might be able to make the static version of the list just be a prop on UsersList? but no need to block this PR on that - I'll make a note on the components list.

PR #398, opened 2019-04-02
Garrulous literature: show top results on search page

modernserf:
## Links
* Remix link: https://garrulous-literature.glitch.me
* Manuscript link: https://glitch.manuscript.com/f/cases/3328158/
* Specs/Notion docs: https://www.notion.so/glitch/Search-244881f18e794507b67ef082a101cd84#0e64be9904114e8b9fd974e7def4ae4f (scroll to "Top Results")

## GIF/Screenshots
Screen Shot 2019-04-04 at 1 05 59 PM

## Changes
* add a "Top Results" section to the search results page
* show exact matches from search results in top results
* componentize & storybook SearchResults, Loading, NotFound

## How To Test
* search for users, projects, eg. https://garrulous-literature.glitch.me/search?q=react
* exact matches (for name, login, etc) should be in top results, in addition to the regular list
* if the top result is an exact match, and there are no further results in that type, then don't show the type section in All Results (see Edge Cases in Notion spec)

## Feedback I'm looking for
* does this render correctly?
* are there any edge cases I am missing?


src/components/search/form.js
```diff
@@ -5,6 +5,7 @@ import TextInput from '../inputs/text-input';
 import useDevToggle from '../../presenters/includes/dev-toggles';
 import PopoverContainer from '../../presenters/pop-overs/popover-container';
 import AutocompleteSearch from './autocomplete';
+import styles from './form.styl';
```
sheridanvk: I think this was in a previous PR which I didn't review, but per style.MD, please could you ensure that the names of the components you're exporting line up with the file name? Just makes it more predictable when developers are searching for these things in future (i,e, search-form.js/SearchForm, autocomplete-search.js/AutocompleteSearch)

ttseng:
Left a comment in the #pod-search channel about limiting the # of top results, particularly for name-matched collections https://fogcreek.slack.com/archives/CGUV81212/p1554407646013900

I don't think Top Results should have more than one from each type, so as an initial pass, we might want to use some tie-breaking criteria to determine which collection to show, for example (in order)

1. # of projects in the collection
2. maybe collections that have an edited description?  (does not contain "of projects that does") - let me know if you have another suggestion here

ttseng:
some un-evenness in the grid (try searching for react and open up the projects filter):
image

ttseng:
grid seems a little tight for collections - on production, we have a top margin on the card header (which contains the user list).  
can we add that to `.curator` in `collection-item.styl`?
image

ttseng:
awesome to see user avatars appearing for collections - should we do the same for projects?

modernserf:
@ttseng I think the grid uneveness should be fixed now, and i've limited the top results to one each of users, teams projects.

In a separate PR, I'll switch over to the new algolia indexes & get the project users rendering correctly.

PR #399, opened 2019-04-02
Clover socks: componentize user & team items

modernserf:
## Links
* Remix link https://clover-socks.glitch.me
* Manuscript link https://glitch.manuscript.com/f/cases/3328197/

## GIF/Screenshots
Screen Shot 2019-04-03 at 4 27 16 PM
Screen Shot 2019-04-03 at 4 26 14 PM

## Changes:
* port UserItem, TeamItem, Cover, Thanks to components
* make the whole component a real link instead of using WrappingLink (works better for keyboard navigation)
* create storybooks for UserItem, TeamItem (which both include Cover & Thanks)
* use UserItem, TeamItem on search page
* fix some styling issues with UserList

## How To Test:
* search page: https://clover-socks.glitch.me/search?q=computer

## Feedback I'm looking for:
* do team & user components look right on the search page?

## Next steps (for future PRs)
- use in other team/user item contexts
- convert remaining dependencies (link containers, verified badge)
- delete the old presenters


sarahzinger:
this is looking very rad, will this also support things like these containers ?
Screen Shot 2019-04-03 at 4 54 43 PM

modernserf:
@sarahzinger nope, this is just the user/team items when they appear in lists

src/components/user/user-item.js
```diff
@@ -0,0 +1,56 @@
...
+import { UserLink } from '../../presenters/includes/link';
+import styles from './user-item.styl';
+
+const ProfileAvatar = ({ user }) => <Image className={styles.avatar} src={getAvatarUrl(user)} defaultSrc={ANON_AVATAR_URL} alt="" />;
+
+const NameAndLogin = ({ user }) =>
+  user.name ? (
+    <>
+      <Button decorative>{user.name}</Button>
+      <p className={styles.login}>@{user.login}</p>
```
taravancil: On glitch.com this text is the tertiary color (#636363). Also maybe we want to use the `<Text>` component here?

sheridanvk:
Some weirdness with long usernames cutting off here. I can't find them on glitch.com to compare though, which is odd:
Screenshot 2019-04-03 at 5 14 41 PM

taravancil:
> Some weirdness with long usernames cutting off here

I'm seeing that on glitch.com too, so I don't think it's related to these changes. Sometimes the names are cut off, and other times they overflow

modernserf:
they clip on glitch.com, too, but they clip _differently_ -- i'll move the overflow: hidden to the outer component

src/components/blocks/cover.js
```diff
@@ -0,0 +1,19 @@
...
+};
+
+const Cover = ({ type, item, size }) => <div className={styles.cover} style={getProfileStyles[type]({ ...item, size })} />;
+
+Cover.propTypes = {
+  type: PropTypes.oneOf(['user', 'team']).isRequired,
+  size: PropTypes.oneOf(['medium', 'large']).isRequired,
+};
+
+export default Cover;
```
sheridanvk: given this is only used in the context of these result items, could we make the name a bit more specific (esp as cover images are used in other ways in other places)? something like SearchResultCover maybe?

sheridanvk: would also be great to get stories for this to make it faster for people to see whether this is what they want or not

sheridanvk:
we're pulling this style in from the global stylesheets still (links are rendering as blue in the component without it). Although maybe we should put it in Markdown?
Screenshot 2019-04-03 at 5 45 00 PM

src/components/blocks/thanks.js
```diff
@@ -0,0 +1,47 @@
+import React from 'react';
```
sheridanvk: Stories missing for this one too!

sheridanvk:
Ah I see in your comment that you've considered the stories for Thanks and Cover to be included within the User and TeamItems. We want stories for all components in src/components though - the idea being that if you need to create a new component that involves thanks, it's easy for you to visually confirm that this component is what you want. Storybook is definitely getting overwhelming though; I'll so some work to visually organise it.

PR #401, opened 2019-04-03
Bumpy child: Componentize Embed and Project Embed

sarahzinger:
## Links
* https://bumpy-child.glitch.me/
* related to: https://glitch.manuscript.com/f/cases/3328003/

## GIF/Screenshots:
Screen Shot 2019-04-04 at 1 38 51 PM
Screen Shot 2019-04-04 at 1 39 05 PM
Screen Shot 2019-04-04 at 1 39 19 PM

## Changes:
* Create 3 new components:
    * `Embed` - used both on it's own and as part of other components
    * `ProjectEmbed` - used on the project `~` page, shows the embed and also shows buttons below to allow users to remix/edit/add to collection/report depending on their relationship to the project
    * `FeaturedProject` - used on team and user page (and soon! collection page), all of the above plus a header and a btn to unfeature the project if the viewer owns the project.
* Replaces the embed with `ProjectEmbed`
* Replaces the user and team page embeds with `FeaturedProject`
* leaves other embeds alone for now, but should be able to swap other embeds (such as on the homepage with the new component)
* adds some new tracking info to our remix button. I noticed we were only sometimes tracking the origin of clicking on a remix button. I figured more information can probably only be better. 

This is some precursor clean up to create a featured project for a collection.

## How To Test:
* Look at a `FeaturedProject` on a user page such as https://bumpy-child.glitch.me/@sarahzinger
* Look at a `FeaturedProject` on a team page such as https://bumpy-child.glitch.me/@clever-assembly-test
* Look at a `ProjectEmbed` on a project page such as https://bumpy-child.glitch.me/~bumpy-child 
* See the stories for `Embed`, `ProjectEmbed`, and `FeaturedProject` in storybook https://bumpy-child.glitch.me/storybook (don't forget to hardrefresh)

## Feedback I'm looking for:
* am I missing anything in porting over these styles?

## Things I punted on for this pr:
* you'll notice some of the buttons in storybook such as the report button and the add to collection button and the dropdown on the top right side to "unfeature" a project are all stylistically and functionally weird. This is because those are not components yet. In the future those will get properly componentized and I believe they should all work as expected. 

- [ ] cross browser qa


src/components/images/emoji.js
```diff
@@ -43,8 +43,8 @@ const EMOJIS = {
  * Emoji Component
  */
 
-const Emoji = ({ name }) => {
-  const classNameObj = { emoji: true, [name]: true };
+const Emoji = ({ name, isInTitle }) => {
+  const classNameObj = { emoji: true, [name]: true, 'is-in-title': isInTitle };
```
modernserf: think this needs to be
```js
const classNameObj = { emoji: true, [name]: true, [styles.isInTitle]: isInTitle };
```
since the actual class name will be something like `emoji__isInTitle___1UliZ`

src/components/images/emoji.styl
```diff
@@ -19,4 +19,9 @@
     
   &.sunglasses
     width: auto
-    height: 15px
\ No newline at end of file
+    height: 15px
+
+.is-in-title
```
modernserf: this should be camelCased as `isInTitle`, so you can refer to it as `styles.isInTitle` instead of `styles['is-in-title']`

src/components/images/emoji.js
sheridanvk: this totally works, but I was thinking it would be nice to be able to take away this need to specify whether the emoji is in the title or not and just have it work whether the emoji is - and I came across this: https://stackoverflow.com/questions/18435777/adjust-image-size-based-on-the-font-size

Perhaps we can set emoji height/width to just be say 1.1 or 1.2em instead and remove all this situation-specific sizing?

src/components/project/embed.styl
```diff
@@ -0,0 +1,7 @@
+  .embed-container
```
modernserf: these should be camelCased as well

src/components/project/project-embed.styl
```diff
@@ -0,0 +1,15 @@
+.project-embed
```
modernserf: 🐫

src/components/images/emoji.js
sarahzinger: wait....then how does this work?

modernserf: oh! its being handled by `const cx = classNames.bind(styles)` on line 6

sarahzinger: cool, do we have a preference for camelcased css names? I don't really care either way, but if that's our preferences I can update/make it so going forward :)

modernserf: its in STYLE.md

sheridanvk: we do, good catch :D I totally missed those

sarahzinger: oh cool! I didn't know we had a style.md :) will read through thanks.

src/components/containers/projects-list.js
```diff
@@ -68,6 +68,7 @@ function ProjectsList({ title, placeholder, extraClasses, enableFiltering, enabl
             onChange={setFilter}
             opaque
             placeholder="find a project"
+            labelText="project search"
```
sarahzinger: this is unrelated to this pr, but I saw the error in the console and it seemed like an easy enough fix.

src/components/project/embed.js
```diff
@@ -0,0 +1,61 @@
...
+        allow="geolocation; microphone; camera; midi; encrypted-media"
+        height="100%"
+        width="100%"
+        border="0"
+        allowvr="yes"
+      />
+    ) : (
+      // Error message if JS not supported
+      // TODO(sheridan): Refactor this once we have a true error component
+      <div>
```
sarahzinger: I found this challenging to qa btw. I can see this error message in browserstack on chrome 41, but I can't see it with this remix. I'm not sure if that's because I'm introducing a bug or if it's because babel isn't babeling as much on remixes or something. 

I believe the only style changes I made here was changing the width from 50% to 35% as it was looking funky when I forced the error by making browserSatisfiesRequirements always return false. There were also some classnames in the old version, but I'm pretty sure they didn't really do much.

sarahzinger: AHH jk, Sheridan helped me qa this on chrome 41 looks pretty good :)

src/components/project/featured-project.styl
```diff
@@ -0,0 +1,9 @@
+.header 
+  img
+    margin-left: 5px
+    height: 1.1em
+    width: 1.1em
```
sheridanvk: I was wondering whether this should just be moved up to the emoji component, so all uses of Emoji in future take advantage of this dynamic sizing? The one-off styles can keep the fixed pixel heights perhaps?
New: 
Screenshot 2019-04-04 at 3 48 48 PM
Old: 
Screenshot 2019-04-04 at 3 47 03 PM

New:
Screenshot 2019-04-04 at 3 50 37 PM
Old:
Screenshot 2019-04-04 at 3 51 02 PM

This uses a size of 1.3em - affects the design a little but I actually like the additional space and large emoji (we could check with design on it). What do you think, does it make the component too restrictive like this?

sheridanvk:
I tried unfeaturing my project on (https://bumpy-child.glitch.me/@sheridan) and got this error. 

Screenshot 2019-04-04 at 3 57 46 PM

src/components/project/featured-project.styl
sheridanvk: and actually the margin could be em-based too - maybe 0.25em? that makes the space a bit smaller for the button:
Screenshot 2019-04-04 at 4 47 17 PM
Screenshot 2019-04-04 at 4 46 56 PM

sarahzinger: oh neat.... yeah I'll play around with it :)

sarahzinger: cool did what you said looks great :) Awesome suggestion!

sarahzinger:
I believe the unfeature button should be back in business @sheridanvk thanks for catching that! 🤦‍♀️

src/components/project/featured-project.styl
sheridanvk: yay no problem!

PR #402, opened 2019-04-03
Update Button to enable passing HTML attributes

taravancil:
## Links
* [copper-origami.glitch.me](https://copper-origami.glitch.me)
* [storybook](https://copper-origami.glitch.me/storybook/index.html?path=/story/button--with-aria-label)
* [manuscript case](https://glitch.manuscript.com/f/cases/3328205#BugEvent.21311456)

## Changes
This makes it possible to pass HTML attributes like `aria-label` into the Button component

## How To Test:
See the [example in storybook](https://copper-origami.glitch.me/storybook/index.html?path=/story/button--with-aria-label) or inspect the pagination buttons on the recent projects section of a [team page](https://copper-origami.glitch.me/@glitch).

## Feedback I'm looking for:
Are we ok with passing in HTML attributes via `props` or do we prefer to specifically define a `label` or `ariaLabel` prop?


modernserf:
Is there a category of button we'd want to _require_ an aria label for? i.e. labels are optional for text buttons but required for icon buttons?

taravancil:
@modernserf yeah that would be ideal

sheridanvk:
> Is there a category of button we'd want to _require_ an aria label for? i.e. labels are optional for text buttons but required for icon buttons?

That makes sense - we can achieve that via PropTypes by passing in a function similar to how we check that there's an onClick or href. So we can add an additional type="icon" and then yeah I think it would be best to define ariaLabel as a specific prop so that we can check it exists if type is icon.

sheridanvk:
In general we've proposed avoiding having components accept a props object, as that would allow for generic things to be passed in, defeating the goal of consistent components. As always though, happy to make exceptions for this where it makes sense, but here it sounds like we can avoid it with the above approach.

src/components/buttons/button.js
```diff
@@ -69,6 +70,20 @@ Button.propTypes = {
     }
     return null;
   },
+  /** it's an icon button (no text label) */
```
modernserf: If these go together it might make more sense to make a separate IconButton component

sarahzinger:
@taravancil this has been open a while, should we just close it maybe?

taravancil:
@sarahzinger yep, I don't even remember what this was, so odds are good I'm not gonna finish it!

PR #404, opened 2019-04-04
Iris okapi: Update email sign in to have consistent error styles

oliviacpu:
## Links
* Remix link: https://iris-okapi.glitch.me/
* Manuscript link: https://glitch.manuscript.com/f/cases/3327691/
* Specs/Notion docs: https://www.notion.so/glitch/87bda09f85184c1d81e1659aa62e4392?v=78db6cf11f8a43859bd17bc800cebde0&p=40dcbd076b574f1695d94094742608e5

## GIF/Screenshots:
Screenshot 2019-04-04 at 10 49 57 AM

## Changes:
* Previously the error would show after submitting that an invalid email was submitted. This checks for a valid email first and then displays an error below the text input.

## How To Test:
* Try to enter an invalid email and see what happens on ~community and this remix.

## Feedback I'm looking for:
I think this is just the quickest way to make the styles consistent. There are probably notes we could make as we componentize more things.

### DIFF
--- a/src/presenters/pop-overs/sign-in-pop.js
+++ b/src/presenters/pop-overs/sign-in-pop.js
@@ -106,7 +106,14 @@ class EmailHandler extends React.Component {
             <section className="pop-over-actions first-section">
               {!this.state.done && (
                 <form onSubmit={this.onSubmit} style={{ marginBottom: 0 }}>
-                  <input value={this.state.email} onChange={this.onChange} className="pop-over-input" type="email" placeholder="new@user.com" />
+                  <TextInput
+                    type="email"
+                    labelText="new@user.com"
+                    value={this.state.email}
+                    onChange={this.onChange}
+                    placeholder="new@user.com"
+                    error={this.state.errorMsg}
+                  />
                   <button type="submit" style={{ marginTop: 10 }} className="button-small button-link" disabled={!isEnabled}>
                     Send Link
                   </button>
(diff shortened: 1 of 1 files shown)


src/presenters/pop-overs/sign-in-pop.js
```diff
@@ -55,8 +57,9 @@ class EmailHandler extends React.Component {
     this.onSubmit = this.onSubmit.bind(this);
   }
 
-  onChange(e) {
-    this.setState({ email: e.target.value });
+  onChange(email) {
+    const isValidEmail = parseOneAddress(email) !== null;
+    this.setState({ email, errorMsg: isValidEmail ? undefined : 'Enter an email address.' });
```
keithk: Should this have a separate error message for valid email, or do you think this is enough for someone to understand?

oliviacpu: It will say "valid" after like "olivia@g" so I think that would be more confusing than handling just the cases it was going to 400 in.

I also left the case for handling if we get a 400 back just in case.

ttseng:
Can we add a debounce of ~500 ms so that it doesn't show the error message until the user pauses typing?  Have the error messaging appear immediately is a bit startling.

oliviacpu:
@ttseng This has the debounce now if you want to see how it looks on the remix.

ttseng:
debounce looks good!  I would just tweak the copy a bit so that it's more indicative of why the input is throwing an error (the email is invalid): `Enter a valid email address` 
No period needed either. (:

PR #405, opened 2019-04-04
Stingy print: store active search filter in query params

modernserf:
## Links
* Remix link: https://stingy-print.glitch.me/search
* Manuscript link: https://glitch.manuscript.com/f/cases/3328201/

## Changes:
* activeFilter is now a query param, so you can link to a category of search results, and the search filter is persistent across page refreshes

## How To Test:
* search: https://stingy-print.glitch.me/search?q=community
* apply a filter and refresh the page
* it should show the filtered search results

## Feedback I'm looking for:
- Is this useful / worthwhile?
- Does this impact loading performance?

### DIFF
--- a/src/presenters/pages/router.js
+++ b/src/presenters/pages/router.js
@@ -125,9 +125,10 @@ const Router = () => (
       <Route
         path="/search"
         exact
-        render={({ location }) => (
-          <SearchPage key={location.key} query={parse(location.search, 'q')} activeFilter={parse(location.search, 'activeFilter')} />
-        )}
+        render={({ location }) => {
+          const query = parse(location.search, 'q');
+          return <SearchPage key={query} query={query} activeFilter={parse(location.search, 'activeFilter')} />;
+        }}
       />
 
       {categories.map((category) => (
--- a/src/presenters/pages/router.js
+++ b/src/presenters/pages/router.js
@@ -122,7 +122,8 @@ const Router = () => (
         <Route key={name} path={`/${name}`} exact render={({ location }) => <TeamPage key={location.key} id={rootTeams[name]} name={name} />} />
       ))}
 
-      <Route path="/search" exact render={({ location }) => <SearchPage key={location.key} query={parse(location.search, 'q')} />} />
+      <Route path="/search" exact render={({ location }) => 
+        <SearchPage key={location.key} query={parse(location.search, 'q')} activeFilter={parse(location.search, 'searchType')} />} />
 
       {categories.map((category) => (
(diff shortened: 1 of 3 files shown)


src/presenters/pages/router.js
```diff
@@ -122,7 +122,14 @@ const Router = () => (
         <Route key={name} path={`/${name}`} exact render={({ location }) => <TeamPage key={location.key} id={rootTeams[name]} name={name} />} />
       ))}
 
-      <Route path="/search" exact render={({ location }) => <SearchPage key={location.key} query={parse(location.search, 'q')} />} />
+      <Route
+        path="/search"
+        exact
+        render={({ location }) => {
+          const query = parse(location.search, 'q');
+          return <SearchPage key={query} query={query} activeFilter={parse(location.search, 'activeFilter')} />;
```
modernserf: using `query` as the key, so this doesn't reload the whole view when you change the filter.

src/presenters/pages/search.js
```diff
@@ -75,8 +76,15 @@ const MAX_UNFILTERED_RESULTS = 20;
...
 const groupIsInFilter = (id, activeFilter) => activeFilter === 'all' || activeFilter === id;
 
-function SearchResults({ query, searchResults }) {
-  const [activeFilter, setActiveFilter] = useState('all');
+const SearchResults = withRouter(({ query, searchResults, activeFilter, history }) => {
+  const setActiveFilter = (filter) => {
+    history.push(`/search?q=${query}&activeFilter=${filter}`);
+  };
+
+  if (!searchResults[activeFilter] || searchResults[activeFilter].length <= 0) {
```
modernserf: if the filter is invalid, or there's no results for a filter, just show all results

src/presenters/pages/search.js
```diff
@@ -75,8 +76,15 @@ const MAX_UNFILTERED_RESULTS = 20;
 
 const groupIsInFilter = (id, activeFilter) => activeFilter === 'all' || activeFilter === id;
 
-function SearchResults({ query, searchResults }) {
-  const [activeFilter, setActiveFilter] = useState('all');
+const SearchResults = withRouter(({ query, searchResults, activeFilter, history }) => {
+  const setActiveFilter = (filter) => {
```
modernserf: :neckbeard: technically :neckbeard:  we could make the buttons work as actual links, but this is a substantially less invasive change

taravancil: yeah this seems fine for now and probably forever

PR #407, opened 2019-04-05
Capricious kayak - small UI fixes

ttseng:
## Links
* Remix link: https://capricious-kayak.glitch.me
* Manuscript link: https://glitch.manuscript.com/f/cases/3328137/Bug-Team-Email-invite-blurb-formatting-off-separating-c-and-an-on-to-separate-lines#BugEvent.21311718

## GIF/Screenshots:
(before on left, after on right)
image

## Changes:
* Fixed word break in pop-overs
* Fixed avatar styling in teams project filter pop

## How To Test:
* try filtering on a project on a team page
* try adding a whitelist on team pages


ttseng:
also fixed add-project-to-collection pop buttonwrap styling:
image

src/presenters/includes/user-result-item.js
```diff
@@ -65,7 +65,7 @@ InviteByEmail.propTypes = {
 export const WhitelistEmailDomain = ({ domain, onClick }) => (
   <button onClick={onClick} className="button-unstyled result">
     <WhitelistedDomainIcon domain={domain} />
-    <div className="result-name">Allow anyone with an @{domain} email to join</div>
+    <div>Allow anyone with an @{domain} email to join</div>
```
clottman: is `.result-name` a class that's no longer used and can be removed?

styles/collections.styl
```diff
@@ -120,18 +121,13 @@ collection-avatar-size = 55px
     display: inline-block
 
   .collection-name-description
-    width: 100%
-    margin-left: 12px
-
+    width: "calc(100% - %s - %s - 12px)" % (collection-avatar-size collection-padding)
```
clottman: Wanna throw in a comment explaining what the factors in the calc() are doing? I'm not sure where %s is defined.

src/presenters/includes/user-result-item.js
ttseng: it's used elsewhere, but it's unnecessary here and adds inconsistent styling.

styles/collections.styl
ttseng: sure - %s is just used to reference variables in a calc (in this case, placeholders for collection-avatar-size and collection-padding) https://stackoverflow.com/questions/32272158/how-to-use-a-stylus-variable-in-calc

clottman: ohh that makes sense! thanks!

PR #409, opened 2019-04-08
Melodious fifth: make algolia search fully usable

modernserf:
## Links
* Remix link: https://melodious-fifth.glitch.me/
* Manuscript link: https://glitch.manuscript.com/f/cases/3328254/

## GIF/Screenshots:
Screen Shot 2019-04-08 at 7 03 03 PM

## Changes:
* use new algolia indices
* create project item wrapper that can fetch its users
* assorted enhancements to cover image & avatar to more elegantly handle missing images
* make autocomplete search only update when results are loaded

## How To Test:
* enable algolia search at https://melodious-fifth.glitch.me/secret
* search using the search bar
* view the full results on the search page

## Feedback I'm looking for:
* Does the user placeholder in projects/collections look acceptable?
* Does autocomplete feel right?
* Are we making a reasonable number of network requests?

## Additional work needed before algolia search is complete
* collections need cover colors
* teams need verified badges when applicable
* glitch projects need `showAsGlitchTeam` when applicable
* should probably handle starter kits in algolia instead of on front end


ttseng:
This is really coming along.  A few UI suggestions;

1. Is it possible to override the default grid alignment for Top Results search-results?  
Setting `align-items: end` to the <ul> makes it more readable. 
image

2. Autocomplete top results only include starter kits - so sometimes top results are different from autocomplete vs on the search page (for example, looking up `react`).  Are there performance reasons for not including the exact name match in autocomplete top results?

3. In autocomplete, we can rename the "Collection Results" header to "Collections"

4. Currently before the search has finished, we show a loading animation and the note All Results for X.  The loading state seems to have been removed (it flashes More Ideas before showing the results).  Can we add the loading animation back (for people searching on slow networks, etc.)?
loading-missing

5. Will project descriptions be added in a followup PR?

Going to experiment on adding a simple animation to the placeholder avatar and will post a comment here when done.

modernserf:
- fixed top results alignment
- added top results to autocomplete (i must have forgotten to do this in a previous PR)
- the search _does_ render the spinner, if the network is slow enough -- you can test this out by setting the network speed to "Slow 3G" in the network tab

src/components/user/users-list.styl
```diff
@@ -9,6 +9,12 @@
   display: inline-block
   margin-right: -7px
   
+.placeholder
```
ttseng: Added some placeholder styling:

```
.placeholder
  width: 32px
  height: 32px
  border-radius: 50%
  background: linear-gradient(to right, secondary-background-hover 8%, input-border 20%, secondary-background-hover 35%);
  background-size: 800px 104px;
  animation: loadingGradient 1.8s cubic-bezier(0.4, 0, 1, 1) infinite;

@keyframes loadingGradient{
  0%{
      background-position: 50% 0
  }
  100%{
      background-position: -50% 0
  }
}
```

Viewable at https://sticky-cereal.glitch.me/search?q=react&activeFilter=project

(CSS as https://glitch.com/edit/#!/sticky-cereal?path=src/components/user/users-list.styl:30:0)

ttseng:
Also noticing that the autocomplete feels tight on mobile. 
image

in `styles/includes/header.styl` we can change the `.header-search` element to have 100% width on smallViewports to help alleviate the issue:

```
    .header-search
      display: inline-block
      width: 250px
      position: relative
      @media (max-width: smallViewport)
        width: 100%
```

src/components/user/users-list.styl
modernserf: this is great!

modernserf:
@ttseng updated:
- wider autocomplete on mobile
Screen Shot 2019-04-10 at 10 03 51 AM
- added loading animation

PR #411, opened 2019-04-09
Zest ostrich: Feature a Project Embed in Collections

sarahzinger:
## Links
* https://zest-ostrich.glitch.me/@sarahzinger/blossom-mix
* https://glitch.manuscript.com/f/cases/3328003/
* https://www.notion.so/glitch/Narrative-Collections-162c85d294a0453b9d57f7f8cdaa6783

## GIF/Screenshots:
Screen Shot 2019-04-09 at 4 20 44 PM
Screen Shot 2019-04-09 at 4 20 57 PM
Screen Shot 2019-04-09 at 4 21 09 PM
Screen Shot 2019-04-09 at 4 21 27 PM
Screen Shot 2019-04-09 at 4 21 46 PM

## Changes:
* Adds a `Feature` button to each project item on a collections page
* Adds `FeaturedProject` to the Collections Page that will render if there's a featured project
* Adds an optional `Note` to the `FeaturedProject` component
* Adds an `Add Note` button to the `FeaturedProject` component
* Some syntax changes to `FeaturedProject` in how we pass components to `ProjectEmbed`
* Change how we update state when an individual project changes in a collection so it's a bit more DRY
* Remove an unused class
* small DRY change to the project-options-pop
* update `Note` to be more explicit when users are authorized to update a note or not rather than depend on passing an update function, along with some var name changes
* Fix feature/unfeature animation bug

## How To Test:
* here's a collection with a featured project and some notes: https://zest-ostrich.glitch.me/@sarahzinger/blossom-mix
* on your own collection try to feature and unfeature a project, add and remove a note, etc
* double check featured projects on team and user pages still look good. 

## Feedback I'm looking for:
* Thoughts on weird things I might be doing codewise? In particular how do we feel about `FeaturedProject` it continues to feel weird to me the way I'm passing child components and I wonder if I'm approaching this the wrong way or if I'm over thinking it. 
* ~~I feel like the manipulation of `collection` state in `CollectionEditor` (particularly for `collection.projects` is getting to be a real pain in the neck. I wonder if it's time to reconsider the approach we've taken. I feel like in an ideal world we make a data request to a backend and then we just setState with the data we get back from the Patch, but the data we get back is incomplete. I can think of a few possible approaches to handling this, none of which feel easy or obviously right to me. Eager to hear other thoughts on this! Here are some of my thoughts on paths we could take:`~~
    * ~~maybe this is a sign we need redux here? Certain parts of state know to subscribe to certain patch requests and figure out how to update accordingly? This isn't fully fleshed out just an impulse? @modernserf I bet you have thoughts!~~
    * ~~we recall the `loadCollection` call from `collection.js` after each patch to update a collection. This would make more http requests but ensures we set state in the same way in the same place. We refetch all the users for each project which is maybe not performant but ensures a single source of truth, our backend. I wonder if this approach is easier/necessary for the work @keithk is doing for reordering projects in a collection? I suspect whatever your state needs for reordering, keith, are more pressing, and I'm happy to just follow whatever lead you take with that :) This PR can be merged after all that stuff is done if you want, happy to deal with the inevitable weird merge requests we get (sorry didn't think this would be so intertwined when I picked up the ticket...but I'm thinking it might be, or maybe not who knows)~~
    * ~~we continue down the path we're already on, as demonstrated in this pr, by making data requests when we need and (mostly) ignoring those responses. We manually manage state updates ourselves as needed.~~
    * ~~some other approach I'm just not thinking about?~~
* ui feedback. For example, I wonder if the note looks a little strange? Particularly in edit mode it expands, this looked ok-ish on a project item but is starting to look a bit weird to me above the featured project. Should it be full width I wonder? Probably at least on mobile/small screens? @ttseng (and others!) lend me your wisdom :) 

- [x] double check backend calls are working once those are finished
- [x] figure out why animation for unfeature is...not working I think? I think this might have been a regression I caused when componentizing featuredProjectEmbed
- [x] remove add note button when note is already present from featured project
- [x] fix featuring private projects bug
- [x] talk out if my approach to state management here is yucky and if it is revise it so it's better. (now that I'm not manually storing a featuredProject but just filtering based on the featuredProjectId, the state management feels much more reasonable and less weird)
- [x] clean up css and qa it
- [x] cross browser test
- [x] a quick voice over check to make sure I can navigate to all the buttons and stuff. 
- [x] fix bug where notes are out of place when you feature an item when there is already a feature
- [x] fix bug where logged in non-authors see feature button on user page

Post deploy checks
- [ ] Featured project: 
    - [ ] User (owns)
        - [ ] Can’t add note
        - [ ] Can’t feature private projects
        - [ ] Can pin private projects
        - [ ] Can feature project
        - [ ] Can unfeature project
    - [ ] Team (owns)
        - [ ] Can’t add note
        - [ ] Can’t feature private projects
        - [ ] Can pin private projects
        - [ ] Can feature project
        - [ ] Can unfeature project
    - [ ] Collection (owns)
        - [ ] Can add note 
        - [ ] Can delete note
        - [ ] Can’t feature private projects
        - [ ] Can feature regular projects
        - [ ] Can un feature project
        - [ ] Can add project
        - [ ] Can remove project
    - [ ]  Not logged in can view all that ^
    - [ ] Logged in but not owner can view and add project to collections
- [ ] ProjectEmbed
    - [ ] looks good on project page
- [ ] Animations:
    - [ ] Featuring/unfeaturing projects
    - [ ] adding/Removing pin
    - [ ] Showing/hiding all deleted projects
    - [ ] Delete a collection
    - [ ] Delete a project


src/components/project/project-embed.js
```diff
@@ -43,7 +43,7 @@ const ProjectEmbed = ({ project, topLeft, topRight, isAuthorized, currentUser, a
   );
 
   return (
-    <section className={styles.projectEmbed}>
+    <section id="project-embed">
```
modernserf: why are you using a global ID here, instead of a componentized class?

sarahzinger: yeah great question, so we use it to grab the element to animated when we feature/unfeature: https://github.com/FogCreek/Glitch-Community/blob/94286df7cf4d9f9ebb19859af13a0f6c04870d26/src/presenters/pop-overs/featured-project-options-pop.js#L10

but maybe a better way to do this would be to find the closest section the way we do with finding the closest li: https://github.com/FogCreek/Glitch-Community/blob/d8e6c8d1e2d5ba3c65a87afd664a7b87b68e8843/src/presenters/pop-overs/collection-options-pop.js#L10

modernserf: Can you use a ref for this?

sarahzinger: probably! good idea let me try :)

ttseng:
Really cool to see this combine your project-embed componenitization work!

I had some thoughts/questions about the structure of the project-embed component:

* The component has a `buttonContainer` that contains `topLeft` and `topRight` divs.  I think `topLeft` is always going to contain the `heading` element though, so it's not clear to me why this is in a `buttonContainer`.  

Having the heading + the dropdown button in the same parent container leads to some alignment issues – For example, the dropdown for the embed should appear in the top right, directly above the embed, but once the note is injected between the header and the dropdown, the dropdown appears disconnected from the embed:

image

I would expect the project-embed to be nested in this way:
```
<section id="project-embed">
   <h2 class="heading"/>
   <div class="note"/>
   <div class="dropdown"/>
   <div class="embedContainer"/>
   <div class="buttonContainer"/>
</section>
```
(The last buttonContainer makes sense because it does contain only button elements)

With this restructing and the styling changes listed below, we get this collection page:
image

CSS Changes:
* 10px margin added to `project-embed` section so that its left aligned with other components on the page.  Note, this also removes the `margin-top: 2rem` default that adds unnecessary space above the `project-embed` section when viewing collections you don't own, for example: 
https://zest-ostrich.glitch.me/@scientiffic/svg
(https://zest-ostrich.glitch.me/@scientiffic/svg)

* note width adjusted to `width: calc(95%)` - I agree with you that it looks better wider!

Other suggestions:
* Add Note in the options pop-over should appear above Unfeature Project (the general convention is that stuff you add appears below stuff you remove in a pop-over)

src/presenters/collection-editor.js
```diff
@@ -53,38 +53,63 @@ class CollectionEditor extends React.Component {
...
+    });
+
+    if (this.state.featuredProject && this.state.featuredProject.id === projectId) {
+      stateUpdates.featuredProject = { ...this.state.featuredProject, ...projectUpdates };
+    }
+
+    this.setState(stateUpdates);
+  }
+
+  async featureProject(id) {
```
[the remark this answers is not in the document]
sarahzinger: ahhhh that is much simpler lol, thank you.

src/components/project/featured-project.js
```diff
@@ -1,49 +1,95 @@
...
-    <ProjectEmbed
-      topLeft={<TopLeft />}
-      topRight={<TopRight />}
-      project={featuredProject}
-      unfeatureProject={unfeatureProject}
-      isAuthorized={isAuthorized}
-      currentUser={currentUser}
-      addProjectToCollection={addProjectToCollection}
-    />
+    <div ref={featuredProjectRef}>
```
sarahzinger: I find it a little annoying I can't put the ref on ProjectEmbed directly and had to wrap it with this div to get it to work, but maybe there's a better way and I just don't know enough about refs?

modernserf: this might help https://reactjs.org/docs/forwarding-refs.html

sarahzinger: Yeah that does seem to be the right direction but for some reason I can't get this to work that way? I've tried creating a separate `ProjectEmbedWithRef` that is made using `React.forwardRef` and passing the ref along that was created in `featuredProject` but it errors our with `Warning: Function components cannot be given refs. Attempts to access this ref will fail. Did you mean to use React.forwardRef()` which is weird because I thought you could with the `useRef` hook? Started refactoring to make it a class, but honestly this all works as is I kind of wonder if it's fine as is? What do you think? Feel free to be like "nah we should use forwardRef keep at it" maybe I'm fundamentally misunderstanding something about refs here?

modernserf: div wrapper is probably fine.

keithk: I think the div wrapper is fine!

sarahzinger:
@ttseng made your styling suggestions also added https://github.com/FogCreek/Glitch-Community/pull/411/files#diff-16c0d76395eb1bb86715c0dda2c51391R41 to overwrite the margin on projectEmbed as it was looking a bit misaligned to me on the project page but let me know if you think that's off!

ok I think it should be good for another look, this thing is weirdly complicated and touches a lot of stuff, so I made a bit post deploy checklist to test once this is on staging to make sure it's not gunna break anything. Went through it all on my remix and I think we're good! 🤞

src/components/project/featured-project.styl
```diff
@@ -1,3 +1,14 @@
...
+  
+.left
+  width: calc(99%)
+  
+.top
+  display: flex
+
+.note
+  margin-bottom: 15px
+
```
ttseng: Small suggestion - when you first add the note, the avatar and the featured embed dropdown are very close to one another.

image

I'd suggest we add `width: 95%` to the `.note` element here:

image

ttseng:
awesome!  the structure of `featured-project` and `project-embed` make a lot of sense to me now - thanks a lot for making those changes.  Tested everything and didn't run into any issues!

src/components/project/featured-project.styl
sarahzinger: good catch thanks :)

PR #413, opened 2019-04-10
Set up proxy /sitemaps → sitemaps.glitch.me 

taravancil:
This routes /sitemaps to [sitemaps.glitch.me](https://sitemaps.glitch.me)

This change does not deploy all sitemaps, it just allows us to submit [this test sitemap](https://sitemaps.glitch.me/projects/sitemap-test.xml) to the google crawler (it needs to live on glitch.com in order to submit it to google).

After submitting this to google, we'll better understand if and how crawlers wake up projects, and be able to make a more final plan for deploying all sitemaps.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,3 @@
-
 # Glitch Community Projects
 
 Discover new reasons to code, remix cool and helpful projects made with Glitch.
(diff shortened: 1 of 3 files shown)


taravancil:
@GregWeil Yep! We'll have a top-level sitemap index, and then one each for projects, users, teams, and collections

PR #414, opened 2019-04-10
Nova chance: Fix a regression with anon user avatars, use the avatar component everywhere

GregWeil:
## Links
* https://nova-chance.glitch.me/
* https://glitch.fogbugz.com/f/cases/3328270/Anon-users-aren-t-always-rendering-their-background-color

## GIF/Screenshots:
https://glitch.com/@gregs-test-team-1
image
https://nova-chance.glitch.me/@gregs-test-team-1
image

## Changes
* Added a backgroundColor prop to the avatar component
* Removed src/presenters/includes/avatar.js

## How To Test:
Check out the two sites side by side, look for differences in how they're rendered, particularly with alignment or sizing

## Feedback I'm looking for:
Did I create any new style regressions in the process?


GregWeil:
They should! thanks for catching that

taravancil:
Besides my comment above, this looks good! I very well could've missed something though, we have a lot of avatars 😅

ttseng:
made some design decisions a few weeks ago to remove the anonymous user avatar from the header because we got feedback that it was confusing & there was no longer a direct sign in button as a result: https://github.com/FogCreek/Glitch-Community/pull/334

So the anonymous user avatar + options-pop should be replaced by the sign-in button:
image

ttseng:
looking more into this, I'm worried that the bright default colors are making the anonymous users stand out more than they should in search results, team pages, etc.  
going to discuss in #design

GregWeil:
Hmm, I don't see the user menu when I'm anon on nova-chance.

ttseng:
ok never mind - I was thinking these were anonymous users, but actually they're users who created an account via email and never changed their avatar.  So showing background colors is fine!

ttseng:
ah! I was logged in to an old user account on your remix.  logging out, I see the sign in button now.  👍

PR #416, opened 2019-04-10
Organic dance: UI Tweaks to Team Analytics

ttseng:
## Links
* Remix link: https://organic-dance.glitch.me/@community
* Specs/Notion docs: some of the issues resolved here referenced https://www.notion.so/glitch/Adding-analytics-to-project-pages-c1278a2992aa4a0cb4bef21376fe30b8#d6ed76428abe4c1abc66f758ae7613ac

## GIF/Screenshots:
analytics

## Changes:
* Moved App Views & Remixes to segmented tabs and separated corresponding charts
* c3 chart styling (including tooltips + colors, axes scaling to always start from 0, etc.)
* Added dropdown icon to filters

## How To Test:
* Go to any team page that you're a part of and test the analytics at the bottom of the page
* Switch between App Views & Remixes
* Add project filter
* Add timeframe filter


styles/c3.styl
```diff
@@ -180,3 +196,22 @@
 
 .c3-chart-arc.c3-target.c3-focused g path
   opacity: 1
+
+/*-- Custom Styling for Glitch Analytics --*/
+.c3-line-Total-App-Views{
+  stroke: analytics-app-views-color !important
+  stroke-width: 2px
```
taravancil: Now that this stroke is thicker, it feels like maybe the circle radius (2.5) is a bit too small? Not a big deal, but I did feel like I had to strain my eyes a bit when I was trying to hover on one of the dots

Before:
Screen Shot 2019-04-10 at 3 20 28 PM

After:
Screen Shot 2019-04-10 at 3 20 35 PM

src/presenters/includes/team-analytics-activity.js
```diff
@@ -90,14 +90,36 @@ const renderChart = (c3, analytics, currentTimeFrame) => {
...
+  } else if (activeFilter === 'remixes') {
+    chart.hide(['Total App Views']);
+  }
 };
 
 class TeamAnalyticsActivity extends React.Component {
   componentDidUpdate(prevProps) {
-    if (prevProps.isGettingData === true && this.props.isGettingData === false) {
-      renderChart(this.props.c3, this.props.analytics, this.props.currentTimeFrame);
+    if (prevProps.activeFilter !== this.props.activeFilter || (prevProps.isGettingData === true && this.props.isGettingData === false)) {
```
taravancil: Don't _need_ to do this, but might be nice to break this down into two variables, something like:

```js
const newFilter = prevProps.activeFilter !== this.props.activeFilter;
const stillFetching = prevProps.isGettingData && !this.props.isGettingData;
// ^ also i think we can use this shorthand instead of === true or === false

if (newFilter || stillFetching) {
  renderChart(...);
}
```

Feel free to decide against this or use different variable names than what I used!

styles/includes/team-analytics.styl
```diff
@@ -1,31 +1,34 @@
...
     padding: 3px 5px
     color: primary-background
     &.app-views    
-      background-color: analytics-app-views
+      background-color: analytics-app-views-color
     &.remixes
-      background-color: analytics-remixes
-
+      background-color: analytics-remix-color
+
```
taravancil: I think these are extraneous spaces

styles/c3.styl
ttseng: good idea!  I made the points bigger: 
image

src/presenters/includes/team-analytics-activity.js
ttseng: that makes sense!

PR #419, opened 2019-04-10
Cheerful flamingo: use "Thanks" component everywhere

modernserf:
## Links
* Remix link: https://cheerful-flamingo.glitch.me/

## GIF/Screenshots:
Screen Shot 2019-04-10 at 3 46 30 PM
Screen Shot 2019-04-10 at 3 46 45 PM
Screen Shot 2019-04-10 at 3 50 20 PM

## Changes:
* use Thanks component everywhere
* get rid of redundant thanks count checking
* delete old Thanks presenter

## How To Test:
* example team: https://cheerful-flamingo.glitch.me/@glitch
* click on a team member to see the user info pop
* example user: https://cheerful-flamingo.glitch.me/@anildash
* example user with zero thankses https://cheerful-flamingo.glitch.me/user/123

## Feedback I'm looking for:
* do these look right?
* should the Emoji component have configurable sizes? (It seems like its overridden in a lot of places)


oliviacpu:
`should the Emoji component have configurable sizes? (It seems like its overridden in a lot of places)`

If there's a set of sizes that are used frequently it could be worth having them be set with a prop or something.

PR #420, opened 2019-04-10
Pointy mine: use Embed component everywhere

modernserf:
## Links
* Remix link: https://pointy-mine.glitch.me/

## GIF/Screenshots:
Screen Shot 2019-04-10 at 4 18 38 PM
Screen Shot 2019-04-10 at 4 18 54 PM

## Changes:
* use Embed component on landing page
* make container responsible for Embed size, not component itself
* delete old Embed presenter

## How To Test:
* landing page: https://pointy-mine.glitch.me
* user page: https://pointy-mine.glitch.me/@modernserf

## Feedback I'm looking for:
* are there cases where this breaks?
* does this interfere with @sarahzinger's current PR for embeds?


src/presenters/includes/embed.js
```diff
@@ -1,58 +0,0 @@
...
-    console.log(
-      "Sorry, you don't have the necessary JavaScript features to run Glitch code editors. Try applying your latest system updates, or try again with a different web browser.",
-      error,
-    );
-  }
-  return false;
-})();
-
-const Embed = ({ domain }) => (
-  <div className="glitch-embed-wrap">
```
sarahzinger: I think this class is sprinkled in a bunch of places, some of which we might be able to clean up? not 100% probs good tho!

src/presenters/includes/embed.js
```diff
@@ -1,58 +0,0 @@
...
-        allow="geolocation; microphone; camera; midi; encrypted-media"
-        height="100%"
-        width="100%"
-        border="0"
-        allowvr="yes"
-      />
-    ) : (
-      // Error message if JS not supported
-      // TODO(sheridan): Refactor this once we have a true error component
-      <div className="error-container">
```
sarahzinger: you might want to verify on Chrome 41 that this error thing looks good (if you haven't already!), I only tested it at the size for the user/team/project pages I think?

src/presenters/includes/embed.js
modernserf: yeah -- i ended up using this to wrap the embed on the landing page.

PR #426, opened 2019-04-11
Nova chance: Check for cancelled oauth signin, refactor login to a functional component

GregWeil:
## Links
https://glitch.fogbugz.com/f/cases/3328102/login-facebook-page-attempts-token-exchange-API-request-without-a-valid-code-on-declined-auth

## Changes:
* The LoginPage components are all functional now
* All pages respect `error=access_denied` rather than just slack

## How To Test:
* Try signing in (you'll have to block glitch.com so you can steal the url/code and point it to nova-chance instead)
* Try signing in with an expired code (and get an error page)
* Try canceling sign in (and end up back where you were) (only facebook and slack allow it)


src/presenters/pages/login.js
```diff
@@ -43,38 +43,33 @@ const RedirectToDestination = () => {
...
-      errorMessage: null,
-    };
-  }
+const LoginPage = ({ provider, url }) => {
+  const api = useAPI();
+  const { login } = useCurrentUser();
 
-  async componentDidMount() {
-    const { api, provider, url } = this.props;
+  const [done, setDone] = React.useState(false);
```
modernserf: these all change together, and done / error are mutually exclusive. how about something more state-machine-y, like

```js
const [state, setState] = useState({ status: 'loading' });
const setDone = () => setState({ status: 'done' });
const setError = (errorData) => setState({ status: 'error', message: errorData && errorData.message });
```

oliviacpu:
@GregWeil I think this is good now. I pulled in the changes Justin suggested into my remix based off this.

PR #427, opened 2019-04-12
Stitch close: no, actually fix the search page under algolia

modernserf:
Manuscript case: https://glitch.manuscript.com/f/cases/3328292/

turn on algolia search with https://stitch-close.glitch.me/secret
try the search page https://stitch-close.glitch.me/search?q=react

### DIFF
--- a/src/components/search/search-results.js
+++ b/src/components/search/search-results.js
@@ -42,7 +42,7 @@ const FilterContainer = ({ filters, activeFilter, setFilter, query }) => {
 // so those need to be fetched after the search results have loaded.
 const useUsers = createAPIHook(async (api, userIDs) => {
   if (!userIDs.length) {
-    return [];
+    return null;
   }
   const idString = userIDs.map((id) => `id=${id}`).join('&');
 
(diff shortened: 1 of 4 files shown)


src/components/search/search-results.js
```diff
@@ -42,7 +42,7 @@ const FilterContainer = ({ filters, activeFilter, setFilter, query }) => {
 // so those need to be fetched after the search results have loaded.
 const useUsers = createAPIHook(async (api, userIDs) => {
   if (!userIDs.length) {
-    return [];
+    return undefined;
```
modernserf: As i understand it, this is what was happening on the search page:
A  project would have users, but no team members.
First, this would result in `{ users: undefined, teams: undefined }`, and the placeholder user list would appear.
Then, the team result would return `[]` immediately, since no API call would result.
This would result in `{ users: undefined, teams: [] }`, and the placeholder user list would _not_ appear (because of an earlier "fix" for the non-algolia search results: #417. Instead, it would attempt to render the user list, and crash on `if (users.length) {...}`

PR #430, opened 2019-04-12
Statuesque father: remove `notSafeForKids` items from algolia search

modernserf:
Not Safe for Kids hasn't been released publicly yet, but it's tracked in the search index and may be used by the support team. This PR enables it on by default.

(Incorporates #427 to make algolia search usable.)


src/state/search.js
```diff
@@ -124,11 +124,21 @@ const formatAlgoliaResult = (type) => ({ hits }) =>
 
 const algoliaProvider = {
   ...mapValues(searchIndices, (index, type) => (query) => index.search({ query, hitsPerPage: 100 }).then(formatAlgoliaResult(type))),
+  project: (query, { notSafeForKids }) =>
```
modernserf: all the real changes for this PR are in this file.

PR #432, opened 2019-04-15
different-print: Change "delete project" animation (and test it?)

oliviacpu:
## Links
* Remix link: https://different-print.glitch.me/
* Manuscript link: https://glitch.manuscript.com/f/cases/assign/3328210/Deleting-a-project-then-deleting-another-can-be-confusing
* Specs/Notion docs: N/A

## Changes:
* The delete animation for a project is now a partial fade, rather than sliding down

## How To Test:
* You can go to your profile, delete a project, and watch the animation
- OR -
* You can try running `./sh/cypress.sh different-print` and it should run through the scenario for you (running it against `community` will show you how it looks on that remix)

## Feedback I'm looking for:
- If you're at all familiar with Cypress I would like another set of eyes on some of the testing and test setup.
- How does the delete animation look?

- [ ] IDK maybe remove my test account information, maybe not?
- [ ] Clean up some of the test setup probably.
- [ ] Decide if we like the animation change.


styles/pages/user-and-team-page.styl
```diff
@@ -57,7 +57,7 @@
       animation-fill-mode: forwards
 
     .slide-down
-      animation: slideDown 0.1s ease-out
+      animation: fadeHalf 0.2s ease-out
```
keithk: This looks way better!

keithk:
Yessss this cypress test, so cool 😍

oliviacpu:
💁 I think I can put the fixtures into `.data` with them ignored, and come up with a script to generate them for people if they don't exist.

PR #435, opened 2019-04-16
Chipped bangle: Changes to filename hashes for bundles

keithk:
## Links
* https://chipped-bangle.glitch.me/
* https://glitch.fogbugz.com/f/cases/3328316/Webpack-Bundle-Caching

## Changes:
- Minor version upgrades for webpack related packages
- Bundles now have hash in the file name, so there should never be a case where the wrong version is cached. Ex: `/styles.689f0b95.css` and `/client.8571faf9.js`
- Switched to using *chunkhash* for the javascript files, which is best practice
- Using `clean-webpack-plugin` to clear the dist directory before every build

## How To Test:
- Trigger a webpack rebuild. I tested to make sure that there wasn't a time between triggering a rebuild & it finishing where we wouldn't have any output on dev, but it didn't seem to be the case.

## Open Questions
- Look back at #52 and consider moving back to html-webpack-plugin, but that isn't part of the issue


package.json
```diff
@@ -75,17 +75,18 @@
...
     "url-join": "^4.0.0",
     "url-search-params-polyfill": "^5.1.0",
-    "webpack": "^4.29.3",
+    "webpack": "^4.30.0",
     "webpack-cli": "^3.3.0",
-    "webpack-dev-middleware": "^3.6.1",
-    "winston": "^3.2.1"
+    "webpack-dev-middleware": "^3.6.2",
+    "winston": "^3.2.1",
+    "clean-webpack-plugin": "^2.0.0"
```
oliviacpu: `npx sort-package-json` in the project console will alphabetize this.

keithk: omg i've always wanted this

keithk:
@GregWeil great question, i have no idea. deleted!

PR #436, opened 2019-04-16
Componentize collection avatars

taravancil:
[congruous-oriole.glitch.me](https://congruous-oriole.glitch.me)

## Changes:
This moves the collection avatar component to the components directory and updates all instances of it to use this new component

**Note**
There are some old collection avatar styles that need to be "componentized". First I'm going to componentize `ProjectAvatar`, then in a separate PR I'll move on to pulling out *all* old avatar styles.

## How To Test:
View a collection, a collection list, and an add to collection pop

## Feedback I'm looking for:
Are there any design regressions?

### DIFF
--- a/src/components/images/avatar.js
+++ b/src/components/images/avatar.js
@@ -3,6 +3,7 @@ import PropTypes from 'prop-types';
 
 import TooltipContainer from 'Components/tooltips/tooltip-container';
 import Image from 'Components/images/image';
+import DefaultCollectionAvatar from 'Components/collection/defaultAvatar';
 
 import { DEFAULT_TEAM_AVATAR, getAvatarUrl as getTeamAvatarUrl } from 'Models/team';
 import { ANON_AVATAR_URL, getAvatarThumbnailUrl, getDisplayName } from 'Models/user';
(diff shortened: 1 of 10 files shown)


src/components/images/avatar.js
```diff
@@ -95,3 +88,24 @@ UserAvatar.defaultProps = {
   hideTooltip: false,
   withinButton: false,
 };
+
```
modernserf: think this was supposed to be replaced with a reference to `components/collection/collectionAvatar.js`

taravancil:
@sheridanvk suggested we see about consolidating the `Avatar` component to get rid of the `ProjectAvatar`, `TeamAvatar`, `UserAvatar`, `CollectionAvatar` variants. I'm going to work on that in a separate PR and will likely close this.

ttseng:
would it make sense to consolidate `components/collection/defaultAvatar.js` and `components/collection/collectionAvatar.js` into one file?  (probably get rid of `defaultAvatar` and put everything in `collectionAvatar`

src/components/images/avatar.js
sheridanvk: Or if we're importing all Avatars from avatar.js, would it make more sense to just delete components/collection/collectionAvatar.js and have the logic live here? If the standalone component file exists separately, it seems odd to import it into avatar.js, and then import it from here into other files - we could just import it directly from collectionAvatar.js if we're choosing to creating a separate component.

taravancil:
Doing this work together with some other avatar work

PR #437, opened 2019-04-16
Quixotic rocket: Lots of refactors and fixes for server code + forever

GregWeil:
## Links
https://glitch.fogbugz.com/f/cases/3328298/

## Changes:
* Use `forever` to restart the server if it crashes (it'll wait 5s if it crashes within 5s of starting)
* Refactored api.js a bunch, it had a lot of 404 handling that isn't applicable to the new api
* Only print out api errors once, rather than every time we pull the error from the cache
* Print out the cache key with api errors so we know what failed to load
* Wrap sentry helpers in a try/catch because I don't want anything going wrong there
* Sentry grabs uncaught promises, but also `console.error` them to be safe

## How To Test:
Try throwing a `throw new Error('asdfasdf')` in the code somewhere and watch the logs. It should always get printed out and the server should come back up in a few seconds


server/api.js
```diff
@@ -3,98 +3,67 @@
...
+    cache.put(key, promise, CACHE_TIMEOUT);
+  }
   try {
-    const response = await api.get(`https://culture-zine.glitch.me/culture/ghost/api/v0.1/posts/?${client}&${params}`);
-    return response.data.posts;
+    return await promise;
   } catch (error) {
+    if (trackError) {
+      console.warn(`Failed to load ${key}: ${error.toString()}`);
+      captureException(error); // it'll probably get filtered out by beforeSend
```
clottman: Do we want to attempt to add the `key` as a breadcrumb?

server/server.js
```diff
@@ -15,15 +15,31 @@ try {
...
     },
   });
   Sentry.configureScope((scope) => {
     scope.setTag('PROJECT_DOMAIN', process.env.PROJECT_DOMAIN);
   });
+  // Node doesn't log unhandled promise errors if something is listening for
+  // them, which Sentry does. Add our own event listener to keep them logged
+  // https://github.com/getsentry/sentry-javascript/issues/1909
+  process.on('unhandledRejection', (reason) => {
+    console.error('Unhandled Rejection:', reason);
```
keithk: Should we exit the application here, and let forever restart it? I'm concerned about finding unhandled exceptions and then the application thinking that it's in a clean state

server/api.js
clottman: Actually never mind, `console.warn` is automatically a breadcrumb

server/server.js
GregWeil: I don't know, how would that affect other in-flight requests?

keithk: Yeah, the goal is to have it happen infrequently but make it so that it's something we feel pressured to fix

GregWeil: 👍 done

GregWeil:
Fixed the shrinkwrap updating a bunch of extraneous packages after a bit of a struggle

PR #439, opened 2019-04-17
hail-diploma: Change reference of environment

oliviacpu:
## Links
* Remix link: https://hail-diploma.glitch.me/
* Manuscript link: https://glitch.manuscript.com/f/cases/3328329/operations-has-an-error-RUNNING_ON-is-not-defined-in-api-auth-ejs
* Specs/Notion docs: n/a

## Changes:
* Load `RUNNING_ON` in `api-auth.ejs`.

## How To Test:
* Everything should be the same as before.

## Feedback I'm looking for:
* ~There's a note in the `api-auth.ejs` about potential security implications changing this setup. Is this a safe change?~
* ~Should the template reference the `currentEnv` variable instead?~
* ~Should I just reverse the check in the template (`"development" === RUNNING_ON`?)~

- [x] Discuss.

### DIFF
--- a/server/routes.js
+++ b/server/routes.js
@@ -21,10 +21,7 @@ module.exports = function(external) {
   // CORS - Allow pages from any domain to make requests to our API
   app.use(function(request, response, next) {
     response.header('Access-Control-Allow-Origin', '*');
-    response.header(
-      'Access-Control-Allow-Headers',
-      'Origin, X-Requested-With, Content-Type, Accept',
-    );
+    response.header('Access-Control-Allow-Headers', 'Origin, X-Requested-With, Content-Type, Accept');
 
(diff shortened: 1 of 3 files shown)


GregWeil:
It looks like api-auth.ejs ignores the `RUNNING_ON` template value. I think it just needs `var RUNNING_ON = '<%= RUNNING_ON %>';` like index.ejs has.

GregWeil:
`config.currentEnv` is never set to 'development', it uses prod while the client hits your dev server. That might be an oversight, since the client env used to be picked based on the url only until recently.

oliviacpu:
@GregWeil That makes sense. Thank you! I changed it to load that constant and also ran prettier on the routes.

PR #440, opened 2019-04-17
Apple tang: bring back notes to the homepage

sarahzinger:
## Links
* https://apple-tang.glitch.me
* https://glitch.manuscript.com/f/cases/3328336/

## GIF/Screenshots:
On the homepage:
Screen Shot 2019-04-18 at 10 19 26 AM
In a collection:
Screen Shot 2019-04-18 at 10 19 38 AM

## Changes:
* bring back notes to the homepage

## How To Test:
* add `  { owner: 'glitch', name: 'sketch-with-p-5' }` to src/curated/collections.js so you can see a collection with notes in it on the homepage (plz remove when done!)
* verify you can edit notes on collections you own on the collections page but not on the homepage
* verify you can't edit notes on collections you don't own
* verify that ProjectsUL looks good in all the places it's used:
    * homepage collections
    * user and team pages: pinned projects + "recent projects" 
    * project page "related projects"
    * category page
    * collection page
     
## Feedback I'm looking for:
* this is right yeah? Just want to make sure we're not losing any style fixes as well. One thing that seems noticable to me is the distance between the note and the faces is different on the homepage than on the collection page, but I'm thinking that might be intentional seeing as there's css in the code to make that happen. (UPDATE: yes the spacing was weird, yes it's been fixed!)


src/components/containers/projects-list.js
```diff
@@ -158,17 +160,22 @@ PaginatedProjects.defaultProps = {
   projectsPerPage: 6,
 };
 
-export const ProjectsUL = ({ showProjectDescriptions, ...props }) => (
+export const ProjectsUL = ({ showProjectDescriptions, collection, ...props }) => (
   <ul className="projects-container">
     {props.projects.map((project) => (
       <li key={project.id}>
+        <Note
```
modernserf: Do we want the note component in _every_ list of projects, or just in collections?

sarahzinger: hmmm good point, maybe I should make a CollectionProjectsUL or something like that and then use it in collections and featuredcollections yeah?

sarahzinger: maybe I'll just conditionally show note if we're passed a collection prop. Seems easiest, I'll then use this on the collections page

ttseng:
Thanks for adding the notes back!

There shouldn't be a 7px top margin on the note element - can you remove it?

Also, I'm noticing that the top margin on the `project-item__containers` is disproportionally large compared to the bottom margin in `wide-collection` which you can especially see with collections w/o notes.

image

I think it's because we're using the `top-margin` on the `profile-list__header` to set spacing between rows of profile-items, when we should be handling that on the grid level.

We can get rid of this problem with these changes:
* in `src/components/project/project-item.styl` set 
```
.header
    margin-top: 0rem
```
* in `styles/note.styl` set 
```
.note
     margin-bottom: -15px
```

* in `styles/pages/user-and-team-page.styl` set
```
.projects-container, .collections-container
    .row-gap: 1.5rem
```

sarahzinger:
ok thanks for the feedback @modernserf @ttseng mind taking another look for me?

PR #442, opened 2019-04-18
Superb guide: Remove EditableField and EditableWrappingField

GregWeil:
## Links

## Changes:
Use componentized inputs in all places, removed redundant presenter inputs

## How To Test:
Try using text inputs on user/project/team pages


GregWeil:
Sure, I'll add some stories. EditableDescription is a textarea that renders as markdown when it isn't focused. When I componentize it I'll rename it to MarkdownInput since that's its real purpose

GregWeil:
Added a whole pile of stories
https://superb-guide.glitch.me/storybook/index.html?path=/story/text-input--text-input
I also made a new file to keep input stories, because stories/index.js is really big.

stories/inputs.js
```diff
@@ -0,0 +1,42 @@
...
+import UserNameInput from 'Components/fields/user-name-input';
+import UserLoginInput from 'Components/fields/user-login-input';
+
+const inputStory = storiesOf('Text Input', module);
+inputStory.add('text input', () => <TextInput placeholder="type something!" />);
+inputStory.add('affixes', () => <TextInput placeholder="type something!" prefix="#" postfix="#" />);
+inputStory.add('search', () => <TextInput type="search" opaque={true} search={true} placeholder="bots, apps, users" />);
+inputStory.add('with error', () => <TextInput placeholder="glitch" error="That team already exists" />);
+inputStory.add('text area', () => <TextArea placeholder="This is a multiline text field" error="Reason is required" />);
+inputStory.add('wrapping text', () => <WrappingTextInput placeholder="This is a text field that wraps instead of scrolling" error="An error could go here!" />);
```
sarahzinger: I'm not seeing the difference visually here between `text area` and `wrapping text` is there a way to illustrate it in the example a bit more?

GregWeil: Visually they're the same, but the wrapping text filters out newlines so it acts like a TextInput rather than a TextArea

PR #447, opened 2019-04-22
Strong duchess: enable algolia search for everyone

modernserf:
Turn algolia search flag on by default.

if all goes well, a future PR will remove the non-algolia search code.


cori:
I mentioned this in Slack but in case I buried the lede; it's pretty important from an abuse / support standpoint not to turn Algolia on for everyone until either NSFK is fully working or private projects are filtered - until we have NSFK our solution to NSFK content in projects is to request community members to make their NSFK projects private (and to do it ourselves if they don't comply).

PR #448, opened 2019-04-22
Update Style.md with new details for the componentisation process

sheridanvk:
Just a text change to Style.md - looking for feedback (mostly from Justin, but anyone else welcome to weigh in too!) on the change to the componentisation process.


STYLE.md
```diff
@@ -18,11 +18,17 @@ To add a new component, or convert an existing piece of our code into a CSS Modu
     * Only export one component per `.js` file. The component you're exporting should have the same name as the file it's contained in (e.g. `TextArea` is exported from `text-area.js` and styles should be in `text-area.styl`). 
     * Classnames should be camelcased to make them easier to refer to in Javascript (e.g. `markdownContent` not `markdown-content`).
 1. Where possible, we'd like to avoid passing styles into the components via props and instead rely on named props that cover the use cases of the different modes of the component (e.g. `Button` accepts a `type` prop like `"tertiary"` or `"dangerZone"`) that can be passed into the component. The named props then define the styles that apply - the classnames npm package can be used to combine these in more readable ways (see [Button.js](https://glitch.com/edit/#!/community?path=src/components/buttons/button.js:15:0) for an example)
+    * Don't use HTML tag names like button or p to override styles in parent components though - if this is happening, either define a different set of props on the child component to accommodate the style you need or pass in classNames.
```
modernserf: I'm hesitant to even use "don't" here -- i'm sure there are cases where this is good or necessary. I think its more about the order in which we do these modifications -- first consider a wrapper div, then consider adding a prop, then consider passing a className; only if that won't work should you style the tag directly.

sheridanvk: good call 👍  I like being clear about the order

sheridanvk:
@modernserf updated and also added your point about not adding margins, etc, to these components.

STYLE.md
```diff
@@ -17,12 +17,23 @@ To add a new component, or convert an existing piece of our code into a CSS Modu
...
     * Classnames should be camelcased to make them easier to refer to in Javascript (e.g. `markdownContent` not `markdown-content`).
+1. In general don't add margins, absolute positioning or fixed width/heights into the component - leave that to the parent components.
 1. Where possible, we'd like to avoid passing styles into the components via props and instead rely on named props that cover the use cases of the different modes of the component (e.g. `Button` accepts a `type` prop like `"tertiary"` or `"dangerZone"`) that can be passed into the component. The named props then define the styles that apply - the classnames npm package can be used to combine these in more readable ways (see [Button.js](https://glitch.com/edit/#!/community?path=src/components/buttons/button.js:15:0) for an example)
+    * Sometimes you'll need additional styles from parent components. Here's the preferred order of options to use here - ideally use the first one that works for your use case:
```
sarahzinger: if we can think of specific examples for when each of these strategies would be the preferred one that would be super helpful btw, but I could see that happening in another pr if we want

sheridanvk: Added some examples for each case inline (tagging @modernserf to re-review)

PR #450, opened 2019-04-22
luminous-buggy: add renderAsPlaintext option to Markdown option

taravancil:
* [luminous-buggy.glitch.me](https://luminous-buggy.glitch.me)
* [manuscript](https://glitch.manuscript.com/f/cases/3328313/Raw-markdown-rendered-in-add-project-to-collection-pop-and-maybe-elsewhere)

## Changes:
This adds a `renderAsPlaintext` prop to the `Markdown` component.

This was necessary because sometimes markdown content is rendered inside a button or otherwise clickable element, and we don't want the user to click a rendered markdown link

**Before**
markdown1

**After**
markdown2

## How To Test:
Open the add project to collection pop and confirm that collection descriptions are rendered as plaintext with no raw markdown.

## Feedback I'm looking for:
- Is this approach sound? Are there any weird security side effects I'm not thinking of?
- Should we move the `stripHtml` helper to `/utils`?
- Does this break the `Markdown` component in unexpected ways?


taravancil:
@GregWeil I intentionally removed the styling because the `Markdown` component specifies `font-family`, `font-size`, and a few other styles, but at second thought, that doesn't feel quite right.

If we're going to render markdown in a bunch of different contexts, each context will potentially have different styles. We may want to remove those specific styles from the `Markdown` component and apply them to outer wrappers instead. I'll look into it.

modernserf:
What is the advantage of doing this, vs just rendering the markdown as a string?

taravancil:
Added this to `project-result-item.js`! User result items don't include any markdown.

> If we're going to render markdown in a bunch of different contexts, each context will potentially have different styles. We may want to remove those specific styles from the Markdown component and apply them to outer wrappers instead. I'll look into it.

Let's save this for another PR so it's isolated from these changes. I'll create a case and assign myself.

taravancil:
> What is the advantage of doing this, vs just rendering the markdown as a string?

@modernserf I might be misunderstanding your question so forgive me if this isn't what you're getting at, but the raw markdown strings include `[link text](href)`s which is not great.

modernserf:
I like raw markdown! those links have information in them! But I understand why you'd want to strip them.

taravancil:
Totally! I feel like they're kind a pain to look at in this context

image

taravancil:
lol MarkdownIt *does* have this as a builtin helper, but...

image

https://markdown-it.github.io/markdown-it/#MarkdownIt.renderInline

taravancil:
This is ready to merge this, but I *cannot* get this remix's container to behave, not even with `refresh`. I'll wait and try again tomorrow.

taravancil:
Here's how this collection description...

image

...looks in a `*-result-item`:

image

Gonna merge it!

PR #451, opened 2019-04-22
Profuse tea: Include ips in the request log

GregWeil:


cori:
I'm not a privacy expert, but IMO IPs in logs that last about 30 mins on disk seems an acceptable balance between ability to troubleshoot and desire to protect privacy.

GregWeil:
Increased server api cache time from 15 minutes to 1 hour. That will affect metadata and the homepage culture zine posts.

server/server.js
```diff
@@ -49,6 +49,7 @@ try {
 require('dayjs').extend(require('../shared/dayjs-convert'));
 
 const app = express();
+app.enable('trust proxy');
```
clottman: What's this doing?

sheridanvk: Looks like it enables Express to trust the headers and therefore enables us to log the IP addresses: https://stackoverflow.com/questions/23413401/what-does-trust-proxy-actually-do-in-express-js-and-do-i-need-to-use-it

PR #452, opened 2019-04-23
Heady path: use personalized search credentials to remove private projects

modernserf:
## Links
* Remix link: https://heady-path.glitch.me

## Changes:
* use `/search/creds` to get a personalized algolia API key
* rearrange `useAlgoliaSearch` to incorporate the `/search/creds` request 
* correctly render private projects in search results

## How To Test:
* while logged out, enable algolia search at  https://heady-path.glitch.me/secret
* search for `glitchdb` (a private glitch project) https://heady-path.glitch.me/search?q=glitchdb
* it should not be visible
* log in (by setting `localStorage.cachedUser` to include your persistent token)
* search for `glitchdb` again https://heady-path.glitch.me/search?q=glitchdb
* it should now be visible

## Feedback I'm looking for:
* is this correctly showing/hiding private projects?


src/components/project/featured-project.styl
```diff
@@ -10,5 +10,5 @@
   display: flex
 
 .note
-  margin-bottom: 15px
+  margin-bottom: 30px
```
clottman: Was this intentional? Or part of other changes that haven't been merged in somehow?

modernserf: i think this was someone else's hotfix

src/state/search.js
```diff
@@ -135,22 +127,46 @@ const formatAlgoliaResult = (type) => ({ hits }) =>
...
+          facetFilters: [notSafeForKids ? '' : 'notSafeForKids:false'],
+        })
+        .then(formatAlgoliaResult('project')),
+    starterKit: (query) => Promise.resolve(findStarterKits(query)),
+  };
+}
+
 export function useAlgoliaSearch(query, params = defaultParams) {
+  const api = useAPI();
+  const algoliaProvider = useMemo(() => createAlgoliaProvider(api), []);
```
clottman: How does this handle when a user was logged in, but has now logged out (on a shared computer, for example) - will the search credentials still be memoized?

modernserf: Updated so this will re-fetch creds when API changes (which happens when current user's persistent token changes)

clottman: Wouldn't you want `api` in the `[]` to make sure it gets updated only on `api` changes? `[]` will update on every re-render, which I don't think is what you want.

PR #453, opened 2019-04-23
Scrawny puffin 1: fix search results storybook

modernserf:
## Links
* Remix link https://scrawny-puffin-1.glitch.me
* Manuscript link https://glitch.manuscript.com/f/cases/3328368/

## GIF/Screenshots:
Screen Shot 2019-04-23 at 11 26 27 AM

## Changes:
* add `searchResults.starterKit` to storybook and proptypes
* move component into own directory

## How To Test:
* visit https://scrawny-puffin-1.glitch.me/storybook/index.html?path=/story/searchresults--results
* the storybook should load


src/presenters/pages/search.js
```diff
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 import Helmet from 'react-helmet';
 import { withRouter } from 'react-router-dom';
 
-import SearchResults from 'Components/search/search-results';
+import SearchResults from 'Components/search-results/search-results';
```
sheridanvk: What was the reasoning for renaming the search folder to search-results? It seems pretty granular - I was expecting that we would just have search/ and search-form.js, search-results.js would all inside of that. Do you think there's a benefit to breaking these out?

modernserf: https://glitch.manuscript.com/f/cases/3328371/Build-searchResultsCoverBar-component#BugEvent.21312572

sheridanvk: I agree with that philosophy! The directory structure is really more guiding than anything, and it's totally fair that certain things don't seem to have an obvious home. I guess here it just seems like we're adding more granularity than we need (as we now have components/search-results and components/search-form, instead of just components/search), so it might be nice to collapse into one - will leave to your judgement though.

modernserf: Well, we're already giving the components descriptive names (e.g. SearchForm instead of Form); if anything, we should give each component its own directory, but have the component live at `index.js`. The component consumer imports it with `import SearchForm from "Components/search-form"`; `components/search-form` then includes the "main" component at `index.js`, plus its stylesheets and any "inner" components (eg. Autocomplete). 

You may _also_ `import Autocomplete from "Components/search-form/autocomplete"` if you need to access an inner component, just as you may `import { BaseSearchForm } from "Components/search-form"`, but in most cases the consumer would want the default export of `index.js`.

sheridanvk: oh that's cool, I didn't realise that shorthand was possible. I'm down to refactor everything like this if you think it makes sense? It's felt kind of redundant to do things like import Button from "Components/button/button"

PR #454, opened 2019-04-23
Actually frill: storybooks for search form

modernserf:
## Links
* Remix link https://actually-frill.glitch.me
* Manuscript link https://glitch.manuscript.com/f/cases/3328369/

## GIF/Screenshots:
image

## Changes:
* move `/components/search/form` & `/components/search/autocomplete` -> `/components/search-form`
* move storybook mock data to `/stories/data`
* change search form to accommodate mocking
* create storybook for search form (incorporating autocomplete)

## How To Test:
* visit https://actually-frill.glitch.me/storybook/index.html?path=/story/searchform--results
* search for something in our mock database (e.g. "modernserf")

## Feedback I'm looking for:
* Is this an acceptable storybook?


sheridanvk:
The storybook entry for SearchForm looks great! It's a bit unexpected that it uses BaseSearchForm instead of SearchForm, but I think this is a fine tradeoff to make this show up correctly. It does more than I expected with the redirect to the actual search results page - I was expecting just a form that visually looks correct, but didn't need to redirect, especially because we have Search Results represented elsewhere. Admittedly though, having Autocomplete represented here is good so perhaps that distinction is moot. 

It's a lot of code in index.js just for this component though - what do you think about breaking it (and perhaps the other search-related stories) out from index.js into stories/search.js (like Greg has with inputs.js)? I guess it's a similar question to the component directory structure, but I think it'll make these files less overwhelming overall.

Overall though LGTM.

PR #461, opened 2019-04-24
Sweet blinker: Fix removing projects

sarahzinger:
## Links
* https://sweet-blinker.glitch.me
* https://glitch.manuscript.com/f/cases/3328335/

## GIF/Screenshots:
http://g.recordit.co/ENHzTG95uw.gif

## Changes:
* we used to overwrite `props.projects` to whatever the filteredProjects results were which was fine in the initial render, but then if we tried to filter again after a delete or a remove it was filtering with the existing filtered projects rather than the whole list. By using another variable `displayedProjects` we ensure that whenever we call filterProjects we are filtering from the full and current list.
* useEffect to re-filter anytime props.projects changes.  
* add some animation to the removeProject from team function

## How To Test:
* on a team page: add a project, filter projects, and click "remove project" 
* on user page: create a throwaway project, filter projects, and click "delete project"
* just futz around with the projects filter on any page you see it and ensure it works as expected

## Feedback I'm looking for:
* I do find this thing in general to be a little...for lack of a better word....glitchy? filtering is very smooth, like buttah. But adding/removing projects not so much. Like you can see it sort of flash a bit, there's also a bit of a delay that I don't totally understand why (I do see we debounce on filter which makes sense but why on add/remove?). If you see things we can do to make this smoother or have ideas on things I should look for let me know! Otherwise this is probably at least an improvement to the current situation?
* any other use cases I'm missing?

- [x] crossbrowser test (tested briefly in chrome/firefox/safari)

### DIFF
--- a/src/components/containers/projects-list.js
+++ b/src/components/containers/projects-list.js
@@ -17,41 +17,35 @@ import styles from './projects-list.styl';
 
 const cx = classNames.bind(styles);
 
-function ProjectsList({ title, placeholder, extraClasses, enableFiltering, enablePagination, ...props }) {
+function ProjectsList({ title, placeholder, extraClasses, enableFiltering, enablePagination, projects, ...props }) {
   const [filter, setFilter] = useState('');
   const [filteredProjects, setFilteredProjects] = useState([]);
   const [isDoneFiltering, setIsDoneFiltering] = useState(false);
 
   const validFilter = filter.length > 1;
-  let { projects } = props;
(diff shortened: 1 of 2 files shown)


oliviacpu:
`But adding/removing projects not so much.` - For removing at least there is an animation change I started but didn't finish because it was mixed up with testing stuff. I'll check out the rest of this now.

oliviacpu:
Tried this in Chrome and Firefox. Filtering is good on both, even when I'm deleting old projects. Looks :100: to me!

PR #462, opened 2019-04-24
Peppered fahrenheit - Add Delete button to Project Pages

ttseng:
## Links
* Remix link: https://peppered-fahrenheit.glitch.me/
* Manuscript link: https://glitch.manuscript.com/f/cases/3328419/

## GIF/Screenshots:
delete-project
Delete Project button launches pop-over with notice that user can always undelete a project from their profile page

image
Also updated the delete collection button to match styling

## Changes:
* Delete Project button is added to the project page underneath project Read.Me.
* In lieu of having a pop-over Component, I reused the existing `popover-with-button` pop-over.  Updated to use the Button and Emoji components as necessary.

## How To Test:
* Try deleting one of your projects from a project page.  Deleting should redirect you to your profile page.
* You should only be able to delete projects you're a member of.


oliviacpu:
I tried deleting a few of my projects from the project page.

I can see where we should be redirecting to the user profile when the request is done, but it looks like it is hanging on the project page for a minute and gets some errors from the project being deleted before it actually does the redirect.

Screenshot 2019-04-24 at 1 35 16 PM

ttseng:
ah good catch - I think it should be fixed now @oliviacpu!

src/presenters/pages/project.js
```diff
@@ -88,7 +95,75 @@ ReadmeLoader.propTypes = {
   domain: PropTypes.string.isRequired,
 };
 
-const ProjectPage = ({ project, addProjectToCollection, currentUser, isAuthorized, updateDomain, updateDescription, updatePrivate }) => {
+function DeleteProjectBtn({ projectDomain, deleteProject, currentUser }) {
```
clottman: Do we have project standards on when it's okay to use abbreviations in names? I think usually we'd go with `DeleteProjectButton` here but I could be wrong.

src/presenters/pages/project.js
```diff
@@ -88,7 +95,75 @@ ReadmeLoader.propTypes = {
...
+          <>
+            <dialog className="pop-over delete-project-pop" open>
+              <section className="pop-over-actions">
+                <div className="action-description">You can always undelete a project from your profile page.</div>
+              </section>
+              <section className="pop-over-actions danger-zone">
+                {loading ? (
+                  <Loader />
+                ) : (
+                  <Button
```
clottman: I think this button should receive focus immediately when the pop appears for accessibility reasons. That's how the Report Abuse pop works - when you click Report Abuse, your focus moves to the text input where you can enter your report, signaling to the user that there is new stuff on the screen to interact with.

styles/buttons.styl
```diff
@@ -14,7 +14,8 @@ button
   position: relative
   vertical-align: middle
   &.danger-zone
-    background-color: warning-background
+    &:hover
```
clottman: Also `&:focus`!

src/presenters/pages/project.js
ttseng: good point - I'm used to using 'btn' in css but for components it seems like we use Button more frequently, so I'll change it to Button.

src/presenters/pages/project.js
ttseng: it's common for input fields to autofocus but non-standard for buttons, especially "destructive" ones, to receive focus by default, so I'm going to keep the behavior as is.

ttseng:
made some changes if you're able to re-review / approve @oliviacpu @clottman - thanks!

src/presenters/pages/project.js
clottman: Want to set focus on the message text or title, then? According to [WAI-ARIA guidelines on dialogs](https://www.w3.org/TR/2017/NOTE-wai-aria-practices-1.1-20171214/#dialog_modal), 

> 1. When a dialog opens, focus placement depends on the nature and size of the content.
>   In all circumstances, focus moves to an element contained in the dialog.

clottman: I wouldn't say it's nonstandard for buttons to receive focus- if they're in a popup that opened due to a user action, focus needs to move somewhere inside the popup. If the button was just 'ok' or 'continue' then focus would go to the button. You're right about that being less common for destructive buttons, but we do need to move focus somewhere inside the dialog.

styles/buttons.styl
```diff
@@ -14,7 +14,8 @@ button
   position: relative
   vertical-align: middle
   &.danger-zone
-    background-color: warning-background
+    &:hover, &:focus
+      background-color: warning-background
   &:hover
```
clottman: I think maybe I was looking at this line - buttons should have `background-color: secondary-background` on both hover & focus. Right?

ttseng: the danger-zone buttons have been modified so that they aren't pink by default and turn pink on hover.  the button component already has this behavior, but PopoverWithButton (which hasn't been componentized yet), uses a regular button, so that's why I added that style here.

src/presenters/pages/project.js
ttseng: The link you shared is for modals – the closest element would be a Menu Button in this case I believe https://www.w3.org/TR/2017/NOTE-wai-aria-practices-1.1-20171214/#menubutton which still suggests that the first item be focused.

I'm still unsure about what this means for all of our pop-overs though.  For pop-overs containing forms (for example, the add-to-collection-pop), it makes sense to focus on an element by default because the form requires that a selection be made.

For our non-form pop-overs (like user-options-pop, project-options) it's not clear to me why we need to focus on the first element as per the ARIA guidelines.  I think that's potentially misleading the user to select the first element, when the options are not necessarily of equal weight.

And in this case, with the help text `You can always undelete...` - text is normally not selectable / focusable on tab right? 

Sorry for all the questions - just genuinely trying to understand the reasoning for adding a default focus here and its implications on all other pop-overs.

clottman: I followed the link to modals from [Alert and Message Dialogs](https://www.w3.org/TR/2017/NOTE-wai-aria-practices-1.1-20171214/#alertdialog), which says to use the keyboard interaction patterns defined in the modal section. 

The reason we need to focus on the first element when a pop is open is so a user knows "hey, this button did something which was to cause new elements to appear on the page". Whether the right thing to focus on is the first thing or not is open to interpretation, but whether focus needs to change is not. 

Text usually isn't focusable, but you can make it focusable by setting the `tabindex` property on the element. If we'd rather not set focus to a destructive action, I think that's the way to go.

keithk: I agree with Tiff's intuition here, we had a few of these style of popups on hcgov and I think having the destructive action be selected isn't great for user experience.

My recommendation is that we focus on the entire div when you click into it, and then your next tab is the action. That way we can alert the voiceover user that something has changed and they're in the new div and that the following action will be the destructive one.

ttseng: what do you think about handling pop-over focus in a separate PR?  I think it would be good to handle this on the `PopoverContainer` element so it affects all pop-overs and not just this one.

src/presenters/pages/project.js
```diff
@@ -88,7 +95,75 @@ ReadmeLoader.propTypes = {
...
 };
 
-const ProjectPage = ({ project, addProjectToCollection, currentUser, isAuthorized, updateDomain, updateDescription, updatePrivate }) => {
+function DeleteProjectButton({ projectDomain, deleteProject, currentUser }) {
+  const [done, setDone] = useState(false);
+  const [loading, setLoading] = useState(false);
+
+  useEffect(() => {
+    if (done) {
+      window.location = getUserLink(currentUser);
```
clottman: You might want to pass a second argument for the dependencies to this function - you'll want to declare `done` and `currentUser`.

src/presenters/project-editor.js
```diff
@@ -34,6 +34,12 @@ class ProjectEditor extends React.Component {
   async deleteProject() {
     await this.props.api.delete(`projects/${this.state.id}`);
   }
+
```
clottman: Did you mean for this to get in there?

ttseng: looks like the changes I made to incorporate the latest changes on master didn't get pushed at once - should all be there now.  This one specifically was to keep `deleteProject` consistent with all other class functions in `project-editor` (with a line space above and below)

clottman: Awesome, thanks for clearing that up!

PR #464, opened 2019-04-24
frosted-aquarius: Handle case where OAuth login fails due to no email provided

oliviacpu:
## Links
* Remix link: https://frosted-aquarius.glitch.me/
* Manuscript link: https://glitch.manuscript.com/f/cases/edit/3327833/Handle-case-where-Facebook-or-Github-login-fails-due-to-no-email-provided-by-Facebook-Github
* Specs/Notion docs

## GIF/Screenshots:
Screenshot 2019-04-24 at 12 35 52 PM

## Changes:
When we get a `403` from the API after login we know that we didn't get an email back from the provider. This handles that case, suggests login with email instead, and no longer reports the error to Sentry.

I based this off of `nova-chance` because Greg has work on the same stuff that I didn't want to conflict with. It could go to `master` but it will have more changes in the diff.

## How To Test:
Signing up with a Facebook account with no email address associated will reproduce this. I have one if you want to use it. :(

## Feedback I'm looking for:
Always looking for any feedback.

- [ ] Review the copy and UI.

### DIFF
--- a/src/presenters/pages/login.js
+++ b/src/presenters/pages/login.js
@@ -49,7 +49,7 @@ const LoginPage = ({ provider, url }) => {
 
   const [state, setState] = React.useState({ status: 'active' });
   const setDone = () => setState({ status: 'done' });
-  const setError = (message) => setState({ status: 'error', message });
+  const setError = (title, message) => setState({ status: 'error', title, message });
 
   const perform = async () => {
     try {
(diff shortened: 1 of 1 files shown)


src/presenters/pages/login.js
```diff
@@ -66,11 +66,19 @@ const LoginPage = ({ provider, url }) => {
       notifyParent({ success: true, details: { provider } });
     } catch (error) {
       const errorData = error && error.response && error.response.data;
-      setError(errorData && errorData.message);
-
-      if (error && error.response && error.response.status !== 401) {
-        console.error('Login error.', errorData);
-        captureException(error);
+      setError(undefined, errorData && errorData.message);
```
clottman: If there is no error, and so error stays as title=undefined & message = undefined or default, is it clear to the user what happened?

src/presenters/pages/login.js
```diff
@@ -66,11 +66,19 @@ const LoginPage = ({ provider, url }) => {
...
+
+      if (error && error.response) {
+        if (error.response.status === 403) {
+          // Our API returns a 403 when the login provider didn't return an email address
+          // We can suggest using email for login and avoid capturing this error in Sentry
+          const title = 'Missing Email Address';
+          const message = `${provider} didn't return an email address for your account.  Try using "Sign in with Email" instead to create an account on Glitch.`;
```
clottman: Are all of the `provider`s defined in the case you want (capitalize first letter) or do we need to enforce that programmatically?

oliviacpu: It looked like they were all capitalized but I can throw a format around it to make sure it is?

src/presenters/pages/login.js
oliviacpu: If it's undefined it will just use the message "Hard to say what happened, but we couldn't log you in. Try again?" and have the title "[Provider] Login Problem" like it did before. It should also still catch the exception if it doesn't match either a 403 or 401, so we can investigate and add a better error if other things come up.

PR #467, opened 2019-04-24
boundless-raccoon-1 letting super_users edit project names/descriptions directly

sarahzinger:
## Links
* https://boundless-raccoon-1.glitch.me/
* https://glitch.manuscript.com/f/cases/3326627/

## Changes:
* makes project page editable for super_users

## How To Test:
1. with a super_user account try to edit a project name/description that does not belong to you. Here is a test account with test project I don't care about: https://boundless-raccoon-1.glitch.me/~faithful-thistle
2. verify you can see that action in the logs along with who did that action. this can be done with the glitchdb command, for those not that great with sql queries (such as myself) we talked about how to do this here: https://fogcreek.slack.com/archives/CGFKN6HSS/p1556047277027700
3. verify that once super_user is turned off you don't have this super power
4. verify that you can still edit your own stuff fine. 

## Feedback I'm looking for:
* anything I'm missing? security/privacy concerns I didn't think about?

- [ ] test this again once I have super powers or get someone else with super powers to test

### DIFF
--- a/src/presenters/pages/project.js
+++ b/src/presenters/pages/project.js
@@ -90,6 +90,7 @@ ReadmeLoader.propTypes = {
 
 const ProjectPage = ({ project, addProjectToCollection, currentUser, isAuthorized, updateDomain, updateDescription, updatePrivate }) => {
   const { domain, users, teams } = project;
+  console.log(isAuthorized);
   return (
     <main className="project-page">
       <section id="info">
--- a/src/presenters/pages/project.js
+++ b/src/presenters/pages/project.js
@@ -181,7 +181,13 @@ const ProjectPageLoader = ({ domain, ...props }) => {
             {(currentProject, funcs, userIsMember) => (
               <>
                 <Helmet title={currentProject.domain} />
-                <ProjectPage project={currentProject} {...funcs} isAuthorized={userIsMember || currentUser.isSupport} currentUser={currentUser} {...props} />
+                <ProjectPage
+                  project={currentProject}
+                  isAuthorized={userIsMember || currentUser.isSupport}
+                  currentUser={currentUser}
+                  {...funcs}
+                  {...props}
+                />
               </>
             )}
           </ProjectEditor>
--- a/src/presenters/pages/project.js
+++ b/src/presenters/pages/project.js
@@ -90,7 +90,6 @@ ReadmeLoader.propTypes = {
 
 const ProjectPage = ({ project, addProjectToCollection, currentUser, isAuthorized, updateDomain, updateDescription, updatePrivate }) => {
   const { domain, users, teams } = project;
-  console.log(isAuthorized);
   return (
     <main className="project-page">
       <section id="info">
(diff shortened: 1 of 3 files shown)


src/presenters/pages/project.js
```diff
@@ -172,6 +172,7 @@ async function getProject(api, domain) {
 const ProjectPageLoader = ({ domain, ...props }) => {
   const api = useAPI();
   const { currentUser } = useCurrentUser();
+  const superUser = currentUser.features && currentUser.features.find((feature) => feature.name === 'super_user');
```
clottman: This should be equivalent to the value at `currentUser.isSupport`, saving you another check and a place to maintain what the feature is named.

clottman:
The case says "Allow isSupport users to edit team names and descriptions in the community site" but this PR lets support users do that just on the project pages (as far as I can tell). Do we want it on both types of pages? What about user pages, while we're at it?

sarahzinger:
@clottman switched it to isSupport checks

The original ticket was for teams, but after chatting with cori it sounds like the ultimate desire would be for super users to be able to update names/descriptions for everything (I think that's `users`, `projects`, `teams`, and `collections`, yeah?) Does that sounds in line with what you'd like to see too?

The reason I didn't do all of them is that I don't believe we have logging for each of those domains, but it sounds like we do have them for any project updates as well as user renames, so I started with projects. I'm added user renames as you suggested although I do wonder if I should wait till logging is done for the full user patch request? For example does user.rename log description updates for example?

Would you mind inviting me to the super user thing and/or testing this for me? Ideally you'd be able to update any project page and see it logged and you'd be able to update a user's name and description and see that logged as well.

sarahzinger:
ah nevermind, cori just gave me super user status and I can see none of this is working as expected lol....let me investigate for a while, will ping ya again once it seems right

src/presenters/pages/project.js
sarahzinger: @clottman k so I tried using isSupport, but realize I think we still are using the old api for user data so we aren't getting back an isSupport field we only get back a features field. This should work fine however. Once we move to the new api I'm happy to shorten this to isSupport :)

sarahzinger:
@clottman ok I tested out adding user.rename, but as far as I can tell we're only logging user.rename (and that's just the @user-name not even the display name and not the description). I think the thing to do is to just hold until we get full logging for user.patch :) So for right now I think it's good to just deploy this project.patch changes. I will edit the case description so that it more accurately portrays the overall vision :)

sarahzinger:
Just an update on this pr, I'm going to put my work on this on hold for a bit since we have some bugs with the isSupport flag and also fairly soon we're going to have greater logging features. Once those two pieces are done I'll pick this work back up.

clottman:
@sarahzinger Update to the update, logging for project.patch is almost ready! @Johnicholas has an open PR for it that is nearly complete.

sarahzinger:
sounds like we need more discussion on if this should be a thing in the community site or in a separate support tool, so I'm going to close this pr for now. But if y'all need my help let me know always happy to help!

PR #469, opened 2019-04-25
Tame whippet: Fixed jumpiness of loading user avatars on projects

ttseng:
## Links
* Remix link: https://tame-whippet.glitch.me/

## GIF/Screenshots:
There has been some weird resizing / jumpiness when projects transition from the loading placeholder to loaded user list.
jumpy
^ On prod

fixed
^ With min-height

## Changes:
* added min-height to profile-list container

### DIFF
--- a/src/components/profile-list/profile-list.styl
+++ b/src/components/profile-list/profile-list.styl
@@ -5,6 +5,7 @@
   padding: 0
   // TODO: fix margin in tooltip, remove -5px offset
   margin: 0 0 -5px
+  min-height: 37px
   
 .container + .container
   padding-top: 5px
--- a/src/components/profile-list/profile-list.styl
+++ b/src/components/profile-list/profile-list.styl
@@ -5,7 +5,6 @@
   padding: 0
   // TODO: fix margin in tooltip, remove -5px offset
   margin: 0 0 -5px
-  min-height: 37px
   
 .container + .container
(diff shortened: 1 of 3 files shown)


src/components/profile-list/profile-list.styl
```diff
@@ -5,6 +5,7 @@
   padding: 0
   // TODO: fix margin in tooltip, remove -5px offset
   margin: 0 0 -5px
+  min-height: 37px
```
sarahzinger: seems harmless but probs won't help on mobile right? should the min height be on the project items instead?

ttseng: It'll work on mobile too because we only ever show a single row of users on project items:

Jumpy behavior on prod:


Fixed with min-height
mobile-fixed

I think the underlying problem is that when the profile-list switches from the placeholder to the loaded content, the profile-list doesn't have any size, so even if we added a min-height on project items, we'd have to add some additional css to ensure the project `linkBody` is aligned to the bottom of the `projectItem` container.

modernserf:
i think the reason this isnt an issue on the search page is because the items are bottom-aligned. could we do that here?

ttseng:
at justin's suggestion, changed project-container to `align-items: end` rather than setting a fixed height on `profileList`

PR #474, opened 2019-04-25
Coordinated mascara: componentizing note

sarahzinger:
## Links
* https://coordinated-mascara.glitch.me/@sarahzinger/another-collection
* https://glitch.manuscript.com/f/cases/3328271/
* related bug fix: https://glitch.manuscript.com/f/cases/3328463/Add-style-for-MarkdownInput-to-render-with-editable-font-when-the-user-is-authorised

## GIF/Screenshots:
Screen Shot 2019-04-25 at 6 01 03 PM

## How To Test:
* check out storybook and look for the note section
* go to a collection and create a note
* go to a collection, feature a project, add a note to that featured project
* go to a collection, hide a note

## Feedback I'm looking for:
* double check there aren't any regressions here I'm missing
* I left a lot of the direct tag styles (overwriting textarea and stuff) as that's how we had it and it all just seemed to work? But is this desirable? Is there another way to do this? Should I go into these markdown inputs and give them a type of note and pass that to them maybe?
* there are probably some general style bugs with this component as well, I may or may not get to them in this pr but if you see em let me know and I'll make sure to fix them here or make a case for it. For example I know [placeholder text and underline colors are still not always enough contrast depending on the background color](https://glitch.manuscript.com/f/cases/3328432/). 

- [ ] I'd like to give these styles another double check, to make sure they are all necessary and we haven't missed anything. 
- [ ] crossbrowser test


src/components/collection/note.js
```diff
@@ -5,16 +5,20 @@ import _ from 'lodash';
 
 import { ProfileItem } from 'Components/profile-list';
 
-// TODO: let's move these into components
-import AuthDescription from './includes/auth-description';
+import styles from './note.styl';
 
-import { isDarkColor } from '../models/collection';
+import { isDarkColor } from '../../models/collection';
```
ttseng: this can now be from `Models/collection`

ttseng:
there was a design regression introduced at some point where the markdown component displays `sans-serif` font in its default state, even when the user is able to edit its content.

markdown-input

(notice difference between title and description)

I'm wondering if we can reuse the behavior that @GregWeil added with textarea input fields to solve this problem? https://coordinated-mascara.glitch.me/storybook/index.html?path=/story/input-fields--generic

ttseng:
Some CSS suggestions for spacing between the note + project entities:

in `featured-project.styl`
```
.note
   margin-bottom:  45px
```

In `note.styl`
```
margin-bottom: -30px
````

That gives us
image

File: src/components/collection/note.js [hunk 2]
```diff
@@ -5,16 +5,20 @@ import _ from 'lodash';
...
+
+import AuthDescription from '../../presenters/includes/auth-description';
+
+const cx = classNames.bind(styles);
 
 /**
  * Note Component
  */
 const Note = ({ collection, project, updateNote, hideNote, isAuthorized }) => {
```
sheridanvk: This is something that I'm curious to get @modernserf's take on - is there a more optimal way to do this now that we have Redux and Hooks? Like should we be determining whether or not the current user is authorised based on a callout to state from the component (where we could store all the collections this user is an owner of), or should we continue to pass props like updateNote and isAuthorized down from parent components? I get the impression that it would be easier to understand our components if we had the updateNote/hideNote functions defined in the actual Note component, but maybe I'm missing some downsides to this?

src/components/collection/note.styl
```diff
@@ -0,0 +1,46 @@
+.note
+  margin-bottom: -15px
+  width: 95%
```
sheridanvk: is this something that can be defined in the parent component where the note is being used, rather than hardcoding it into the note itself?

sheridanvk:
Seeing some React errors on https://coordinated-mascara.glitch.me/@sarahzinger/another-collection

Screenshot 2019-04-30 at 10 41 02 AM

src/components/collection/note.js
modernserf: Short answer: yes we should do this eventually, but that's well beyond the scope of this PR

src/components/collection/note.styl
```diff
@@ -0,0 +1,46 @@
...
+  
+.descriptionContainer
+  width: calc(100% - 36px)
```
modernserf: instead of `calc` here, you could put `padding-right: 36px` on `.note`

sheridanvk:
For the markdown component, it's not clear to me why we need all these different ways of rendering markdown in different places - e.g. I noticed in CollectionHeader that we have no underline or different colour on links, so they're pretty undiscoverable. I think we should stop overriding these tags on a one-off basis, but we don't have to block this PR on it - I created a task to determine whether there are different Markdown modes that we need, and we can create those instead of overriding CSS with tags on an individual basis: https://glitch.manuscript.com/f/cases/3328455/Work-out-what-Markdown-modes-we-need-so-we-can-stop-overriding-styles-with-selectors-on-the-HTML-tags

modernserf:
@sheridanvk  the react errors are unrelated, this PR just needs to merge in master

src/components/collection/note.js
sheridanvk: 👍 ok sounds good!

src/components/collection/note.styl
sarahzinger: hmm, that isn't having the same effect? maybe I'm missing something else but we need to specify that the width of note is (close to 100% of it's container) or only expands to be the width of the text, and if we say it's 100% it pushes the avatar down

Screen Shot 2019-04-30 at 4 19 47 PM

Screen Shot 2019-04-30 at 4 16 02 PM

modernserf: oh i see- the user is floated, not absolutely positioned. I guess you could lay this out with flexbox then?

modernserf: But don't let this block you if you're ready to merge

sarahzinger:
@ttseng I ended up adding the margin-bottoms to the components that wrap the notes, hopefully it adequately captures it. I believe I have fixed the font issue, but feel free to double check since my eyes are terrible

@sheridanvk good thoughts! Can confirm I haven't seen that error since pulling in master, also I moved the margins/widths of note to their container components

@modernserf good call adding in flex! just did so I believe it looks good

PR #475, opened 2019-04-25
Hail locket: convert user & team pages to use css modules

modernserf:
## Links
* Remix link: https://hail-locket.glitch.me/

## Changes
* replace styles for user & team page with CSS modules
* move un-modularized styles (eg. project containers, team members, analytics, popovers) into `:global` scope of css modules
* delete unused styles

## How To Test
* verify user & team pages look & work correctly

## Feedback I'm looking for
* is this a reasonable strategy for modularizing CSS without going "full component"?


src/components/data-loader/index.js
```diff
@@ -28,6 +28,6 @@ DataLoader.propTypes = {
 };
 DataLoader.defaultProps = {
   renderError: () => 'Something went wrong, try refreshing?',
-  renderLoader: () => Loader,
+  renderLoader: () => <Loader />,
```
sheridanvk: Could it make sense to storybook this component with basic non-default loader, to make it clear what the difference between this and the Loader component are? And/or should we rename Loader to LoaderAnimation, as that's really all it is?

src/presenters/pages/team.styl
```diff
@@ -0,0 +1,110 @@
+@import "../../components/constants"
```
sheridanvk: could use aliased path here: 'Components/constants'

src/presenters/pages/team.js
```diff
@@ -60,6 +60,25 @@ const TeamPageCollections = ({ collections, team, currentUser, currentUserIsOnTe
...
+    <div>
+      <Heading tagName="h4">Teams are in beta</Heading>
+      <Text>Learn More</Text>
+    </div>
+  </a>
+);
+
+const ProjectPals = () => (
+  <aside className="inline-banners add-project-to-empty-team-banner">
+    <div className="description-container">
```
sheridanvk: So would the idea for this part be to move the styles from styles/includes/inline-banners.styl into team.styl in a future PR, and then switch out these classnames here with styles. classnames?

sheridanvk:
Noticed a regression in the grid for recent projects relative to production
Screenshot 2019-04-26 at 3 20 19 PM

src/components/data-loader/index.js
modernserf: This is a bugfix for a different PR

sheridanvk:
This approach seems good overall - I'm wondering if we should just try to do this without :global though, as otherwise we'll have to do another pass on these down the line to have full clarity on where the CSS is being used. I tried taking :global out in a remix (https://remix-hail-locket.glitch.me/@glitch), and that worked pretty well tbh - the only issue I see is a messed up margin underneath Recent Projects. I wonder if we could just try doing that and see whether it's close enough to just fix the difference, as that way we'd know for sure that these styles were only used here?

And actually I think that bottom margin needs to be in the ProjectList component (between the projects container and the pagination bar), which apparently doesn't have a storybook entry - maybe we would have caught it there.

What do you think?

sheridanvk:
Here's what the grid looks like in prod - the project elements are always the same height as each other:
Screenshot 2019-04-26 at 3 41 49 PM
Screenshot 2019-04-26 at 3 42 09 PM

sheridanvk:
And @modernserf pointed out to me that I was missing some of the :global selectors - we definitely need them for now!

src/presenters/pages/team.styl
modernserf: are you sure about that?

modernserf:
Updated with correct medium-width layout
Screen Shot 2019-04-29 at 10 17 53 AM

src/presenters/pages/team.styl
sheridanvk: ahhh I seem to have only added the aliases to .js in webpack:
  resolve: {
    extensions: ['.js'],
    alias: aliases,
  }
we have to add .styl there for this to work (and possibly we have to add this to .eslintrc.js also). I can file a ticket on myself to do that, or you could add it to this PR if you want.

PR #477, opened 2019-04-26
Quixotic rocket: Use the rest api in current-user.js

GregWeil:
## Links
https://glitch.fogbugz.com/f/cases/3327984/Use-the-new-api-for-user-management

## Changes:
Gets rid of some calls to the old api and replaces them with equivalent rest api calls

## How To Test:
Try signing in and out, and use different pages on the site to make sure they aren't broken from a new data layout or something

### DIFF
--- a/src/state/current-user.js
+++ b/src/state/current-user.js
@@ -129,11 +129,13 @@ class CurrentUserManager extends React.Component {
   }
 
   async getSharedUser() {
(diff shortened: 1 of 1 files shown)


oliviacpu:
👀

oliviacpu:
I am getting that same warning on another remix now so this is probably fine.

PR #484, opened 2019-04-29
Cuddly dormouse: componentize collection lists

modernserf:
## Links
* Remix link: https://cuddly-dormouse.glitch.me
* Manuscript link: https://glitch.manuscript.com/f/cases/3328387/

## GIF/Screenshots
Team with empty collections https://cuddly-dormouse.glitch.me/@community
Screen Shot 2019-04-29 at 5 44 39 PM

Team has collections https://cuddly-dormouse.glitch.me/@glitch
Screen Shot 2019-04-29 at 5 45 07 PM

More from collection curator https://cuddly-dormouse.glitch.me/@glitch/glitch-this-week-april-24-2019
Screen Shot 2019-04-29 at 5 45 24 PM

Included in collections https://cuddly-dormouse.glitch.me/~csv-name-randomizer
Screen Shot 2019-04-29 at 5 46 39 PM

## Changes:
* componentize CollectionList
* componentize MoreCollections **(now using CollectionItemSmall)**
* inline IncludedInCollections into project page
* add `showCurator` to CollectionItemSmall
* delete unused CSS

## How To Test:
* visit https://cuddly-dormouse.glitch.me and log in
* verify collection lists render correctly on the above pages


src/presenters/pages/project.js
```diff
@@ -32,6 +33,35 @@ function syncPageToDomain(domain) {
   history.replaceState(null, null, `/~${domain}`);
 }
 
+const useIncludingCollections = createAPIHook(async (api, projectId) => {
```
sarahzinger: why useIncludingCollections instead of useIncludedInCollections btw, is that a convention for hooks I should learn or just a typo?

modernserf: no, the name was arbitrary, not sure why i chose it

sarahzinger:
Overall this looks really great to me! The only thing that looks a little off to me is the different heights on the collection items depending on the description on the `More from collection curator` that you highlight in your screenshot. Like I find it odd that this description ends in an ellipses but then isn't the full height in comparison to the others

Screen Shot 2019-04-30 at 11 05 34 AM

But I wonder if that's beyond the scope of this pr? If it is maybe we should make a case to think about these things later.

src/components/collections-list/styles.styl
```diff
@@ -0,0 +1,39 @@
...
+
+.moreByLinkWrap a
+  color: primary
+  text-decoration: none
+  &:hover
+    text-decoration: underline
+
+.collectionsContainer
+  display: grid
+  grid-template-columns: repeat(auto-fill, minmax(300px, 1fr))
```
ttseng: Can we handle the grid similar to how we show related projects (show 3 collections in a single row with `grid-template-columns: 1fr 1fr 1fr` for min-width 670px)?  Then we avoid having a single collection orphaned for the included in collections section. https://cuddly-dormouse.glitch.me/~community
image

modernserf: @ttseng updated

ttseng:
oops looks like the `collectionsContainer` styling was removed from `collections-list/styles.styl` so collections on team and user profiles are not appearing correctly: 
image

ttseng:
maybe related to the comment above - any reason to use `styles.styl` instead of `collections-list.styl`? 

did we want to standardize on our naming more generally (`collections-list.js` vs `collections-list/index.js` and `collections-list.styl` vs `collections-list/styles.styl`)?

I see how choosing the naming convention `styles.styl` makes it easier to find the file you want if you're searching by component name, so that seems like a promising direction – it would be helpful for us to establish that as a general rule going forward so other components are designed that way / we update the existing components.

modernserf:
Updates: 
- fixed broken collections container
- collection components now stretch to fill their container 
Screen Shot 2019-05-02 at 3 23 08 PM
Screen Shot 2019-05-02 at 4 28 00 PM
- collection projects use the same flexible column/row behavior as other groups of three
Screen Shot 2019-05-02 at 3 23 31 PM

src/components/containers/row.styl
```diff
@@ -0,0 +1,22 @@
+// http://www.heydonworks.com/article/the-flexbox-holy-albatross-reincarnated
+.row
+  list-style-type: none
+  margin: 0
+  padding: 0
+  display: flex
+  flex-wrap: wrap
+  --gap: 10px
+  --min-width: 250px
```
ttseng: with this min-width, we're ending up with really tall recent project + collections on the homepage, even at 840px width (half of my 15" macbook pro)

image

image

can we set this to 180px instead?  then the rows don't stack until mediumViewport size.

src/components/collection/collection-item.js
```diff
@@ -62,46 +61,44 @@ ProjectsPreview.propTypes = {
 };
 
 const CollectionItem = ({ collection, deleteCollection, isAuthorized, showCurator }) => (
-  <div className={classNames(styles.collectionItem, { [styles.authorized]: isAuthorized })}>
-    <div>
+  <div className={styles.collectionItem}>
+    <div className={styles.header}>
```
ttseng: Should probably just show the header conditionally based on whether showCurator || isAuthorized - otherwise, because of the fixed flex-basis in `collection.item.styl` `.header`, there is extra spacing between collection items when viewing someone else's profile because of the fixed height: https://cuddly-dormouse.glitch.me/@sarahzinger
image

(should be):
image

src/components/containers/projects-list.styl
```diff
@@ -3,13 +3,16 @@
...
-  margin: 0
-  padding: 0
-  // 
+.projectsGrid
+  --row-gap: 20px
+  --min-width: 350px
+  
+.projectsGridCompact
+  --row-gap: 20px
+  --min-width: 250px
```
ttseng: I think we need `align-items: end` here or else collections with notes don't render correctly:

image

(should be): 
image

Try at https://cuddly-dormouse.glitch.me/@sarahzinger/another-collection

ttseng:
I really like that your changes mean that empty collection-items don't take up unnecessary space vertically if they're on their own row 👍
image

ttseng:
in `icons.styl` there is an unnecessary `margin-left: 5px` which is making the dropdown for project-options and collection-options not square:

image

I don't see the extra left margin on prod, but I also don't see that you made a change to icons.styl in this PR...

src/components/containers/row.styl
```diff
@@ -0,0 +1,22 @@
...
+  --modifier: calc((var(--min-width) * var(--item-count)) + (var(--gap) * (var(--item-count) - 1)) - 100%)
+  margin: calc(var(--gap) * -0.5)
+
+.item,
+.filler
+  flex-grow: 1
+  flex-basis: calc(var(--modifier) * 999)
```
ttseng: read the blog post you linked to - maybe we can include a comment inline here to summarize what's happening? (force horizontal layout to vertical layout at container query breakpoint (defined by modifier))

PR #487, opened 2019-04-30
Superb guide: Sort projects locally instead of using the api sort

GregWeil:
## Links
https://glitch.manuscript.com/f/cases/3328449/odd-behavior-on-the-community-site

## Changes:
Stop using orderKey=userLastAccess and do the sort locally. Old projects (seemingly before March 2019, so not that old) don't have userLastAccess set, so it ends up looping the requests once it tries to make a request with `?lastOrderValue=&`. This sorts locally, using the lastAccess field for projects without a userLastAccess field. (Projects with userLastAccess are always put before projects without it)

### DIFF
--- a/src/models/project.js
+++ b/src/models/project.js
@@ -25,13 +25,15 @@ export function getRemixUrl(domain, editorUrl = EDITOR_URL) {
   return `${editorUrl}#!/remix/${domain}`;
 }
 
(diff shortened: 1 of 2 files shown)


GregWeil:
I tested with a lower number of projects loaded per request 👍

PR #488, opened 2019-04-30
High cat: Pop-over Focus

ttseng:
This PR adds some default focusing behavior for when pop-overs are opened from keyboard input.  

The expected behavior is:
* Automatically focusing on the pop-over / pop-over contents only happens when the pop-over is triggered from keyboard input _not_ from mouse input (following ARIA guidelines here: https://www.w3.org/TR/2017/NOTE-wai-aria-bractices-1.1-20171214/examples/menu-button/menu-button-links.html)
* We use default styling for focusing
* We default to focusing on the first selectable element in the pop-over, defined in `focusFirstElement` in `popover-container.js`, which includes these elements:
```
a:not([disabled]),
button:not([disabled]),
input:not([disabled]), 
[tabindex]:not([disabled]):not([tabindex="-1"]), 
select:not([disabled]);
```

* When the only options within a pop-over are destructible (for example, deleting a team or deleting a collection from collection-options-pop), focus on the dialog rather than the destructible button.  This is made possible by adding `tabIndex=0` to the pop-over's dialog element.

## Links
* Remix link: https://high-cat.glitch.me/

## GIF/Screenshots:
keyboard-focus
Example behavior with user-options-pop

## Changes:
* Added pop-over focusing to all popover-with-button items. 
* Added exception to eslintrc.jc to enable focusing on dialog component

## How To Test:
You can test these pop-overs with voice-over
**Any page:**
* new-project-pop
* user-options-pop
* sign-in-pop

**Project Page:**
* delete-project-pop - **note**: focuses on dialog

**Profile Page**
* project-options-pop
* collection-options-pop
* feature-project-options-pop

**Collection Page**
* edit-collection-color-pop

**Team Page**
* team-user-info-pop
* analytics-time-pop
* delete-team-pop - **note**: focuses on dialog

Note that these pop-overs were not edited as they have filter input fields, and those automatically receive focus:
* add-collection-project-pop 
* analytics-projects-pop 
* add-project-to-collection-pop


clottman:
Focus styling should be visible by default - it helps users who are navigating via the keyboard and can see. Otherwise those users may feel a bit lost when focus moves but now they don't know exactly what it's on.

src/presenters/pop-overs/create-team-pop.js
```diff
@@ -161,11 +161,11 @@ CreateTeamPopBase.propTypes = {
 
 const CreateTeamPop = withRouter(CreateTeamPopBase);
 
-const CreateTeamPopOrSignIn = () => {
+const CreateTeamPopOrSignIn = ({ ...props }) => {
```
clottman: I think there's an extra layer of spreading we don't need here. 

Does something like this work? 

```
const CreateTeamPopOrSignIn = (props) => {
... 
<CreateTeamPop api={api} {...props} />
```

src/presenters/pop-overs/new-project-pop.js
```diff
@@ -91,7 +92,7 @@ function NewProjectPopButton() {
 
   return (
     <PopoverWithButton buttonClass="button-small" dataTrack="open new-project pop" buttonText="New Project">
-      {() => <NewProjectPop projects={projects} />}
+      {({ focusDialog }) => <NewProjectPop projects={projects} {...{ focusDialog }} />}
```
clottman: This is maybe just personal preference, but for me, it's harder to understand what `{...{ focusDialog }` is doing. 
Can we use `focusDialog={focusDialog}` instead? It's the same thing but feels easier to read & quickly grok.

src/presenters/pop-overs/popover-container.js
```diff
@@ -31,6 +31,10 @@ const PopoverContainer = ({ children, onOpen, outer, startOpen }) => {
   };
   const togglePopover = () => setVisible(!visible);
 
+  const focusDialog = (element) => {
+    if (element) element.focus();
```
clottman: I think in general, we want focus to move to the first naturally focusable element within the popover, and only fall back to focusing the `dialog` if there are no non-destructive focusable elements.

ttseng: I liked how focusing on the dialog leads the screenreader to read all potential options within the menu so you get an overview before making a selection, but if the standard is to focus on the first menu item, then it makes sense to align to that.

Based on the behavior of the menu button here: https://www.w3.org/TR/2017/NOTE-wai-aria-bractices-1.1-20171214/examples/menu-button/menu-button-links.html

It seems like the rules are:
1. if user opens menu via keyboard navigation (enter key) on button, open the menu and focus on the first menu item
2. however, if user opens menu by mouse (clicking the button), we don't automatically focus on any elements within the menu (the button is still in focus)

Not auto-focusing for Rule 2 makes the dialogs less confusing to me because it doesn't make it seem like we're recommending a particular option by default for those navigating the site without a screen reader.

clottman: That approach makes sense to me.

ttseng: initial exploration into this, using `user-options-pop` as an example

* can determine whether or not `togglePopover` is called by keyboard or mouse event by checking the `event.detail`:

```
  const togglePopover = (event) => { 
    setVisible(!visible); 
    if(event.detail === 0 ){
      // pop-over opened by keyboard event - focus on first element of pop-over here
    }
  };
```

attempt to focus on first tabbable child of pop-over:
1. use ref callback?  -> cannot use ref on function components (for example `UserLink`) https://github.com/facebook/react/issues/4936
could potentially a) pass ref down to function component definition, b) turn function component into class component, but both options don't seem desirable (would need to do on a case-by-case basis)

2. simulate a tab keyboard event when the pop-over is opened?
-> no, keyboardEvent.initKeyEvent is deprecated and not recommended on production sites
https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/initKeyEvent

3. ? other ideas TBD

clottman: For 1 - you can use the `useRef` hook to have a ref in a function component. https://reactjs.org/docs/hooks-reference.html#useref

src/presenters/pop-overs/popover-container.js
```diff
@@ -25,11 +25,35 @@ const MonitoredComponent = onClickOutside(UnmonitoredComponent);
...
   const [visible, setVisibleState] = React.useState(startOpen);
+  const [openedFromKeyboard, setOpenedFromKeyboard] = React.useState(false);
+  
   const setVisible = (newVisible) => {
     if (!visible && newVisible && onOpen) onOpen();
     setVisibleState(newVisible);
   };
-  const togglePopover = () => setVisible(!visible);
+  
+  const focusFirstElement = (dialog) => {
```
ttseng: Created this function that's called when the dialog is opened using Refs.  
It focuses the first selectable element in a pop-over if the pop-over was opened via keyboard input.  

I added it to `user-options-pop` to show an example but not to the other pop-overs just yet.  If this approach looks good, I'll add it to the other pop-overs that don't have a single destructable element (in that case, my plan is to focus on the dialog itself).

@clottman let me know what you think – thanks!

ttseng: here's a preview of what it looks like in practice with the `user-options-pop`

keyboard-focus

clottman: I think the general approach looks good! 

Minor note, I think we want all input types (not just text) - so checkboxes, color, etc. Also, `textarea` and `select` when not disabled.

ttseng:
I believe this now has all the pop-over focusing behavior that's expected.  I updated the original PR comment to add a description of the expected behavior and an updated list of everywhere the pop-over focusing is used.  Let me know if you have any questions!

src/presenters/pop-overs/new-project-pop.js
```diff
@@ -31,8 +31,8 @@ const NewProjectResultItem = ({ id, domain, description }) => (
   </div>
 );
 
-const NewProjectPop = ({ projects, opensFromRight }) => (
-  <div className={`pop-over new-project-pop ${opensFromRight ? 'opens-right' : ''}`}>
+const NewProjectPop = ({ projects, focusFirstElement, opensFromRight }) => (
+  <dialog className={`pop-over new-project-pop ${opensFromRight ? 'opens-right' : ''}`} ref={focusFirstElement}>
```
clottman: Loving the switch from `div` to `dialog` for some of these, because it makes our HTML more semantic!

src/presenters/pop-overs/popover-container.js
```diff
@@ -25,11 +25,40 @@ const MonitoredComponent = onClickOutside(UnmonitoredComponent);
...
   const setVisible = (newVisible) => {
     if (!visible && newVisible && onOpen) onOpen();
     setVisibleState(newVisible);
   };
-  const togglePopover = () => setVisible(!visible);
+
+  const focusFirstElement = (dialog) => {
+    // only focus to next selectable element in dialog if popover is triggered from keyboard
+    if (dialog && openedFromKeyboard) {
+      // focus on the dialog if it has tabIndex=0 (used when there is only a destructible item in the popover that shouldn't automatically be focused on
```
clottman: very minor - comment starts a ( but doesn't end it with )

src/presenters/pop-overs/popover-container.js
```diff
@@ -25,11 +25,40 @@ const MonitoredComponent = onClickOutside(UnmonitoredComponent);
...
+      if (dialog.tabIndex === 0) {
+        dialog.focus();
+      } else {
+        const focusableElements =
+          'a:not([disabled]), button:not([disabled]), input:not([disabled]), [tabindex]:not([disabled]):not([tabindex="-1"]), select:not([disabled])';
```
clottman: Still missing `textarea` I think.

ttseng:
thanks for all the feedback and helping me think this through @clottman !

@modernserf I plan on merging this so that all the pop-overs will be accessible in the interim as the existing pop-overs get switched over to your new component (with the focus handling baked in)

PR #490, opened 2019-04-30
Helix tyrannosaurus: Webpack space issues

keithk:
##
Editor: https://glitch.com/edit/#!/helix-tyrannosaurus?path=webpack.config.js:120:11
Link: https://helix-tyrannosaurus.glitch.me

## Changes:
Verbose output of what files are being removed. Does not remove stats.json

## How To Test:
Make changes and watch webpack rebuild

## Feedback I'm looking for:
Are there any glitches or errors? Make changes in the project and watch webpack rebuild and access the site while it's happening.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -150,7 +150,7 @@ module.exports = smp.wrap({
       hash: true,
       publicPath: true,
     }),
-    new CleanWebpackPlugin({ verbose: true, cleanOnceBeforeBuildPatterns: ['!storybook/*', '!stats.json']}),
+    new CleanWebpackPlugin({ dry:true, cleanStaleWebpackAssets: true, verbose: true, cleanOnceBeforeBuildPatterns: ['**/*', '!storybook/*', '!stats.json']}),
   ],
   watchOptions: {
     ignored: /node_modules/,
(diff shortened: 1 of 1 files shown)


webpack.config.js
```diff
@@ -150,7 +150,7 @@ module.exports = smp.wrap({
       hash: true,
       publicPath: true,
     }),
-    new CleanWebpackPlugin({ verbose: true, cleanOnceBeforeBuildPatterns: ['!storybook/*', '!stats.json']}),
+    new CleanWebpackPlugin({ dry: false, cleanStaleWebpackAssets: true, verbose: true, cleanOnceBeforeBuildPatterns: ['**/*', '!storybook/*', '!stats.json']}),
```
taravancil: Do we need `dry: false` if the default is `false`? I'm in favor of being verbose for the sake of clarity, but just wanted to double check that was the intention.

taravancil:
I wonder if there's a way to make this not use warn colors in the console?
Screen Shot 2019-05-01 at 10 13 26 AM

webpack.config.js
keithk: We don't need it - I had it there so it's clear you can toggle it, but doesn't need to be

PR #491, opened 2019-05-01
Superb guide: Throw an error instead of going into an api request loop

GregWeil:
https://superb-guide.glitch.me/ (but there's nothing to see there)

The other option is to silently log it and just return whatever we've downloaded so far, but now we're throwing around incomplete data.

### DIFF
--- a/shared/api.js
+++ b/shared/api.js
@@ -30,12 +30,12 @@ const getAllPages = async (api, url) => {
   while (hasMore) {
     const data = await getFromApi(api, url);
     results.push(...data.items);
-    hasMore = data.hasMore;
-    url = data.nextPage;
-    if (hasMore && !data.lastOrderValue) {
+    if (data.hasMore && !data.lastOrderValue) {
       console.warn('hasMore', data.hasMore, 'lastOrderValue', data.lastOrderValue);
       throw new Error('The rest api responded with hasMore but no lastOrderValue');
     }
+    hasMore = data.hasMore;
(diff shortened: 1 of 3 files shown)


shared/api.js
```diff
@@ -30,6 +30,10 @@ const getAllPages = async (api, url) => {
   while (hasMore) {
     const data = await getFromApi(api, url);
     results.push(...data.items);
+    if (data.hasMore && !data.lastOrderValue) {
+      console.warn('hasMore', data.hasMore, 'lastOrderValue', data.lastOrderValue);
+      throw new Error('The rest api responded with hasMore but no lastOrderValue');
```
sarahzinger: what does the homepage look like in this state? I'm guessing it looks generally ok just with partially loaded data yeah? That sounds appropriate to me, no need to show an error if most of the projects they would want to look at are loaded?

Also do we want to console log on the frontend like this? Maybe better to use captureBreadcrumb https://docs.sentry.io/clients/node/usage/#raven-recording-breadcrumbs or something like that?

GregWeil: The homepage recent projects will get stuck swirling, and your teams/projects/collections won't show in pops.

I didn't use sentry because I didn't know how to properly pick @sentry/node vs @sentry/browser since this code is used in both places, but after some research it looks like the answer is @sentry/minimal

GregWeil:
Updated to report to sentry and return what's been loaded so far.

PR #493, opened 2019-05-01
Quixotic rocket: Some overlay components

GregWeil:
## Links
https://glitch.fogbugz.com/f/cases/3328465/Componentize-overlays
https://quixotic-rocket.glitch.me/storybook/index.html?path=/story/overlays--generic

## GIF/Screenshots:
image
image

## Changes:
- Three new components: Overlay, OverlaySection, and OverlayTitle.
- The only difference between these components and popovers are the outer border on Overlay and default positioning/sizing, which could be covered with a type prop.
- There's a className prop on the overlays, which is used for the video and new-stuff to set a max-width and style some inner/not yet componentized stuff.

## How To Test:
Check out the how it works video, new stuff pop, and the storybook entries.


GregWeil:
The OverlaySection and OverlayTitle components are only meant to be used in the context of an Overlay. Is there any way to enforce/prop type check that?

oliviacpu:
> The OverlaySection and OverlayTitle components are only meant to be used in the context of an Overlay. Is there any way to enforce/prop type check that?

If you passed a `title` and `sections` to the `<Overlay />` component, and had it be responsible for rendering those when appropriate. You could enforce specific props that way, and have the other components not be exported. I'm not sure if that's a pattern we're using though.

oliviacpu:
@GregWeil I am visiting the storybook but it isn't loading for me. I get `Page not found`.

GregWeil:
Oh whoops, I did a `git clean` when I merged in master, and forgot to rebuild storybook

I don't think we're following that pattern much right now

GregWeil:
Those styles can't get removed yet, new stuff still needs them

PR #494, opened 2019-05-01
Miniature ferry: lots of style fixes

sarahzinger:
## Links
* https://miniature-ferry.glitch.me/
* https://glitch.manuscript.com/f/cases/3328426/
* https://glitch.manuscript.com/f/cases/3328468/
* https://glitch.manuscript.com/f/cases/3328464/

## GIF/Screenshots and Changes:
* fixes view -> bug
Screen Shot 2019-05-01 at 5 33 41 PM

* fixes dropdown behavior for new project pop over (arrow where it should be, popover under the button, popover opens from right/left depending on if you're signed in so it doesn't float off the page):
Screen Shot 2019-05-01 at 5 37 00 PM
Screen Shot 2019-05-02 at 9 20 38 AM
Screen Shot 2019-05-02 at 9 22 01 AM

* fixes dropdown placement in general
Screen Shot 2019-05-01 at 5 40 30 PM
Screen Shot 2019-05-01 at 5 40 45 PM
Screen Shot 2019-05-01 at 5 41 05 PM
Screen Shot 2019-05-01 at 5 41 51 PM
Screen Shot 2019-05-01 at 5 44 02 PM

## How To Test:
* to test view -> bug try to add a project to a collection on safari
* look at glitch on a mobile phone or mimic it with the dev tools try to create a new project when signed in/not signed in see if things look right
* test dropdowns on user/team/collection/search pages

## Feedback I'm looking for:
anything I'm missing? See better ways to do what I'm doing that I didn't think about?

- [ ] I mainly tested on safari and chrome, will give this a quick look on edge as well

### DIFF
--- a/src/components/header/header.styl
+++ b/src/components/header/header.styl
@@ -35,6 +35,7 @@
   
 .logoWrap
   flex: 0 0 70px
+  margin-right: 10px
 
 .logo
   width: 100%
--- a/src/components/header/header.styl
+++ b/src/components/header/header.styl
@@ -8,6 +8,7 @@
       display: inline-block
       padding-right: 5px
       border: none
+      min-width: 48px
       img
         width: 28px
         height: auto
--- a/src/components/header/index.js
+++ b/src/components/header/index.js
@@ -35,9 +35,6 @@ const Header = ({ searchQuery, showNewStuffOverlay }) => {
           <li className={styles.buttonWrap}>
             <NewProjectPop />
           </li>
-           <li className={styles.buttonWrap}>
-              <ResumeCoding />
-            </li>
           {!!currentUser && !!currentUser.projects.length && (
             <li className={styles.buttonWrap}>
               <ResumeCoding />
--- a/src/components/header/index.js
+++ b/src/components/header/index.js
@@ -21,6 +21,7 @@ const ResumeCoding = () => (
 
 const Header = ({ searchQuery, showNewStuffOverlay }) => {
   const { currentUser, clear } = useCurrentUser();
+  const showResumeCoding = !!currentUser && !!currentUser.projects.length
   return (
     <header role="banner" className={styles.header}>
       <Link to="/" className={styles.logoWrap}>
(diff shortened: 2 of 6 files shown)


src/components/header/header.styl
```diff
@@ -8,6 +8,7 @@
       display: inline-block
       padding-right: 5px
       border: none
+      min-width: 48px
```
sarahzinger: this fixes the `view ->` button bug

src/components/header/index.js
```diff
@@ -33,7 +33,7 @@ const Header = ({ searchQuery, showNewStuffOverlay }) => {
         </div>
         <ul className={styles.buttons}>
           <li className={styles.buttonWrap}>
-            <NewProjectPop />
+            <NewProjectPop opensFromRight={!(currentUser && currentUser.login)} />
```
sarahzinger: we'll always open from the right if the user is not logged in because there's a lot more space in the header

styles/results-list.styl
```diff
@@ -46,6 +46,7 @@
       right: 1em
       top: 1em
       z-index: 1
+      width: 65px
```
sarahzinger: this is also for the "view ->" button

styles/pop-overs.styl
```diff
@@ -26,7 +26,6 @@ popover-secondary-background()
   cursor: initial
   display: block
   position: absolute
-  top: 55px
```
sarahzinger: all of the changes here are to ensure the popover dialogs are not on top of the dropdowns buttons, and to ensure it only opens from the right if it has that class

src/components/header/header.styl
sheridanvk: Chatting with Sarah this resolves the dropdown caret next to the user signin button

ttseng: nice!

ttseng:
awesome - thanks for id-ing and documenting these changes so thoroughly. 

while you're making these modifications, do you mind also adding this small change?

in `src/components/header/header.styl` add margin-right: 10px to add some space between the logo + search bar on small screens
```
.logoWrap
    flex: 0 0 70px
    margin-right: 10px
```
image

sarahzinger:
@sheridanvk looked into the view -> width thing and realized that the issue is not with the wrapping but that we use `width: fit-content` in our generic popover styles which renders differently depending on safari vs chrome and there's no need for it on this particular thing, so I reset with `width: auto` which fixed it. 

also made a case and updated an existing one with the other thoughts we talked about!
Avatar Dropdown Component Case: https://glitch.manuscript.com/f/cases/3328478/Avatar-dropdown-Component
Added suggestions to the popover component case: https://glitch.manuscript.com/f/cases/3328381/Build-Popover-component

PR #495, opened 2019-05-02
Coconut restaurant: anonymous users don't see project item dropdown

sarahzinger:
## Links
* https://coconut-restaurant.glitch.me/
* https://glitch.manuscript.com/f/cases/3328445/

## Changes:
* removes the project options dropdown on project items for anonymous users

## How To Test:
* go to https://coconut-restaurant.glitch.me/@sarahzinger notice you can't add any of my projects to collections anymore

## Feedback I'm looking for:
- this seemed like the quickest fix, but it does remove some functionality for anonymous users. Right now it looks like anonymous users can feature/pin/delete a project from their anonymous user page. If I deploy this as is they won't be able to feature/ping a project until they get a login. They can still delete the project from their project page though. I think this makes sense but I could see an argument for making this logic more complicated if we want where we keep that functionality. The thing is projects made by anonymous users I think delete after 2 weeks or something like that? So why are you pinning/featuring projects at all? I guess I could see an argument that this lets you test out the features...? I guess my inclination is to remove it just to simplify, but if we think testing those features out is useful I can definitely change my approach.

### DIFF
--- a/src/presenters/pop-overs/project-options-pop.js
+++ b/src/presenters/pop-overs/project-options-pop.js
@@ -92,7 +92,6 @@ const ProjectOptionsContent = ({ addToCollectionPopover, ...props }) => {
   const onClickLeaveTeamProject = useTrackedFunc(leaveTeamProject, 'Leave Project clicked');
   const onClickLeaveProject = useTrackedFunc(leaveProject, 'Leave Project clicked');
   const onClickDeleteProject = useTrackedFunc(animateThenDeleteProject, 'Delete Project clicked');
-
   return (
     <dialog className="pop-over project-options-pop">
       {showPinOrFeatureSection && (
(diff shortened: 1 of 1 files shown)


GregWeil:
It also gets rid of the ability for anon users to leave projects outside of the editor, but I think that's ok

sarahzinger:
@GregWeil  hmmm... that's a good call out tho... now I'm thinking I don't like this approach lol. gunna close this pr and reopen once it keeps existing functionality.

ttseng:
totally agree that we shouldn't be exposing features like featuring + pinning projects for anonymous users – the only option that should be available to anonymous users is 

I'm wary of removing the option to delete projects from the profile page for a few reasons:
1. we feature such a prominent link to their profile from the homepage, and we also always show a section for deleted projects on their profile page without a clear link of how to do delete projects from that location
2. deleting from a project page is less discoverable than the project options pop

note: you also can't delete a project from the editor if you're anonymous.

I'd suggest adding this logic to `ProjectOptionsContent` in `project-options-pop` which only enables the delete project option for anonymous users who have permissions:
```
  // anonymous users can only delete projects
  if (!(props.currentUser && props.currentUser.login && showDeleteProject)) {
    return (
      <dialog className="pop-over project-options-pop">
        <section className="pop-over-actions danger-zone last-section">
          <PopoverButton onClick={onClickDeleteProject} text="Delete Project " emoji="bomb" />
        </section>
      </dialog>
    );
  }
```

a remix with that change inline: https://glitch.com/edit/#!/defiant-punishment?path=src/presenters/pop-overs/project-options-pop.js:96:0

PR #496, opened 2019-05-02
abrupt-grill: Fix collection deletion

taravancil:
Before: you couldn't delete a collection. Now you can! 🗑

* [abrupt-grill.glitch.me](https://abrupt-grill.glitch.me)
* [Case: Can't delete collections](https://glitch.fogbugz.com/f/cases/3328477/Can-t-delete-collections)

## Changes:
The root cause of this bug was that we expected the `CollectionItem` in the collections list on profiles/team pages to be an `<li>`, but it was a `<div>`.

*Most* `CollectionItem`s are contained in `<ul>`s, but not all (see search results), so I decided that whenever we want the `CollectionItem` to be inside an `<li>`, we should explicitly wrap it as such.

### Other changes
I also fixed a layout bug where the "Included in collections" section was only ever displaying one grid column.

## How To Test:
* Delete a collection from your profile page
* Review collections lists on [user pages](https://abrupt-grill.glitch.me/@taravancil) and [team pages](https://abrupt-grill.glitch.me/@glitch)
* Review the "More collections" list on [a collection page](https://abrupt-grill.glitch.me/@taravancil/glitch)
* Review the "Included in collections" list on [a project page](https://abrupt-grill.glitch.me/~sponcron)
* Review [collection search results](https://abrupt-grill.glitch.me/search?q=butt&activeFilter=collection)

## Feedback I'm looking for:
Did I introduce any design regressions? I checked thoroughly, but you know how that goes...


src/presenters/collections-list.js
```diff
@@ -129,16 +128,17 @@ export const CollectionsUL = ({ collections, deleteCollection, isAuthorized }) =
...
       {orderedCollections.map((collection) => (
-        <CollectionItem
-          key={collection.id}
-          {...{
-            collection,
-            isAuthorized,
-            deleteCollection,
-          }}
-        />
+        <li>
```
sarahzinger: I think you need the key on this top level li by the way or you'll get an error in the console

sarahzinger:
a small style regression for you is on the collections pages things seem to have moved up in a funky way I think, the user avatars are much closer to the add project button at the top of the collection. But otherwise tested deleting both from the collections page and from the user page and it worked! great catch!

This has to do with the popovers yeah? What is it with those today I keep running into bugs with them lol. They are tricky too I feel like every time I touch them I risk making them worse lol.

How you have it works but another alternative (though maybe a bigger refactor that could happen in another pr) would be to pass the ref of the component we want to delete rather than trying to find it via guessing what is the thing we're trying to delete from inside the child.

So instead of : 
https://github.com/FogCreek/Glitch-Community/blob/d8e6c8d1e2d5ba3c65a87afd664a7b87b68e8843/src/presenters/pop-overs/collection-options-pop.js#L10

https://github.com/FogCreek/Glitch-Community/blob/ff0b1a1effee119463adf30afacbc39c7bd86478/src/presenters/pop-overs/featured-project-options-pop.js#L10

(we set the ref here and pass it along)
https://github.com/FogCreek/Glitch-Community/blob/7f52836fdc9811653e6c2b520f473fa0edd151f2/src/components/project/featured-project.js#L52

src/presenters/collections-list.js
taravancil: Oh right, good catch

taravancil:
> on the collections pages things seem to have moved up in a funky way I think

Wowwww I really filed a case on a bug I created. Totally thought I was looking at glitch.com 🤦‍♀ 

https://glitch.fogbugz.com/f/cases/3328476#BugEvent.21313032

taravancil:
That styling regression on collections is fixed.

> How you have it works but another alternative (though maybe a bigger refactor that could happen in another pr) would be to pass the ref of the component we want to delete rather than trying to find it via guessing what is the thing we're trying to delete from inside the child.

Let's do it in another PR 🙂

taravancil:
I was cleaning up these merge conflicts and realized this has been fixed, maybe inadvertently, by some other changes. Closing 🙃

PR #498, opened 2019-05-02
Coconut restaurant: anon users shouldn't see an add to collections button 

sarahzinger:
## Links
* remix: https://coconut-restaurant.glitch.me
* case: https://glitch.manuscript.com/f/cases/3328445/
* this bug fix got me thinking about popovers more and I documented some of my thoughts in a comment in the pr but also on manuscript: https://glitch.manuscript.com/f/cases/3328381/Build-Popover-component (feel free to tell me if it's not clear)

## GIF/Screenshots:
Before: 
Screen Shot 2019-05-02 at 2 25 17 PM
After:
Screen Shot 2019-05-02 at 2 24 42 PM

## Changes:
* anonymous users will only be able to delete they are admin to or leave projects they are apart of otherwise they never see the dropdowns. anon users will no longer see a broken add to collections button, nor will they be able to feature/unfeature/pin/unpin their projects (until they sign up for an account)

## How To Test:
* logout and create a new project 
* grab your token from local storage (the editor is always on glitch.com so there's no access to this token in the remix)
* go to the remix and add use the token to login
* check out your profile page and click on the dropdown on your newly minted project notice you can only delete
* if you ever join a project I believe you will also see the leaveProject option there (I haven't confirmed this but I trust the system) 
* check out the user/team/collection/search pages and notice you do not see dropdowns to add a project to a collection anymore
* verify stuff still looks good when you are logged in.

## Feedback I'm looking for:
Does this code feel understandable and am I missing any use cases? I don't really think this is a long term solution but a giant refactor felt like something that could happen later. I tried to document in the popover component case, my thoughts on how we could consider approaching this down the road.


sarahzinger:
Hey @GregWeil @ttseng closed the old pr and opened this one if you want to take a look, thanks for your wisdom on the last one!

ttseng:
since we're already passing `currentUser` as a prop to `ProjectOptionsPop`, does it make sense to just define `shouldShowAnonView` in `ProjectOptionsPop`?  That way we also don't check `showLeaveProject` and `showDeleteProject` in both `ProjectOptions` and `ProjectOptionsPop`

sarahzinger:
@ttseng the issue is that if we're not going to show that view and the user is anonymous then we don't want to show the dropdown button either

ttseng:
ah I see. tested this and it works, though something feels off to me about repeating code.  

the only thing I could think of is essentially modifying the `projectOptions` passed to `ProjectOptionsPop` based on whether or not the user is anon:

```
  if (isAnon && !shouldShowAnonView) {
    return null;
  }
  
  const anonProjectOptions = {};
  if (isAnon && shouldShowAnonView) {
    if (showLeaveProject) {
      anonProjectOptions.leaveProject = projectOptions.leaveProject;
    }
    if (showDeleteProject) {
      anonProjectOptions.deleteProject = projectOptions.deleteProject;
    }
  }

  if (anonProjectOptions) {
    projectOptions = anonProjectOptions;
  }
return (
    <PopoverWithButton
      buttonClass="project-options button-borderless button-small"
      buttonText={<div className="down-arrow" aria-label="options" />}
      containerClass="project-options-pop-btn"
    >
...
)
```
but that doesn't seem ideal either.  maybe @GregWeil will have a suggestion.

GregWeil:
As long as this isn't a proper refactor, I'd prefer if it filtered out functions rather than adding a second render path. (unless the split can happen higher up in the tree)

Could it be: (at the top of ProjectOptions, replacing the `Object.keys` check)
```
if (isAnon) {
  projectOptions.addProjectToCollection = null;
}
if (Object.values(projectOptions).every(value => !value)) {
  return null;
}
```

sarahzinger:
ok cool y'all got me thinking about more ways to refactor this... going to try exploring this and reopen when I have it ready sorry for the back and forth and thank you again for you thoughts!

PR #501, opened 2019-05-03
Boiling sunflower: Hide suspended project details

oliviacpu:
## Links
* Remix link: https://boiling-sunflower.glitch.me/
* Manuscript link: https://glitch.fogbugz.com/f/cases/edit/3328294/Suspended-projects-shouldn-t-display-a-standard-project-page
* Specs/Notion docs: n/a

## GIF/Screenshots:
Screenshot 2019-05-03 at 11 33 12 AM

## Changes:
* Project pages for suspended projects will show "suspended project" for the project domain and description, and a "no entry" emoji image for the project avatar to non-owners.
* Project owners will still see the information for a suspended project as normal, in case they need to change it.
* `<ProfileContainer type="team|user|project" />` has been replaced with specific profile containers for each type, and have been moved into `@Components/containers/profile` with an index to allow importing them at that level. (Like `import { ProjectProfileContainer } from @Components/containers/profile;`)
* `TrackedButton` was added to buttons and allows the same props as `Button` to be passed through, to wrap the tracking function.

## How To Test:
* Check out https://boiling-sunflower.glitch.me/~a-bad-project and confirm that you only see "suspended project" and the "no entry" emoji sign
* Make sure avatar/cover actions still work as expected for this

## Feedback I'm looking for:
- I think we could just always pass `avatarActions` and `coverActions` to these now with `isAuthorized` and conditionally show them based on that, rather than doing a map/filter (which is repeated a lot)
- Is there anything about `TrackedButton` that could be improved? Would people prefer `trackingLabel` over `label` or something else?

- [x] Remove type props from places

💁 We should follow up on an editor PR that shows the suspended project pop-up, since it still displays the name. (Replace it with "This project has been suspended")

💁 This also gets put into OpenGraph data for unfurls. We can do a separate PR to hide that as well if we want to.


cori:
Small concern here: when viewing that project as a `super_user` I see the "suspended" content, not the actual content. I think `super_user`s should see the project as an owner would. Will that be onerous?

Repro: 
1. navigate to https://boiling-sunflower.glitch.me/~a-bad-project
1. in the browser console send `window.localStorage.cachedUser = "{\"persistentToken\":\"$IAMGROOT\"}"`
    1. `super_user` banner now shows, so I _believe_ that I'm browsing that Remix as a `super_user`
1. view "suspended project" replacement content, not actual project content.

src/components/buttons/tracked-button.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import Button from 'Components/buttons/button';
+import { useTrackedFunc } from '../../presenters/segment-analytics';
+
+const TrackedButton = ({ label, onClick, ...props }) => {
+  const trackedOnClick = useTrackedFunc(onClick, label);
+  return (
+    <Button props={props} onClick={trackedOnClick}>
```
modernserf: should probably be `<Button {...props}`, right?

oliviacpu:
@cori I'll update the check so that super users also see the original content! Thank you!

src/components/buttons/tracked-button.js
modernserf: Also, does TrackedButton render `children`, or the tracking label?

src/components/containers/cover-container.js
```diff
@@ -1,35 +1,44 @@
...
 
-const CoverContainer = ({ buttons, children, type, item }) => {
+const CoverContainer = ({ coverActions, children, type, item }) => {
   const cache = item._cacheCover; // eslint-disable-line no-underscore-dangle
   return (
     <div className={styles.coverContainer} style={getProfileStyles[type]({ ...item, cache })}>
       {children}
       <div className={styles.buttonWrap}>
-        {buttons}
+        {coverActions &&
```
modernserf: You repeat this idiom in a bunch of places -- maybe this could be `<TrackedButtonGroup/>` or something?

modernserf:
This really should be two different PRs, right? the containers have very little to do with suspended projects.

src/presenters/pages/project.js
```diff
@@ -185,7 +185,7 @@ const ProjectPage = ({
               />
             ) : (
               <>
-                {domain} {project.private && <PrivateBadge />}
+                {suspendedReason ? ('suspended project') : (domain)} {project.private && <PrivateBadge />}
```
sarahzinger: is it possible to be suspended without a reason attached, or is that our only data point to indicate if a project is suspended?

oliviacpu: That is our only data point, I believe.

oliviacpu:
> This really should be two different PRs, right? the containers have very little to do with suspended projects.

It definitely could be. Separating the containers was mostly to make the logic around suspended projects easier to reason about (not having to read through the team/user stuff as well) and then thought "I can go ahead and replace the others so that this isn't creating churn." I can make separate PRs for these if you'd like.

src/presenters/pages/project.js
clottman: That sounds not correct but I'm not sure what the actual answer is. I know that the reason is settable by the suspender at suspend-time, so it won't necessarily always be the same. 

@cori - do you know how to tell programmatically if the project is suspended?

clottman: Ah - I misread the code and thought you were expecting reason to always be == 'suspended project' which wouldn't be true. Suspendedproject being set at all though might be.

clottman:
Does this PR handle not showing the full readme, also? Some of our suspended projects might have a readme with content that was part of the reason they got suspended. 
@cori @jennschiffer

oliviacpu:
@clottman The README will error out and not load, I believe. `Access to XMLHttpRequest at 'https://api.glitch.com/projects/a-bad-project/readme' from origin 'https://boiling-sunflower.glitch.me' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource.`

src/presenters/pages/project.js
cori: Sorry for the delay here. `suspended_at` will always be set if a project is suspended. `suspended_reason` is not programmatically required so it may be `undefined` but is "required" by human workflow (i.e. the expectation is that one is always set so the community knows why we suspended it).

oliviacpu:
> I tested this as a super_user as well and was able to see the image/description/name, but not the readme nor the project embed but I'm guessing that's chill/part of devtools?

Yeah, once it's deployed the first thing I will do is make tasks for that and some other things I found while working on this.

PR #507, opened 2019-05-13
Lemon molecule: Add User / Team Toggle to Add project to collection pop

ttseng:
## Links
* Remix link: https://lemon-molecule.glitch.me/~lemon-molecule
* Manuscript link: https://glitch.manuscript.com/f/cases/edit/3328498/Use-segmented-buttons-to-segment-collections-team-collections-in-add-project-to-collection-pop

## GIF/Screenshots:
segmented-button-toggle

## Changes:
* by default, load user collections and load team collections on demand using segmented buttons in `add-project-to-collection-pop` 

## How To Test:
Log in and try testing these cases:

* if user has no collections and isn't part of any teams, they should see the default message without any segmented buttons:
image

* if user isn't part of any teams, they should not see the segmented button toggles:
image

* if user is part of team(s) but none of the teams have any collections, show segmented buttons + default hint text:
image

* if user is part of teams that have collections, those collections should show up under the team collections
image

## Feedback I'm looking for:
* I did some refactoring of `add-project-to-collection-pop` so that the loader appears within the results section when toggling between the segmented buttons (as opposed to reloading the entire pop-over contents, including the segmented buttons + filter input) but this means that some of the hooks are only conditionally rendered, which gives these warnings:
```
 77:29  error  React Hook "React.useState" is called conditionally. React Hooks must be called in the exact same order in every component render     react-hooks/rules-of-hooks

  78:26  error  React Hook "useDebouncedValue" is called conditionally. React Hooks must be called in the exact same order in every component render  react-hooks/rules-of-hooks

  79:31  error  React Hook "React.useMemo" is called conditionally. React Hooks must be called in the exact same order in every component render      react-hooks/rules-of-hooks

```
https://glitch.com/edit/#!/lemon-molecule?path=src/presenters/pop-overs/add-project-to-collection-pop.js:77:1

Any suggestions for a good way to move the hooks out of the conditional?  (note that it's an issue when `collections` is null, which is the case when we're reloading the results list to show teams / user collections)

### DIFF
--- a/src/presenters/pop-overs/add-project-to-collection-pop.js
+++ b/src/presenters/pop-overs/add-project-to-collection-pop.js
@@ -3,7 +3,6 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { flatten, orderBy } from 'lodash';
 import Loader from 'Components/loader';
-
 import SegmentedButtons from 'Components/buttons/segmented-buttons';
 import TextInput from 'Components/inputs/text-input';
 import { getAllPages } from 'Shared/api';
@@ -19,7 +18,6 @@ import CreateCollectionPop from './create-collection-pop';
 import CollectionResultItem from '../includes/collection-result-item';
 
 import { NestedPopover, NestedPopoverTitle } from './popover-nested';
-
 
 const filterTypes = ['Your collections', 'Team collections'];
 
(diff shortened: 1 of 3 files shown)


src/presenters/pop-overs/add-project-to-collection-pop.js
```diff
@@ -52,18 +56,38 @@ const AddProjectToCollectionPopContents = ({
   fromProject,
   project,
   togglePopover,
+  collectionType,
+  setCollectionType,
 }) => {
+  if (!collections) {
```
GregWeil: I think the simplest way to have this switch is where the component is rendered, where it can pick either this component or an alternate loader form component that just renders this bit without the hooks. It'd also let this component expect that it's always rendered with a collections array that's loaded.

src/presenters/pop-overs/add-project-to-collection-pop.js
```diff
@@ -52,18 +56,38 @@ const AddProjectToCollectionPopContents = ({
...
+    return (
+      <dialog className="pop-over add-project-to-collection-pop wide-pop">
+        {/* Only show this nested popover title from project-options */}
+        {!fromProject && <AddProjectPopoverTitle project={project} />}
+
+        {currentUser.teams.length > 0 && <UserOrTeamSegmentedButtons activeType={collectionType} setType={setCollectionType} />}
```
GregWeil: This line and above are duplicated between the loaded and unloaded contents. Could you shift it around so they get shared between the two.

src/presenters/pop-overs/add-project-to-collection-pop.js
ttseng: 👍I created a new component `AddProjectToCollectionResults` that requires a collections array and contains all the query/filtering hooks.  
`AddProjectToCollectionPopContents` then only renders `AddProjectToCollectionResults` if the collections array exists; otherwise, it shows the loader.

src/presenters/pop-overs/add-project-to-collection-pop.js
ttseng: done - this only exists in `AddProjectToCollectionPopContents` now

ttseng:
this is ready for another review now - thanks @GregWeil !

PR #509, opened 2019-05-13
wandering-save: Show which collections a project is already in

taravancil:
## Links
* [wandering-save.glitch.me](https://wandering-save.glitch.me)
* [Manuscript](https://glitch.fogbugz.com/f/cases/3328305/Add-to-collection-pop-indicate-if-the-project-is-already-in-a-given-collection)
* Spec:
image

## GIF/Screenshots:
**Project in no collections**
image

**Project in one collection**
image

**Project in many collections**
image

## Changes:
This adds a section to `add-project-to-collection-pop` that lists which collections a project is already included in.

## How To Test:
Add a project to a collection

### DIFF
--- a/src/presenters/pop-overs/add-project-to-collection-pop.js
+++ b/src/presenters/pop-overs/add-project-to-collection-pop.js
@@ -1,9 +1,11 @@
 // add-project-to-collection-pop -> Add a project to a collection via a project item's menu
 import React from 'react';
 import PropTypes from 'prop-types';
+import Pluralize from 'react-pluralize';
 import { flatten, orderBy } from 'lodash';
 import Loader from 'Components/loader';
 import TextInput from 'Components/inputs/text-input';
+import { CollectionLink } from 'Components/link';
 import { getAllPages } from 'Shared/api';
 import { captureException } from '../../utils/sentry';
 import useDebouncedValue from '../../hooks/use-debounced-value';
@@ -18,7 +20,7 @@ import CollectionResultItem from '../includes/collection-result-item';
 
 import { NestedPopover, NestedPopoverTitle } from './popover-nested';
 
-const NoSearchResultsPlaceholder = () => <p className="info-description">No matching collections found – add to a new one?</p>;
+const NoSearchResultsPlaceholder = () => <p className="info-description">No matching collections found – add to a new one?</p>;
 
 const NoCollectionPlaceholder = () => <p className="info-description">Create collections to organize your favorite projects.</p>;
 
@@ -32,21 +34,21 @@ AddProjectPopoverTitle.propTypes = {
 };
 
 const AddProjectToCollectionResultItem = React.memo(({ onClick, collection, ...props }) => {
-  const onClickTracked = useTrackedFunc(onClick, 'Project Added to Collection', {}, {
-    groupId: collection.team ? collection.team.id : 0,
-  });
-  return (
-    <CollectionResultItem
-      onClick={onClickTracked}
-      collection={collection}
-      {...props}
-    />
+  const onClickTracked = useTrackedFunc(
+    onClick,
+    'Project Added to Collection',
+    {},
+    {
+      groupId: collection.team ? collection.team.id : 0,
+    },
   );
+  return <CollectionResultItem onClick={onClickTracked} collection={collection} {...props} />;
 });
 
 const AddProjectToCollectionPopContents = ({
   addProjectToCollection,
   collections,
+  collectionsWithProject,
   createCollectionPopover,
   currentUser,
   fromProject,
@@ -55,10 +57,11 @@ const AddProjectToCollectionPopContents = ({
 }) => {
   const [query, setQuery] = React.useState('');
   const debouncedQuery = useDebouncedValue(query.toLowerCase().trim(), 300);
-  const filteredCollections = React.useMemo(
-    () => collections.filter((collection) => collection.name.toLowerCase().includes(debouncedQuery)),
-    [debouncedQuery, collections],
-  );
+  const filteredCollections = React.useMemo(() => collections.filter((collection) => collection.name.toLowerCase().includes(debouncedQuery)), [
+    debouncedQuery,
+    collections,
+  ]);
+
   return (
     <dialog className="pop-over add-project-to-collection-pop wide-pop">
       {/* Only show this nested popover title from project-options */}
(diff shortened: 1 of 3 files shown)


src/presenters/pop-overs/add-project-to-collection-pop.js
```diff
@@ -122,48 +134,51 @@ const AddProjectToCollectionPop = (props) => {
...
+        const [projectCollections, ...collectionArrays] = await Promise.all(requests);
+
+        const alreadyInCollectionIds = new Set(projectCollections.map((c) => c.id));
+        const collections = flatten(collectionArrays).filter((c) => !alreadyInCollectionIds.has(c.id));
+
+        const orderedCollections = orderBy(collections, (collection) => collection.updatedAt, 'desc');
+
+        if (!canceled) {
+          setMaybeCollections(orderedCollections);
+          setCollectionsWithProject(projectCollections);
```
ttseng: I think we should only show collections here that the user is authorized to edit, rather than all the collections the project is a part of.  Then the notice is more of a hint as to why some of their collections aren't automatically shown in the results list (and potentially prevents the list from becoming too long).

We can use lodash [partition](https://lodash.com/docs/4.17.11#partition) to get the array of collectionsWithProject:

```
    const alreadyInCollectionIds = new Set(projectCollections.map((c) => c.id));
        const [collections, collectionsWithProject] = partition(flatten(collectionArrays), (c) => !alreadyInCollectionIds.has(c.id));
        console.log('collections', collections, 'collectionsWithProject', collectionsWithProject);
        
        const orderedCollections = orderBy(collections, (collection) => collection.updatedAt, 'desc');

        if (!canceled) {
          setMaybeCollections(orderedCollections);
          setCollectionsWithProject(collectionsWithProject);
        }
```
(testable at https://suave-rosemary.glitch.me/~magic-eye)

modernserf:
What if this is in, like, hundreds of collections? is there a point where we should say "foo is already in collections A, B, C and 9 others"?

ttseng:
^ yeah, limiting the list to 3 collection names makes sense. we can use badge styling for the count of excluded collection names

image

src/presenters/pop-overs/add-project-to-collection-pop.js
taravancil: Good catch! Thanks 👍

clottman:
I was just thinking yesterday that this would be awesome to have! Yay!!!

src/presenters/pop-overs/add-project-to-collection-pop.js
```diff
@@ -90,6 +93,19 @@ const AddProjectToCollectionPopContents = ({
         <section className="pop-over-info">{query ? <NoSearchResultsPlaceholder /> : <NoCollectionPlaceholder />}</section>
       )}
 
+      {collectionsWithProject.length ? (
```
clottman: Probably want to check length > 0 here instead of just length, right?

clottman: Also instead of the ternary (`condition ? result : null`) we can simplify to `condition && result` for the same effect.

src/presenters/pop-overs/add-project-to-collection-pop.js
```diff
@@ -122,48 +138,51 @@ const AddProjectToCollectionPop = (props) => {
...
-
-    loadCollections().catch(captureException);
-    return () => {
-      canceled = true;
-    };
-  }, [project.id, currentUser.id]);
+  const [collectionsWithProject, setCollectionsWithProject] = React.useState([]);
+
+  React.useEffect(
+    () => {
```
clottman: What's going on with your prettier settings?

src/presenters/pop-overs/add-project-to-collection-pop.js
```diff
@@ -122,48 +138,51 @@ const AddProjectToCollectionPop = (props) => {
...
+      const loadCollections = async () => {
+        const requests = [
+          getAllPages(api, `v1/projects/by/id/collections?id=${project.id}&${orderParams}`),
+          loadUserCollections(currentUser),
+          ...currentUser.teams.map((team) => loadTeamCollections(team)),
+        ];
+        const [projectCollections, ...collectionArrays] = await Promise.all(requests);
+
+        const alreadyInCollectionIds = new Set(projectCollections.map((c) => c.id));
+        const [collections, _collectionsWithProject] = partition(flatten(collectionArrays), (c) => !alreadyInCollectionIds.has(c.id));
```
clottman: I'm having a hard time grokking the difference between `alreadyInCollectionsIds` and `_collectionsWithProject`.

src/presenters/pop-overs/add-project-to-collection-pop.js
modernserf: `collectionsWithProject.length ? result : null` is equivalent to `collectionsWithProject.length > 0 && result`; idk if we have any preference indicated in our style guide

taravancil: I don't see anything written about it in STYLE.md 🤷‍♀

src/presenters/pop-overs/add-project-to-collection-pop.js
taravancil: Not sure, are these not right?

```
{
  "arrowParens": "always",
  "printWidth": 150,
  "singleQuote": true,
  "trailingComma": "all" 
}

File: src/presenters/pop-overs/add-project-to-collection-pop.js (hunk 4 above)
Comment (taravancil) [reply src/presenters/pop-overs/add-project-to-collection-pop.js]: `alreadyInCollectionsIds` is just the IDs, `_collectionsWithProject` is an array of the collections. You think we should consider renaming one or both to make it more clear?

File: src/presenters/pop-overs/add-project-to-collection-pop.js [hunk 5]
```diff
@@ -126,6 +131,29 @@ const AddProjectToCollectionPopContents = ({
         <AddProjectToCollectionResults {...{ addProjectToCollection, collections, currentUser, project, togglePopover, query }} />
       )}
 
+      {collections && collectionsWithProject.length ? (
```
taravancil: Add a check for `collections` so this section isn't visible as you toggle between your collections and team collections

taravancil:
I pulled in the segmented buttons changes: 

image
image

PR #511, opened 2019-05-13
Jagged flea: thread api through DataLoader

modernserf:
## Links
* Remix link: https://jagged-flea.glitch.me/

## Changes:
* make `api` the argument to `get` prop in DataLoader
* remove now unnecessary `useAPI` hooks
* replace some uses of useAPI / createAPIHook with DataLoader
* add optional error logging to DataLoader / createAPIHook

## Feedback I'm looking for:
DataLoader currently only fetches on mount; that's potentially a hazard for bugs. Should we have an optional argument for dependencies that we pass into useState? It could look like

```js
const getUser = async (api, userId) => { ... }

<DataLoader get={api => getUser(api, user.id)} getArgs={[user.id]} ... >
```
or, if we're willing to change how we write the getter functions, we could do
```js
const getUser = async (api, { userId }) => { ... }

<DataLoader get={getUser} getArgs={{ userId }} ... />
```


oliviacpu:
> or, if we're willing to change how we write the getter functions, we could do

```
const getUser = async (api, { userId }) => { ... }

<DataLoader get={getUser} getArgs={{ userId }} ... />
```

I am willing to do this.

PR #512, opened 2019-05-14
Faceted vase: use AnimationContainer for project & collection popup actions

modernserf:
## Links
* Remix link https://faceted-vase.glitch.me
* Manuscript link https://glitch.manuscript.com/f/cases/3328456/
see also discussion in #496

## Changes:
* Manage animations on project & collection items on those items, rather than in the popups that trigger them
* Redesign AnimationContainer to manage its own state & ensure that it's only reacting to its own animation events

## How To Test:
* log into https://faceted-vase.glitch.me
* pinning a project should animate with a slide-up
* unpinning a project should animate with a slide-down
* deleting a collection should animate with a slide down
* undeleting a project should animate with a slide-up

### DIFF
--- a/src/components/animation-container/index.js
+++ b/src/components/animation-container/index.js
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React, { useState, useRef } from 'react';
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
 import styles from './animations.styl';
@@ -7,9 +7,18 @@ const types = ['slideDown', 'slideUp'];
 
 const AnimationContainer = ({ type, children, className, onAnimationEnd }) => {
   const [state, setState] = useState({ active: false, handlerArgs: [] });
+  const ref = useRef();
 
   return (
-    <div className={classnames(className, state.active && styles[type])} onAnimationEnd={() => onAnimationEnd(...state.handlerArgs)}>
+    <div
+      ref={ref}
+      className={classnames(className, state.active && styles[type])}
+      onAnimationEnd={(event) => {
+        if (event.target === ref.current) {
+          onAnimationEnd(...state.handlerArgs);
+        }
+      }}
+    >
       {children((...handlerArgs) => setState({ active: true, handlerArgs }))}
     </div>
   );
--- a/src/components/animation-container/index.js
+++ b/src/components/animation-container/index.js
@@ -5,6 +5,13 @@ import styles from './animations.styl';
 
 const types = ['slideDown', 'slideUp'];
 
+/* Usage:
+<AnimationContainer type="slideDown" onAnimationEnd={(id) => deleteProject(id)}>
+  {(animateOutAndDeleteProject) => (
+    <Button onClick={() => animateOutAndDeleteProject(id)}>Delete Project</Button>
+  )}
+</AnimationContainer>
+*/
 const AnimationContainer = ({ type, children, className, onAnimationEnd }) => {
   const [state, setState] = useState({ active: false, handlerArgs: [] });
   const ref = useRef();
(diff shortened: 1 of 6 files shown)


src/components/animation-container/index.js
```diff
@@ -1,20 +1,32 @@
...
+    <div
+      ref={ref}
+      className={classnames(className, state.active && styles[type])}
+      onAnimationEnd={(event) => {
+        if (event.target === ref.current) {
+          onAnimationEnd(...state.handlerArgs);
+        }
+      }}
+    >
+      {children((...handlerArgs) => setState({ active: true, handlerArgs }))}
```
sarahzinger: This whole thing looks really cool, but I'm not that familiar with this syntax, you mind walking me through this so I can follow it better? I guess specifically what's an example of handlerArgs?

src/components/project/project-item.js
```diff
@@ -21,36 +23,61 @@ const getLinkBodyStyles = (project) =>
...
-        </div>
-      </div>
-      <div className={styles.description}>
-        <Markdown length={80}>{project.description || ' '}</Markdown>
-      </div>
-    </ProjectLink>
-  </div>
-);
+const ProjectItem = ({ project, projectOptions }) => {
+  const dispatch = (type, ...args) => projectOptions[type](...args);
```
sarahzinger: I guess relatedly how does dispatch get `type` here? Apologies if these questions are annoying I can also just take time and dig into it myself if it's too weird to explain over github!

src/components/collection/collection-item.js
```diff
@@ -58,48 +59,53 @@ ProjectsPreview.propTypes = {
...
+        <CollectionLink collection={collection} className={styles.footerLink}>
           <>
-            {`View ${collection.projects.length >= 3 ? 'all' : ''} `}
-            <Pluralize count={collection.projects.length} singular="project" />
-            <span aria-hidden="true"> →</span>
+            {collection.projects && collection.projects.length > 0 && (
+              <>
+                {`View ${collection.projects.length >= 3 ? 'all' : ''} `}
+                <Pluralize count={collection.projects.length} singular="project" />
+                <span aria-hidden="true"> →</span>
```
sarahzinger: totally unrelated to this pr but gets me thinking that we might want to make ` <span aria-hidden="true"> →</span>` into it's own little component or something because I've definitely forgotten that aria-hidden before

src/components/deleted-projects/index.js
```diff
@@ -45,9 +43,7 @@ export const DeletedProjectsList = ({ deletedProjects, undelete }) => {
 
   return (
     <Grid items={deletedProjects} className={styles.deletedProjectsContainer}>
-      {(({ id, domain }) => (
-        <DeletedProject id={id} domain={domain} onClick={() => undeleteTracked(id)} />
-      ))}
+      {({ id, domain }) => <DeletedProject id={id} domain={domain} onClick={() => undeleteTracked(id)} />}
```
sarahzinger: I see now we want to pass the id at a higher level 💯 very cool thanks for sharing Justin, definitely learned a thing.

I suspect this pr would be a good like blog post/talk or something I'm sure I'm not the only one wondering ways to handle situations like this.

src/components/project/project-item.js
sarahzinger: since type here refers to the function in projectOptions, maybe we can choose a name that describes that like `projectOptionKey` or `funcKey` or something ? lol as I list other options `type` seems pretty reasonable lol.

src/components/collection/collection-item.js
taravancil: I've been thinking the same about all `aria-` attributes. I figure we should aim for components to make it hard to accidentally forget

src/components/project/project-item.js
modernserf: going with `projectOptionName`

PR #516, opened 2019-05-15
Subsequent atmosphere: encode `+` in API calls

modernserf:
## Links
* Remix link: https://subsequent-atmosphere.glitch.me
* Manuscript link: https://glitch.manuscript.com/f/cases/3328525/Can-t-load-profile-page-for-users-with-in-login

## Changes:
* encode API calls using `by/login`, `by/url`, `by/fullUrl`
* refactor server-side calls to encode in api.js instead of routes.js
* (projects use punycode and cannot actually have non-url characters in their names, and are therefore unaffected by this)

## How To Test:
* user with + in name should load: https://subsequent-atmosphere.glitch.me/@justin+test3
* collection belonging to user with + in name should load: https://subsequent-atmosphere.glitch.me/@justin+test3/test-collection

## Future work:
- replace bare API calls with helper functions/resource manager that does all of this for us

### DIFF
--- a/server/api.js
+++ b/server/api.js
@@ -15,7 +15,7 @@ const api = axios.create({
 });
 
 function getProjectFromApi(domain) {
-  return getSingleItem(api, `v1/projects/by/domain?domain=${punycode.toASCII(domain)}`, domain);
+  return getSingleItem(api, `v1/projects/by/domain?domain=${domain}`, domain);
 }
 
 function getTeamFromApi(url) {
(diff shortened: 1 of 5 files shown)


server/api.js
```diff
@@ -13,20 +14,22 @@ const api = axios.create({
   timeout: 5000,
 });
 
-async function getProjectFromApi(domain) {
-  return await getSingleItem(api, `v1/projects/by/domain?domain=${domain}`, domain);
+function getProjectFromApi(domain) {
```
clottman: What's the goal of removing the `await`s here?

modernserf: `return await` doesn't do anything useful, it just adds an extra layer of promise wrapping/unwrapping

PR #517, opened 2019-05-15
ginger-trout: even more styling fixes

ttseng:
## Links
* Remix link: https://ginger-trout.glitch.me/
* Manuscript link: 
https://glitch.manuscript.com/f/cases/3328544/Adjust-button-margins-underneath-embed
https://glitch.manuscript.com/f/cases/3328543/Adjust-bottom-margin-for-ProjectsContainer-to-add-spacing-between-container-report-abuse-button
https://glitch.manuscript.com/f/cases/3328541/Segmented-Button-Margins-are-off
https://glitch.manuscript.com/f/cases/3328542/Need-to-handle-overflow-of-markdown-in-project-descriptions

## Changes:
* Fixed overflow issue with markdown in project descriptions:
https://ginger-trout.glitch.me/search?q=saskia
https://glitch.com/search?q=saskia
image

* Fixed segmented button styling (note that there is an override in the add-project-to-collection-pop)
image

* Fixed margins for projects-list__projectsGrid, which was affecting the placement of the Report Abuse button (note there are now overrides when the projects-list is within a container, like more by or recent projects)
https://ginger-trout.glitch.me/@scientiffic-test
image

* Adjusted collection-options-pop placement
image

* Edited small button padding in response to https://glitch.manuscript.com/f/cases/3328544/Adjust-button-margins-underneath-embed

* Fixed grid size for collection pages' projectsGridCompact (currently showing a max of 2 columns on prod instead of 3)
image

## How To Test:
* Confirm the changes listed above and browse to see if the changes might have affected anything else unintentionally?


src/components/containers/projects-list.styl
```diff
@@ -1,16 +1,13 @@
...
-
 .projectsGrid
   --row-gap: 20px
   --min-width: 300px
   align-items: end
   
 .projectsGridCompact
-  --row-gap: 20px
-  --min-width: 250px
+  --row-gap: 20px !important
```
modernserf: What do these do? I'm not seeing any difference between these `!important`s being present or not

src/components/collections-list/index.js
```diff
@@ -97,9 +97,7 @@ function CollectionsList({ collections: rawCollections, title, isAuthorized, may
       <Heading tagName="h2">{title}</Heading>
       {canMakeCollections && (
         <>
-          <div className={styles.createCollectionButtonWrap}>
-            <CreateCollectionButton maybeTeam={maybeTeam} />
-          </div>
+          <CreateCollectionButton maybeTeam={maybeTeam} />
```
sarahzinger: I feel like the margin here though makes sense since it's internal to the component you know? it's just the external ones that I think are potentially wonky...or at least that's my interpretation

src/components/containers/projects-list.styl
modernserf: oh, i see -- this is a specificity thing. hmm, i'd like to fix this in the grid container rather than using !important here

ttseng: yeah, I was trying to figure that out because on staging, it seems like css from `projects-list.styl` overrides anything from `grid-styl`:
image

But on production the opposite is true...
image

styles/pop-overs.styl
```diff
@@ -301,6 +298,8 @@ popover-secondary-background()
     img
       width: 15px
       border-radius: 3px
+  .segmented-button-wrap
+    margin-bottom: -1rem
```
sarahzinger: I think rather than wrapping when we need to negate a margin, I think it might be easier to maintain to not have a margin in the core component, and only add it in a wrapper when we need it. That seems to be the convention we're suggesting in our docs anyway https://github.com/FogCreek/Glitch-Community/blob/master/STYLE.md#components this way the components don't know where they are being used on the page and the parent components don't need to know the internals of their components to reset their margins.

src/components/containers/projects-list.styl
modernserf: oh, how about in `containers/grid.styl`

```
.grid
  display: grid  
  grid-row-gap: var(--row-gap, 10px)
  grid-column-gap: var(--column-gap, 10px)
  grid-template-columns: repeat(auto-fill, minmax(var(--min-width, 300px), 1fr))
  list-style-type: none
  margin: 0
  padding: 0
```

ttseng: yeah! that works perfectly

src/components/collections-list/index.js
ttseng: The reason I got rid of the `createCollectionButtonWrap` is because it was adding an unneeded bottom margin to the `Create Collection button (compare changes on ginger-trout on left with what's on production on right)

image

(and the distance between the button and the collection item already feels further than normal because of the lack of user avatars)

sarahzinger: oh sorry I wasn't being clear I was agreeing with your change here of not wrapping if that makes sense!

ttseng:
one other small adjustment with user-options-pop dropdown - added some left margin between the dropdown and the avatar 
image

ttseng:
also fixed collection avatar SVG sizing in firefox by adding `min-width:0` to autocomplete avatarContainer (https://github.com/philipwalton/flexbugs/issues/184)
image

and removed the "anonymous projects expire after 2 weeks" message on the homepage for anonymous users with projects (since this is no longer the case)
image

not planning on making any additional changes, so this is ready for any additional comments / approval @modernserf @sarahzinger!

PR #519, opened 2019-05-15
Jet socks: modularize new project pop

modernserf:
## Links
* Remix link: https://jet-socks.glitch.me/

## GIF/Screenshots
self-adjusting-popover

## Changes
* create Popover components
* create ResultsList component (possibly only used in popovers, but not really popover-specific)
* incorporate new-project-pop into Header component
* use the v1 API for fetching new projects

## How To Test
* visit https://jet-socks.glitch.me/
* clicking New Project should open a menu aligned to the right of the button
* if you make the window narrow enough that it would go off the edge of the screen, it should adjust to fit
* log in
* the New Project menu should now align to the left of the button

## Feedback I'm looking for
* I think the right approach for this conversion is to work on fully converting each popover one by one, copying styles and components as i go, and then delete all of the legacy styles/presenters when they're all complete; I think this will be easier than trying to convert a single component across multiple popovers. Does this seem sensible?
* ~I left new-project-pop in presenters/pop-overs because I would categorize it as something that's specific to our app, rather than something that someone else would use to build their own app; its more like a very small page than a very complicated component. That said, we have a ton of stuff in components that is also more content-ish than component-ish (e.g. the header & footer). Any thoughts on how we should organize these? I feel strongly that we should have the components that one uses to _build_ popovers separate from the popovers themselves, but I'm otherwise agreeable to suggestions.~ I went through the popovers directory and the majority of them belong to a single, existing component, so that's where they will live; shared popovers will be their own components.


src/components/popover/styles.styl
```diff
@@ -0,0 +1,53 @@
...
+    color: primary
+    background-color: secondary-background
+  &.dangerZone
+    color: primary
+    background-color: warning-background
+    
+.popoverSection + .popoverSection
+  border-top: 1px solid secondary
+    
+.popoverActions
```
ttseng: Since most popover sections have this padding (14px 12px), with the exception of results-lists
I'm wondering if, rather than applying this popoverAction padding styling only to `popoverSections` with the `popoverAction` className, we can get rid of popoverAction altogether and just use this padding as the default `popoverSection` styling (and override it for pop-over results-lists)

ttseng:
Your approach of updating the pop-over component over time as each pop-over gets componentized makes sense to me.

I had been working on adding accessible focus to the pop-overs in this branch here: https://github.com/FogCreek/Glitch-Community/pull/488 (most of the changes for it are here:   const focusFirstElement = (dialog) => {)

We could either try to incorporate those changes in this PR, or I can do a follow-up PR.  What do you think?

modernserf:
I can roll your changes into this PR.

ttseng:
awesome - thanks justin

src/components/popover/dialog.js
```diff
@@ -0,0 +1,85 @@
...
+  const { status: popoverStatus } = useContext(PopoverToggleContext);
+  useEffect(() => {
+    const dialog = ref.current;
+    if (!dialog || popoverStatus !== 'openedFromKeyboard') return;
+    // focus on the dialog if there is only a destructible item in the popover that shouldn't automatically be focused on
+    if (focusOnDialog) {
+      dialog.focus();
+    } else {
+      const focusableElements =
+        'a:not([disabled]), button:not([disabled]), input:not([disabled]), [tabindex]:not([disabled]):not([tabindex="-1"]), select:not([disabled]), textarea:not([disabled])';
```
sarahzinger: pretty sure this is unrelated to this pr, but in qaing this pr I noticed some wonkyness around safari and navigating via tabs, but on chrome it works as expected. Anyway something for us to keep an eye out for I guess, as we continue to iterate on these popovers :)

src/components/header/new-project-pop.styl
```diff
@@ -0,0 +1,31 @@
+@import  "../../components/constants"
+
+.project
+  display: flex
+  align-items: flex-start
+  padding: 12px 6px
+  
+.projectAvatar
+  flex: 0 0 22px
+  :global .avatar
```
sarahzinger: is this ok? should we avoid this stuff/fix it in another pr maybe?

src/components/popover/dialog.js
```diff
@@ -0,0 +1,85 @@
...
+      const focusableDialogElement = dialog.querySelector(focusableElements);
+      if (focusableDialogElement) {
+        focusableDialogElement.focus();
+      }
+    }
+  }, [popoverStatus]);
+  return { ref };
+};
+
+const usePositionAdjustment = ({ margin }) => {
```
sarahzinger: very cool! ✨

src/components/header/new-project-pop.styl
modernserf: we'll clean this up when we componentize ProjectAvatar

src/components/header/index.js
```diff
@@ -33,7 +35,7 @@ const Header = ({ searchQuery, showNewStuffOverlay }) => {
         </div>
         <ul className={styles.buttons}>
           <li className={styles.buttonWrap}>
-            <NewProjectPop opensFromRight={!(currentUser && currentUser.login)} />
+            <NewProjectPop align={currentUser && currentUser.login ? 'left' : 'right'} />
```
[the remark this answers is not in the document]
modernserf: ah, i guess i had misunderstood what was supposed to be happening with this 👍

PR #520, opened 2019-05-16
Stupendous cannon: Super users can view deleted projects

sarahzinger:
## Links
* https://stupendous-cannon.glitch.me/
* https://glitch.manuscript.com/f/cases/3328513/

## GIF/Screenshots:
Screen Shot 2019-05-16 at 11 52 20 AM

## Changes:
* use the new api to ping for deleted projects
* create a view-only version of deleted projects that viewable only by support folks

## How To Test:
* go to https://stupendous-cannon.glitch.me/@test-sarah with super_user enabled, click show deleted projects, notice you can't click on the projects, but you can see the titles. 

## Feedback I'm looking for:
* anything I'm missing?

### DIFF
--- a/src/components/deleted-projects/index.js
+++ b/src/components/deleted-projects/index.js
@@ -75,7 +75,7 @@ function DeletedProjects({ deletedProjects, setDeletedProjects, undelete, user }
   const clickShow = async () => {
     setState('loading');
     try {
-      const { data } = await api.get(`v1/users/${user.id}/deletedProjects`);
+      const { data } = await api.get(`v1/users/${user.id}/deletedProjects?limit=100`);
       setDeletedProjects(data.items);
       setState('ready');
     } catch (e) {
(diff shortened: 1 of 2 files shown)


cori:
functionally LGTM!

src/components/deleted-projects/index.js
```diff
@@ -47,20 +48,35 @@ export const DeletedProjectsList = ({ deletedProjects, undelete }) => {
...
+function DeletedProjects({ deletedProjects, setDeletedProjects, undelete, user }) {
   const api = useAPI();
   // states: hidden | loading | ready
   const [state, setState] = useState('hidden');
   const clickShow = async () => {
     setState('loading');
     try {
-      const { data } = await api.get('user/deleted-projects');
-      setDeletedProjects(data);
+      const { data } = await api.get(`v1/users/${user.id}/deletedProjects?limit=100`);
```
clottman: How does pagination work for this, if the user has more than 100 deleted projects?

sarahzinger:
@clottman great catches!! right now we have no pagination... do you think we need it? I'm thinking if we add the sorting that you suggest, the last 100 should be fine for the ui and if there's a real need we can always revisit and add back pagination, yeah?

sarahzinger:
ah you know what I see we have this helper called getAllPages... let me refactor and try doing that, that way we get will get all the projects, not just the first hundred.

PR #526, opened 2019-05-17
Humdrum triangle: The Cypress PR

oliviacpu:
## Links
* Remix link: https://humdrum-triangle.glitch.me/
* Manuscript link: ?
* Specs/Notion docs: https://www.notion.so/glitch/Integration-Testing-with-Cypress-io-9c936b5a4683435ebebce0dd9d90267c

## Changes:
* Adds Cypress and example integration tests

## How To Test:
* Setup and run the tests

## Feedback I'm looking for:
* Improvements on setup or clarity
* Anything, really

- [x] Put the shared `persistentToken` in Keybase
- [x] Document how to create fixtures using helpers
- [ ] Talk about it, but there's not really any changes to the site to worry about


keithk:
YAYYYYYYY

sarahzinger:
🌟 SO COOL! 🌟 

I tried running the tests on my machine and here's what I learned:
- Had to run npm install lol (I guess I haven't had to do that yet ty glitch, might be good to put in the docs to install npm and to run npm install)
- Cypress opened but then I wasn’t totally sure what to do but I clicked “Run all specs”, should that be in the docs or is that like...obvious lol
- turns out the project fell asleep and took me a minute to remember to check for that, so maybe that could be in docs too
- When run all specs ran, it seemed to only do the suspended projects one, but I might be confused?
- I created a .env file with my own test account but some tests failed because I think some of the tests are still built on your test token, ex: you can only feature/delete projects you are the owner of, etc. Not sure if the answer there is to change the tests or I should setup keybase and get your env token like we were talking about

oliviacpu:
@sarahzinger Updated the TESTING doc with your notes and removed .only from the suspended project scenario

cypress/fixtures/projectsByIds.json
```diff
@@ -0,0 +1,444 @@
...
+        "userId": 3803619,
+        "accessLevel": 30
+      }
+    ],
+    "features": [],
+    "teamIds": []
+  },
+  "349c8d44-934f-4733-b4af-f080b7e3958e": {
+    "id": "349c8d44-934f-4733-b4af-f080b7e3958e",
+    "inviteToken": "6a6e9773-270d-4ba4-ab36-afafdc07c9ed",
```
sarahzinger: should we remove these invites?

cypress/integration/header_actions.js
```diff
@@ -0,0 +1,52 @@
+/// <reference types="Cypress" />
```
sarahzinger: what does this mean? ^

cypress/integration/header_actions.js
```diff
@@ -0,0 +1,52 @@
...
+
+    cy.get('.profile').should('contain', 'Your Projects');
+
+    cy.get('[data-tooltip="User options"]').click();
+    cy.contains('Sign Out').click();
+
+    cy.get('.what-is-glitch').should('exist');
+  });
+
+  describe('Sign In Errors', () => {
```
sarahzinger: any reason to use describe here when we don't use it in the other two? do we want to have opinions on like what's a context vs describe vs it?  also feel free to be like "sarah you are being nitpicky lol"

here's the pattern other I've liked in the past but open to whatever!
describe - top level thing I'm testing (Component/File/etc)
context - situation the thing I'm testing (often I preface the description with "when")
it - expected response aka the test, try to keep one test per `it`.

sarahzinger:
@oliviacpu it all works great for me!! I feel like you may wish to have other people look at this one too, but I vote ship it!! 🎉

cypress/fixtures/projectsByIds.json
oliviacpu: Probably yeah, good catch. I need to borrow some code from our Sentry config. :-/

oliviacpu:
Gonna try to remove secrets from all fixtures like we do for sentry stuff

oliviacpu:
👍

cypress/integration/header_actions.js
sarahzinger: ah nevermind did a google :) sounds good

oliviacpu: I had to Google it too! It is letting me autocomplete in VSCode, it looks like.

cypress/integration/header_actions.js
oliviacpu: Going to try to re-organize some of these as I add a few more cases

sarahzinger: rad, I also think it's not like a blocker to us deploying this btw! I think if we want we can ship as is, and fix up in another pr :) but also up to you!

PR #527, opened 2019-05-17
Noisy scale: sign-in pop

modernserf:
## Links
* Remix link: https://noisy-scale.glitch.me/

## GIF/Screenshots
Screen Shot 2019-05-17 at 3 34 43 PM
Screen Shot 2019-05-17 at 3 34 49 PM
Screen Shot 2019-05-17 at 3 34 55 PM
Screen Shot 2019-05-20 at 3 17 43 PM

## Changes
* modularize CSS for sign-in-pop
* hookify email & sign-in code
* create MultiPopover 

## How To Test:
* visit https://noisy-scale.glitch.me/
* clicking the sign in button in the header should open the sign in menu
* clicking 'sign in with email' should open the email form
* clicking 'use a sign in code' should open the sign-in code form


src/components/sign-in-pop/styles.styl
```diff
@@ -0,0 +1,5 @@
+.termsAndConditions
+  font-size: 13px
```
ttseng: can we also add that the color should be `text-on-secondary-background` (from `components/constants.styl`)?

ttseng:
How do you think we should best handle pop-over widths that should be smaller than the default width of 255px? For example, pop-overs like project-options-pop will feel unnecessarily wide if we use the default. 

image

One option: maybe we could have pop-over buttons set to `white-space: nowrap` and the pop-overs to `width: fit-content` so we end up with variable-width pop-overs sized to fit the buttons  they contain.

src/components/popover/index.js
```diff
@@ -32,11 +34,53 @@ PopoverSection.defaultProps = {
...
 export const PopoverActions = ({ ...props }) => <PopoverSection {...props} className={styles.popoverActions} />;
+export const PopoverInfo = ({ ...props }) => <PopoverSection {...props} className={styles.popoverInfo} />;
 
 const styled = (Component, baseClassName) => ({ className, ...props }) => <Component className={classnames(className, baseClassName)} {...props} />;
 export const PopoverTitle = styled('div', styles.popoverTitle);
 export const InfoDescription = styled('p', styles.infoDescription);
 
+const MultiPopoverContext = createContext();
+
+export const MultiPopover = ({ views, initialView, children }) => {
```
sarahzinger: should we add this to storybook?

src/presenters/pop-overs/create-team-pop.js
```diff
@@ -161,19 +159,9 @@ CreateTeamPopBase.propTypes = {
...
-  ) : (
-    <SignInPopBase
-      hash="create-team"
-      header={<NestedPopoverTitle>Sign In</NestedPopoverTitle>}
-      prompt={<p className="action-description">You'll need to sign in to create a team</p>}
-    />
-  );
-};
-export default CreateTeamPopOrSignIn;
+  return <CreateTeamPop api={api} {...props} />;
```
sarahzinger: should it be `<CreateTeamPopBase api={api} {...props} />` ?

src/components/sign-in-pop/index.js
```diff
@@ -0,0 +1,261 @@
...
+          <SignInCodeSection onClick={setDestinationAnd(showView.signInCode)} />
+        </PopoverDialog>
+      )}
+    </MultiPopover>
+  );
+});
+
+const SignInPopContainer = ({ align }) => (
+  <PopoverWithButton buttonProps={{ size: 'small' }} buttonText="Sign in">
+    {() => <SignInPopBase align={align} />}
```
sarahzinger: could this be:
`<SignInPopBase align={align} />`
instead of:
`{() => <SignInPopBase align={align} />}`
or are there advantages that I'm missing to doing it this way?

modernserf: PopoverWithButton passes `togglePopover` to its children; SignInPopBase just doesn't use it, because it has links that go to other pages.

sarahzinger: I see so it expects a function instead of a react component gotcha 👍

PR #537, opened 2019-05-21
Sideways sound: don't fetch team invitees if there aren't any

modernserf:
## Links
* Remix link: https://sideways-sound.glitch.me
* Manuscript link: https://glitch.manuscript.com/f/cases/3328530/400-error-when-visiting-a-team-page-for-which-current-user-is-only-member

## Changes:
* don't try to fetch users corresponding to a team's invites if there aren't any

## How To Test:
* visit https://sideways-sound.glitch.me/@false-troop
* there should be no network error for /users/by/id 
* visit https://glitch.com/@false-troop
* there will be a network error for /users/by/id

## Future considerations
* maybe invited team users should be at `/team/by/url/invitedMembers`?

### DIFF
--- a/src/presenters/includes/team-users.js
+++ b/src/presenters/includes/team-users.js
@@ -196,7 +196,8 @@ const JoinTeam = ({ onClick }) => (
 
 const useInvitees = createAPIHook(async (api, team, currentUserIsOnTeam) => {
   if (!currentUserIsOnTeam) return [];
-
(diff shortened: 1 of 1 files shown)


modernserf:
I'm tacking on the fix for https://glitch.manuscript.com/f/cases/3328529/ here as well, since its basically a different manifestation of the same bug

PR #540, opened 2019-05-21
Superb guide: Fix project renaming, remove eager success messages

GregWeil:
## Links
- https://superb-guide.glitch.me/
- https://glitch.manuscript.com/f/cases/3328053/Community-site-is-not-using-the-right-endpoint-to-rename-projects
- https://glitch.manuscript.com/f/cases/3328479/On-the-community-site-async-API-calls-that-fail-initially-look-like-success

## Changes:
- Hit an extra endpoint after renaming projects to force a restart and update env vars
- Don't show 'invited to team' or 'removed from team' until the request actually finishes

## How To Test:
- Try inviting people to teams and removing them from teams. The easiest way to test failures is to open two tabs and invite/remove in both, the second will fail since it's already done


src/presenters/pop-overs/add-team-user-pop.js
```diff
@@ -345,7 +344,7 @@ const AddTeamUser = ({ inviteEmail, inviteUser, setWhitelistedDomain, members, i
     try {
       await inviteEmail(email);
     } catch (error) {
-      setInvitee('');
```
sarahzinger: do we have no way to log errors with sentry?

src/presenters/pop-overs/team-user-info-pop.js
```diff
@@ -154,9 +154,9 @@ const TeamUserInfoAndRemovePop = ({ user, team, removeUserFromTeam, updateUserPe
     });
   }, [user.id]);
 
-  function removeUser(selectedProjects = []) {
+  async function removeUser(selectedProjects = []) {
+    await removeUserFromTeam(user.id, Array.from(selectedProjects));
```
sarahzinger: don't we still need to wrap this in a try/catch and stuff? This is like optimistically assuming it just works right? or am I missing stuff

sarahzinger:
also a weird thing:
- invited a test account to test team, the avatar shows up
- tried doing it again (it shows up in search which seems wrong)
- then test user's avatar disapears

src/presenters/pop-overs/team-user-info-pop.js
GregWeil: Sentry will catch anything we don't, so the try/catch is only needed if we want to continue after an error.

sarahzinger: hmm don't we want to handle this use case for users tho? like tell them, "hey we know you tried to remove this user but we couldn't because XYZ"

sarahzinger: we do? where? I'm looking here: https://glitch.com/edit/#!/superb-guide?path=src/presenters/team-editor.js:95:3 I don't see anything doing that? sorry if I'm missing something obvious, my brain is working kind of slowly today lol

src/presenters/pop-overs/add-team-user-pop.js
sarahzinger: cool! is it hard to use here? I feel like because we catch the error ourselves we have to tell sentry about it you know?

GregWeil: yep, already switched in the remix. waiting until i get the other stuff fixed before updating the pr

src/presenters/pop-overs/team-user-info-pop.js
GregWeil: It's down on line 220, every function goes through the same `handleError` function, which just logs it, makes a notification, then throws the error back

sarahzinger: ahhh thanks!

GregWeil:
Added captureException calls, and hid users with pending invites from search results

sarahzinger:
cool stuff!

PR #542, opened 2019-05-21
Nova chance: Fix network error handling for description fields

GregWeil:
## Links
https://glitch.fogbugz.com/f/cases/3328526/Handle-network-errors-in-useOptimisticValue-hook

## Changes:
All description fields now use `handleErrorForInput` instead of `handleError`
`useOptimisticValue` expects error objects from failed promises instead of a string

## How To Test:
Try changing values on names and descriptions: successfully, api errors, and network errors.
The fixed bug can be reproduced by editing a description while disconnected from the internet.

## Feedback I'm looking for:
Every `handleErrorForInput` call is sent along to `useOptimisticValue`. Should `handleErrorForInput` get removed, since `useOptimisticValue` knows which errors it can show vs which go to a notification?


src/presenters/error-handlers.js
```diff
@@ -7,12 +7,11 @@ function handleError(notify, error) {
 }
 
 function handleErrorForInput(notify, error) {
-  if (error && error.response && error.response.data) {
-    return Promise.reject(error.response.data.message);
+  if (!(error && error.response && error.response.data)) {
+    console.error(error);
+    notify();
```
sarahzinger: I'm not getting notified of an error when I go offline, are you? it just looks like it worked which I guess is better than breaking the whole page, but I would guess we'd see like a nice little notification on the top right hand side of the page being like "oh it looks like you're offline" or something

GregWeil: added!

sarahzinger: should it show all the time you think? or just when you try to edit stuff?

sarahzinger: like right now:
- on a collections page: I go offline
- I see the notification
- I type in a note, it all looks ok like it saved? makes me think it might save offline or something
- I then go online and the notification goes away, the note still looks like it successfully saved
- I then refresh the page and the note is as it was before because it didn't actually save

THAT ALL SAID this might be out of scope for this pr? I guess I was expecting the notification to come on the save maybe, but maybe that's a bigger refactor/overkill

GregWeil: Hmm, there should be an on edit error message when you're offline. It takes a few seconds on ff windows for the first error to appear after going offline, but after that it's pretty much instant.

PR #545, opened 2019-05-21
Glossy wednesday: fix spellcheck issue on safari on email input

sarahzinger:
## Links
* https://glossy-wednesday.glitch.me/
* https://glitch.manuscript.com/f/cases/3328488/

## GIF/Screenshots:
http://g.recordit.co/pQsCdng8tG.gif

## Changes:
* stop spellcheck on email and password inputs. this is annoying because you'd think giving an input a type of email or password would fix this issue, but at least on safari on email inputs it's a no go. I tried passing  autoCapitalization=false (and also "off" and also "none" lol) but it turns out those are already done for you with type=email (as you'd expect!) unfortunately if your email contains your name it runs the spellchecker, which then autocorrects if you have a `.` right after your name (ex: `firstname.lastname@email.com` Fun fact, I also tried just turning off the autocorrect, but that doesn't seem to help.

## How To Test:
* try typing in the input field to sign in with email and ensure it doesn't auto capitalize anything.

### DIFF
--- a/src/components/inputs/text-input.js
+++ b/src/components/inputs/text-input.js
@@ -60,7 +60,7 @@ const TextInput = ({
           placeholder={placeholder}
           type={type}
           value={value}
-          autoCapitalize="none"
+          spellCheck={type !== 'email' && type !== 'password'}
         />
         {!!error && (
           <InputPart className={styles.errorIcon}>
--- a/src/components/inputs/text-input.js
+++ b/src/components/inputs/text-input.js
@@ -60,6 +60,7 @@ const TextInput = ({
           placeholder={placeholder}
           type={type}
           value={value}
+          autoCapitalize="none"
         />
         {!!error && (
           <InputPart className={styles.errorIcon}>


GregWeil:
This changes spellcheck on non email/password fields from unset to true. Will that override browser settings/defaults?

sarahzinger:
@GregWeil great question, just tested and I don't think so but let me know if you think I might be missing something:
- on safari set my settings to stop auto correcting spelling and stop spell checking. then tried to add a note on a project in a collection, and did not see any squiggles or auto suggestions when I typed `sarah`. put my settings back as they were before and tried again and it autocorrected and put some squiggles on there. Saw similar behavior in chrome

modernserf:
fwiw i can't reproduce the original bug on production; is there a system setting that needs to be enabled?

sarahzinger:
@modernserf I can't replicate it on my iphone, but I can replicate in safari on my macbook for whatever that's worth
Screen Shot 2019-05-23 at 10 58 21 AM

sarahzinger:
@GregWeil @modernserf just checking in on this one, I think it got lost in the shuffle, y'all think this is ok to go, or have other suggestions on improvements?

PR #547, opened 2019-05-22
Relieved fox: remove blue support banner and move it into user-options-pop

sarahzinger:
## Links
* https://relieved-fox.glitch.me/
* https://glitch.manuscript.com/f/cases/3328257/

## GIF/Screenshots:
Screen Shot 2019-05-22 at 11 01 29 AM
Screen Shot 2019-05-22 at 11 01 46 AM

## Changes:
* move super user helpers to current user so they are available anywhere we need them for the future
* remove "blue super user banner" which used to only show if your local storage was set a certain way. It was meant to allow super users to toggle their super powers more easily without altering the ui (with the option to hide for live streams). The issue with this approach was that it was tricky to remember how to set up when moving between remixes/devices.
* create a super user checkbox in user-options-pop. This is a permanent button for anyone with the power to turn super user features on/off. It will not show to users who don't have the ability to turn super user powers on/off. 

## How To Test:
* try turning your super user powers on/off via the user dropdown shown in the screenshots. If you don't have the ability to use super user features, ensure you do not see this dropdown

## Feedback I'm looking for:
- I'm using cached user here as that was the one that seemed to work in my testing (which makes sense given how we auth on remixes), but I'm a little fuzzy on the diffs between default user, shared user, and cached user, if I'm doing something wonky here definitely let me know!
- any reason I'm not thinking about on why this is a bad idea? any improvements we should make?

### DIFF
--- a/src/components/banners/super-user.js
+++ b/src/components/banners/super-user.js
@@ -9,7 +9,7 @@ const SuperUserBanner = () => {
   if (!superUserFeature) {
     return null;
   }
-  
+
   const expirationDate = superUserFeature && new Date(superUserFeature.expiresAt).toLocaleString();
   const displayText = `SUPER USER MODE ENABLED UNTIL: ${expirationDate} `;
   return (
--- a/src/components/banners/super-user.js
+++ b/src/components/banners/super-user.js
@@ -3,15 +3,13 @@ import classNames from 'classnames/bind';
 
 import { useCurrentUser } from 'State/current-user';
 
-import useLocalStorage from 'State/local-storage';
-
 import styles from './super-user.styl';
 
 const cx = classNames.bind(styles);
 
 const SuperUserBanner = () => {
   const { superUserHelpers } = useCurrentUser();
-  const { superUserFeature, canBecomeSuperUser, toggleSuperUser } = superUserHelpers;
+  const { superUserFeature, toggleSuperUser } = superUserHelpers;
 
   if (superUserFeature) {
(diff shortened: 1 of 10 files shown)


cori:
Also, I don't think I can test this from my iPad until it's in prod, since I can't authenticate myself to https://relieved-fox.glitch.me/ without access to the browser console - is that accurate or am I missing something?

sarahzinger:
@cori not too onerous at all :) updated so the checkbox will show in that circumstance though it doesn't do anything until we make the backend change in support-toggle, tbh, I'm not sure if we really want to let people toggle this stuff without being in the project? not sure if we're doing logging there for example. but I'll let the folks in charge of that repo decide :) 

regarding testing on ipad I'm not sure, but I can't think of a reason it wouldn't work there but would work on a desktop. Also it might be possible to see the console though it's probably not worth it? https://mattvanderpol.com/2013/08/14/how-to-debug-javascript-or-css-on-ios/ I've never done it on ipad but have done so on iphone once or twice. you have to plug it in to your laptop and everything lol.

cori:
That's cool, thanks. I'm pretty confident in the change I'm going to make to the support toggle but plan on running it past the Infrastructure folks - as I said it's a really minor nitpick, but I wanted folks to be able to disable even if they're been removed from the auth project so folks aren't stuck with the banner if they've been removed (since they also won't be able to use the Disable button on the banner in that case).

Probably not worth the effort on the iPad - I'll just check it once it gets to prod.

LGTM!

PR #549, opened 2019-05-22
Silken bandicoot

sarahzinger:
## Links
* https://silken-bandicoot.glitch.me
* https://sentry.io/organizations/fog-creek-software/issues/1035005423/events/5c4946d1391b4fb2958c39ce255591d5/?project=1246508&query=is%3Aunresolved

## Changes:
* check something is an array before we map over it. For whatever reason we sometimes have project.users = an error (when we get rate limited or when we throw a 404 what have you), which we then try to map 😭 which then throws an even more noticeable error. I think we need better error handling in general, but my sense is that is already being worked on (correct me if I'm wrong and I can try to take that on!) but for now it seemed like this would be a good bandaid for this particular situation.

## How To Test:
* you could try rate limiting yourself lol, but something I did was overwriting the response back from one of our api calls to undefined. 

## Feedback I'm looking for:
- is this a bad idea for reasons I'm not thinking about?

### DIFF
--- a/src/components/project/project-item.js
+++ b/src/components/project/project-item.js
@@ -8,6 +8,7 @@ import Image from 'Components/images/image';
 import ProfileList from 'Components/profile-list';
 import { ProjectLink } from 'Components/link';
 import AnimationContainer from 'Components/animation-container';
+import { captureException } from 'Utils/sentry';
 import { FALLBACK_AVATAR_URL, getAvatarUrl } from 'Models/project';
 import { getAllPages } from 'Shared/api';
 import { createAPIHook } from 'State/api';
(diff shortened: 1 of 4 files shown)


modernserf:
What kind of value is project.users if not an array? how can we ensure that its never set to a non-array?

sarahzinger:
@modernserf I believe it's an error instance. Specifically I think it's from when we fetch users by project id, I can wrap stuff in a try/catch if that's a better approach, didn't want to start messing too much with fetching users since I know you're actively refactoring, but can go down that path

modernserf:
also, is this related to this ticket? https://glitch.manuscript.com/f/cases/3328352/

sarahzinger:
@modernserf the way I'm reading these sentry bugs I think these are separate (but slightly related) issues. In the one I'm addressing in this pr, I see in the breadcrumbs that we always hit some kind of error before this is thrown usually with `projects/by/id/users` sometimes with `users/by/id` (usually a 429 or 404) in either case we're getting back an error instead of users but storing it like it's users. In the other sentry ticket (which hasn't happened in a long time for whatever it's worth so we should probs just close it not sure?) it's always after getting a 200 on projects/questions but similar to this issue it's not an array (which is maybe a backend bug that got fixed I wonder?)

The common theme being I don't think we're really handling our errors from our server properly. I can look into that if no one else is.

modernserf:
I think ultimately `createAPIHook` should do better error handling out of the box, but this should work in this circumstance

sarahzinger:
oh interesting 🤔 alright, I'll get this out and we can revisit, I'm sure we could def like streamline our error handling better.

PR #550, opened 2019-05-22
Profuse tea: Fix text fields trimming the value while you're typing

GregWeil:
## Links
https://glitch.fogbugz.com/f/cases/3328616/Don-t-trim-whitespace-from-the-field-in-useOptimisticText

## Changes:
- useOptimisticText tracks the input value and sends a trimmed copy to the api. If the value sent to the api hasn't changed, then it returns the untrimmed input value.
- I swapped the return order of useOptimisticValue and useOptimisticText, from [value,error,set] to [value,set,error] to be more like useState

## How To Test:
Try using the three different types of fields: text inputs, wrapping text inputs (collection name), and markdown areas. They should all still set the values correctly, not cut off whitespace while you are typing, and (excluding markdown) show an error message when empty.

### DIFF
--- a/src/components/fields/optimistic-markdown-input.js
+++ b/src/components/fields/optimistic-markdown-input.js
@@ -5,7 +5,7 @@ import MarkdownInput from '../inputs/markdown-input';
 import useOptimisticText from './use-optimistic-text';
 
 const OptimisticMarkdownInput = ({ value, onChange, ...props }) => {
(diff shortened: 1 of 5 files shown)


modernserf:
Since the order of arguments isn't obvious, I wonder if the hooks should return `{ value, onChange, onError }` instead?

PR #552, opened 2019-05-22
Spectrum splash: rebuild AddCollectionProjectPop

modernserf:
## Links
* Remix link: https://spectrum-splash.glitch.me
* Manuscript link: https://glitch.manuscript.com/f/cases/3327873, https://glitch.manuscript.com/f/cases/3328551/,
https://glitch.manuscript.com/f/cases/3328531/

## GIF/Screenshots
Screen Shot 2019-05-22 at 5 25 43 PM
Screen Shot 2019-05-22 at 5 26 03 PM
Screen Shot 2019-05-22 at 5 26 59 PM
Screen Shot 2019-05-23 at 10 51 06 AM

## Changes:
* use Algolia search API
* add filterTypes to algolia search
* componentize AddCollectionProjectPop
* componentize ProjectResultItem 
* add configurable size to ProfileList
* make ProjectResultItem load its own user list, and only when visible
* make results selectable with arrow keys
* fix tooltip overlaps with private projects

## How To Test:
* log into https://spectrum-splash.glitch.me (with email login or using the console)
* create a collection
* clicking "Add Project" should show you a list of your recent projects
* using the up & down arrow keys should highlight projects and scroll them into view
* pressing enter on a highlighted project, or clicking on one, should add it to your collection
* typing in the search box should search the whole community for projects
* you should also be able to paste in URLs of projects to find them
* it should tell you if you attempt to add a project that is already in the collection

## Feedback I'm looking for
* does this feel responsive?
* does it find the projects you'd expect to find?
* should we get rid of the "excluded N search results" warning and just show that a search result is already in the collection?


src/components/project/project-result-item.styl
```diff
@@ -0,0 +1,47 @@
...
+
+.resultInfo
+  padding-left: 15px
+  width: 100%
+  
+.active .resultName
+  color: inherit
+  
+.resultDescription
+  color: secondary
```
ttseng: the color here should be `text-on-secondary-background`

src/components/project/project-result-item.styl
```diff
@@ -0,0 +1,47 @@
...
+.projectResult
+  padding: 12px 70px 12px 6px
+  color: primary
+  background-color: primary-background
+  position: relative
+  &.private
+    background-color: private
+  &.active,
+  &:hover,
+  &:focus
```
ttseng: will want to add styling here so that .resultDescription `color: secondary-hover-oncolor` when projectResults is in active, hover, or focus state:
image

(correct styling shown on left)

src/components/project/project-result-item.js
```diff
@@ -0,0 +1,86 @@
...
+              </div>
+            )}
+            <div className={styles.profileListWrap}>
+              <ProfileList teams={teams} users={users} layout="row" size="small" />
+            </div>
+          </div>
+        </div>
+      </TransparentButton>
+      <div className={styles.linkButtonWrap}>
+        <Button size="small" href={getLink(project)}>
```
ttseng: The view-> link should open in a new tab rather than in the current window (so the user can confirm whether or not it's the right project before adding it to the collection).  
Should we add a prop to Button so specify whether the link should open in a new tab?

ttseng:
this is so much faster with algolia!  

I agree with removing the `Excluded x search results` message - I'm not seeing it where I would expect to see it anyhow (for example, if my query is `bongo-ca`, I don't end up seeing the message `Excluded x search results`).  I'm guessing this is because of the way Algolia returns the top results?  How were you able to trigger that message in the example above?

modernserf:
I was able to trigger the 'excluded X search results' message by matching a typo, which is not a particularly useful scenario.

PR #561, opened 2019-05-24
Superb guide: Reorder projects in collections

GregWeil:
## Links
https://glitch.manuscript.com/f/cases/3328236/API-Reorder-projects-in-collections

## Changes:
- You can drag around projects in collections you have permissions on
- Tab to focus a whole project item, then space, arrow keys to reorder using keyboard

## How To Test:
- Try reordering projects in a collection, both with and without a featured project
- Make sure other people's collection pages act as expected
- Make sure other grids of stuff still work in general 🤷‍♂ 

## Feedback I'm looking for:
- Projects with notes seem to animate to the wrong position. I can't figure out how to fix it
- I didn't make any changes to project order outside of the collection page. Should previews always show the first three projects, rather than the most recent?
- Draggable project items have their `tabIndex` set to `0`. What should their `role` be set to?

### DIFF
--- a/package.json
+++ b/package.json
@@ -93,6 +93,7 @@
     "react-pluralize": "^1.5.0",
     "react-router-dom": "^4.3.1",
     "react-select": "^2.4.0",
+    "react-sortable-hoc": "^1.9.1",
     "react-textarea-autosize": "^7.1.0",
     "speed-measure-webpack-plugin": "^1.3.1",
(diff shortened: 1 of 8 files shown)


modernserf:
The layout issues with notes are unfortunate but don't seem worth blocking over

keithk:
Missed this!

I think that previews should show the ones that you've ordered (that is what I'd want)

PR #565, opened 2019-05-28
Lumpy wedge: improve image upload error handling

sarahzinger:
## Links
* https://lumpy-wedge.glitch.me/
* https://glitch.manuscript.com/f/cases/3328077/

## GIF/Screenshots:
Screen Shot 2019-05-28 at 11 00 15 AM

## Changes:
* added a new error message for when folks hit a firewall
* fixed bug where sometimes users would hit an error and yet still get a notification that their image uploaded, I think because even if you a throw an error in a catch it still proceeds to the finally case (interestingly though I can't figure out how to recreate this without hardcoding in an error so maybe this isn't actually happening in real life).
* capture file upload exceptions with sentry directly, rather than rely on a global catch.

## How To Test:
* I added this code in the `try {}` not sure if we have a great way of forcing this particular error we're trying to protect around:
```
    let fake = new Error("i'm bad")
    fake.status_code = 0
    throw fake
```
for general error testing you can go offline which seems to work.

## Feedback I'm looking for:
- Do you generally agree with my check for a status_code? I'm basing this assumption on the details described in the manuscript case
- Do you agree with my thoughts on moving this out of a `finally` am I missing stuff here?

### DIFF
--- a/src/presenters/includes/uploader.js
+++ b/src/presenters/includes/uploader.js
@@ -2,6 +2,7 @@ import React from 'react';
 
 import { uploadAsset, uploadAssetSizes } from '../../utils/assets';
 import { useNotifications } from '../notifications';
+import { captureException } from './utils/sentry';
 
 const NotifyUploading = ({ progress }) => (
   <>
@@ -31,11 +32,9 @@ async function uploadWrapper(notifications, upload) {
       notifications.createNotification('Image uploaded!');
     });
   } catch (e) {
+    captureException(e);
     notifications.createErrorNotification(<NotifyError />);
     removeNotification();
-    // throw e;
-    
-    
   }
   return result;
 }
--- a/src/presenters/includes/uploader.js
+++ b/src/presenters/includes/uploader.js
@@ -11,7 +11,7 @@ const NotifyUploading = ({ progress }) => (
   </>
 );
 const NotifyError = ({ e }) => {
-  if (e && Object.hasOwnProperty.call(e, "status_code") && e.status_code === 0) {
+  if (e && Object.hasOwnProperty.call(e, 'status_code') && e.status_code === 0) {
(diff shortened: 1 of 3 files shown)


src/presenters/includes/uploader.js
```diff
@@ -9,7 +10,12 @@ const NotifyUploading = ({ progress }) => (
     <progress className="notify-progress" value={progress} />
   </>
 );
-const NotifyError = () => 'File upload failed. Try again in a few minutes?';
+const NotifyError = ({ e }) => {
+  if (e && Object.hasOwnProperty.call(e, 'status_code') && e.status_code === 0) {
```
modernserf: - what is the purpose of this `hasOwnProperty` check?
- is there a reason this prop isn't called `error`?

sarahzinger: - I guess I didn't want to make an assumption that our error had a status_code at all, and because we're checking to see if it's zero, I didn't want to write if (e.status_code) cuz zero is falsey, but maybe that check is not necessary?
- nope! will check it to error

oliviacpu: I think it's fine to leave out `hasOwnProperty`? I tried it out in the console and it seems to work the same both ways.

```
> const error = { status_code: 0 }
> const noError = { not_status_code: 0 }
> const checkStatusCode = (e) => { return e && e.status_code === 0 }
> const checkForStatusCode = (e) => { return e && Object.hasOwnProperty.call(e, 'status_code') && e.status_code === 0 }
> checkStatusCode(noError)
< false
> checkForStatusCode(noError)
< false
> checkStatusCode(error)
< true
> checkForStatusCode(error)
< true
```

sarahzinger: cool my bad, updated, thanks y'all

PR #562, opened 2019-05-24
Piquant mountain: Project Options Pop etc.

modernserf:
depends on #566,  #567, #596 

## Links
* Remix link: https://piquant-mountain.glitch.me/
* Manuscript link: https://glitch.manuscript.com/f/cases/3328622/

## GIF/Screenshots:
Screen Shot 2019-05-24 at 3 36 49 PM
Screen Shot 2019-05-24 at 3 49 30 PM
Screen Shot 2019-05-24 at 3 50 09 PM
Screen Shot 2019-05-24 at 3 50 21 PM
Screen Shot 2019-05-24 at 3 50 47 PM
Screen Shot 2019-05-24 at 3 50 58 PM
Screen Shot 2019-05-24 at 3 51 55 PM

## Changes:
* componentize ProjectOptionsPop
* componentize AddProjectToCollectionPop
* use Algolia for searching collections
* make collection results always searchable & navigable from keyboard
* componentize CreateCollectionPop
* use CreateCollectionPop for the Create Collection button on profile pages
* add short cache to API requests

## How To Test:
+ log into https://piquant-mountain.glitch.me/
  * visit your profile page
  * clicking on a project's options menu should show the project options popover
  * clicking on "Add to collection" should open a larger popover that lists collections to which you can add this project
  * typing in the search field should show filtered results
  * using the arrow keys should navigate the projects list
  * clicking on a collection result, or hitting enter when one is selected, should add the project to that collection
  * clicking on "Add to a new collection" should show the create collection popover
  * after entering a collection name, clicking "Create Collection" should create a collection with the selected project in it
+ visit your profile page
  * clicking on Create Collection should show the create collection popover (without the team selection dropdown)
  * after entering a collection name, clicking "Create Collection" should create a collection and redirect you to the page

PR #566, opened 2019-05-28
Hyper flyingfish: add caching to API requests

modernserf:
## Links
* Remix link: https://hyper-flyingfish.glitch.me/

## Why does this exist
This is extracted from #562. This short cache is mostly for deduplicating multiple identical requests in flight. We currently mostly do this with something like ProjectsLoader, but this is conceptually a lot simpler.

### DIFF
--- a/src/state/api.js
+++ b/src/state/api.js
@@ -8,17 +8,39 @@ import { captureException } from '../utils/sentry';
 export const Context = createContext();
 
 export const getAPIForToken = memoize((persistentToken) => {
+  const cache = {};
+  const maxAge = 60 * 1000;
+  let api;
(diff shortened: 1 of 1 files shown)


src/state/api.js
```diff
@@ -8,17 +8,39 @@ import { captureException } from '../utils/sentry';
...
+    });
   }
-  return axios.create({
-    baseURL: API_URL,
-  });
+
+  return {
+    ...api,
+    get: (url, config) => {
+      // TODO: support params
```
taravancil: I assume this isn't a blocking TODO, right?

modernserf: nope

PR #574, opened 2019-05-29
Amber ophthalmologist: api PropTypes.func -> any

modernserf:
Fixes some PropTypes errors introduced by the API caching PR.


modernserf:
@sarahzinger this is the error you spotted when reviewing my other PR

PR #576, opened 2019-05-30
Superb guide: collection reorder prompt + minor css cleanup + empty collection fixes

GregWeil:
## Links
* https://superb-guide.glitch.me/

## GIF/Screenshots:
image
image
image

## Changes:
- Added a prompt to the bottom of collections with more than one project explaining how to reorder, not happy with the wording
- Reworked some css to get rid of an `!important` I added during a deploy yesterday
- The empty collection check was broken, so the prompts weren't visible, this brings them back
- Project page 'in collections' area shows the first three projects in the collection. user/team pages still don't because they use the old api to load collections

## How To Test:
- Can you figure out how to reorder a collection without a mouse based on that prompt?
- Do the empty collection messages show up in the right situations?

## Feedback I'm looking for:
- The reorder prompt is there to indicate how to do so to people not using a mouse, since there's no real way to discover it on your own. Any ideas on a better way to do it than this?
- Do we still want those empty collection messages?

### DIFF
--- a/src/components/containers/projects-list.styl
+++ b/src/components/containers/projects-list.styl
@@ -17,7 +17,7 @@
 .projectsItem
   display: flex
   flex-direction: column
-  justify-content: end
+  justify-content: flex-end
   height: 100%
   
 .header
--- a/src/components/containers/projects-list.styl
+++ b/src/components/containers/projects-list.styl
@@ -15,7 +15,7 @@
   align-items: flex-end
 
 .projectsItem
-  display: flex !important
+  display: flex
   flex-direction: column
   justify-content: end
   height: 100%
--- a/src/presenters/collection-editor.js
+++ b/src/presenters/collection-editor.js
@@ -57,7 +57,7 @@ class CollectionEditor extends React.Component {
     const featuredIndex = this.state.projects.findIndex((p) => p.id === this.state.featuredProjectId);
     const index = (featuredIndex >= 0 && filteredIndex > featuredIndex) ? filteredIndex + 1 : filteredIndex;
     this.setState(({ projects }) => {
-      const sortedProjects = projects.filter((p) => p !== project);
+      const sortedProjects = projects.filter((p) => p.id !== project.id);
       sortedProjects.splice(index, 0, project);
       return { projects: sortedProjects };
     });
--- a/src/presenters/collection-editor.js
+++ b/src/presenters/collection-editor.js
@@ -40,7 +40,7 @@ class CollectionEditor extends React.Component {
       }));
     }
     await this.props.api.patch(`collections/${collection.id}/add/${project.id}`);
-    if (collection.id === this.state.id) {
+    if (collection.id === this.state.id && this.state.projects.length > 1) {
       await this.props.api.post(`collections/${collection.id}/project/${project.id}/index/0`);
     }
   }
--- a/src/presenters/collection-editor.js
+++ b/src/presenters/collection-editor.js
@@ -40,7 +40,7 @@ class CollectionEditor extends React.Component {
       }));
     }
     await this.props.api.patch(`collections/${collection.id}/add/${project.id}`);
-    if (collection.id === this.state.id && this.state.projects.length > 1) {
+    if (collection.id === this.state.id) {
       await this.props.api.post(`collections/${collection.id}/project/${project.id}/index/0`);
     }
   }
--- a/src/presenters/pages/collection.js
+++ b/src/presenters/pages/collection.js
@@ -78,7 +78,7 @@ const CollectionPageContents = ({
   unfeatureProject,
   ...props
 }) => {
-  const collectionHasProjects = !!collection && !!collection.projects && !!collection.projects.length;
+  const collectionHasProjects = !!collection && !!collection.projects;
   let featuredProject = null;
   let { projects } = collection;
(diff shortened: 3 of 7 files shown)


src/presenters/pages/collection.js
```diff
@@ -127,58 +127,57 @@ const CollectionPageContents = ({
...
+                  addProjectToCollection,
+                  displayNewNote,
+                  featureProject,
+                  isAuthorized: currentUserIsAuthor,
+                }}
+                fetchMembers
+              />
+            )}
+            {currentUserIsAuthor && projects.length > 1 && <div>Drag to reorder, or press space with a project focused</div>}
+          </div>
```
GregWeil: Might want to ⚙ > hide whitespace changes for this

src/presenters/pages/collection.js
```diff
@@ -127,58 +127,57 @@ const CollectionPageContents = ({
...
-                    addProjectToCollection,
-                    displayNewNote,
-                    featureProject,
-                    isAuthorized: currentUserIsAuthor,
-                  }}
-                  fetchMembers
-                />
+            {!collectionHasProjects && currentUserIsAuthor && (
+              <div className="empty-collection-hint">
+                <Image src="https://cdn.glitch.com/1afc1ac4-170b-48af-b596-78fe15838ad3%2Fpsst-pink.svg?1541086338934" alt="psst" width="" height="" />
```
GregWeil: Images of text seems like a prime candidate for alt text.

GregWeil:
Updated. Now collections won't call the reorder api when they only have one project, which would cause a 500 error.
- The first project added to a collection doesn't bother trying to bump itself to the top
- Drag and drop is disabled until there are two non-featured projects in the collection

GregWeil:
(I'm on vacation until the 11th, please deploy without me if necessary)

src/components/containers/projects-list.styl
```diff
@@ -15,7 +15,7 @@
   align-items: flex-end
 
 .projectsItem
-  display: flex !important
```
sarahzinger: do we remember why we had this in the first place?

src/presenters/collection-editor.js
```diff
@@ -40,7 +40,7 @@ class CollectionEditor extends React.Component {
       }));
     }
     await this.props.api.patch(`collections/${collection.id}/add/${project.id}`);
-    if (collection.id === this.state.id) {
+    if (collection.id === this.state.id && this.state.projects.length > 1) {
```
sarahzinger: pretty sure this bug was fixed btw

src/presenters/pages/collection.js
sarahzinger: not sure I totally follow this comment, mind explaining more?

sarahzinger:
>The reorder prompt is there to indicate how to do so to people not using a mouse, since there's no real way to discover it on your own. Any ideas on a better way to do it than this?

I think there are ways of giving users instructions only when they are using keyboards to nav, but I also think lots of people appreciate a good keyboard shortcut so having it displayed they way you have here makes sense to me! I might suggest we rephrase it since it wasn't clear to me what it was referring to exactly maybe: 
Drag to reorder, or tab to focus a project then press spacebar, followed by arrow keys
or something like that? @clottman you got thoughts?

src/presenters/pages/collection.js
GregWeil: the diff is mostly indentation, so if you filter those out (using the gear menu at the top of the list of files) it'll only show the two or three lines that are actually different

clottman:
It looks like sometimes the layout isn't reflowing correctly if some items have notes and others don't. 
image

clottman:
Also one thing that wasn't clear to me from the description of what to do is that you need to press space again to save it. I refreshed and was surprised to see my changes didn't stick. 

What about something like this? It's a little longer but more clear.
image
> Drag to reorder, or move focus to a project, press space and use the arrow keys to move that project. Press space again to save your new order.

src/presenters/collection-editor.js
GregWeil: Looks like it, I'll remove that check

src/components/containers/projects-list.styl
GregWeil: Nope, but removing it doesn't seem to break anything

GregWeil:
react-sortable-hoc doesn't handle variable-sized elements well, I don't think we can control it that much.

I added a 'space to save' bit to the text, but maybe it should just save as you move the project around. I tried playing with it a bit but it doesn't like the items changing at the same time as the user is doing stuff. Maybe it can be done with some optimisticvalue style logic, where it ignores changes to the items prop during a reorder.

GregWeil:
Now it'll save as you are moving items around, but only on the keyboard. Saving mid mouse drag makes the movement a bit stuttery, and letting go of the mouse is intuitive.

GregWeil:
Extra change to fix project item alignment on chrome when there's mixed notes/not notes.

sarahzinger:
seeing some funky behavior with notes when I have a blank note and try to move it before I unfocus the note it'll make a copy of the project for some reason, notice I started messing around with "Jade Meat" here and it made 3 copies: 
http://recordit.co/UGVhLMS04K

GregWeil:
Nice catch! Should be fixed now.

PR #580, opened 2019-06-03
Cassey-community: accessibility improvements to pagination control

clottman:
## Links
* [Remix link](cassey-community.glitch.me)
* Manuscript - [Focus should go to previous button after clicking last possible 'next' button](https://glitch.manuscript.com/f/cases/3328645/Focus-is-lost-when-you-click-the-next-button-in-a-paginated-project-list-and-arrive-at-the-last-page) and [announcement for screenreader users that page content changed when they pressed pagination buttons](https://glitch.manuscript.com/f/cases/3328694/Screenreader-users-should-hear-an-announcement-that-clicking-the-pagination-buttons-caused-the-visible-projects-to-change)

## GIF/Screenshots:
image

## Changes:
* Adds a space between 'Show all' and 'badge' so screenreader users hear 'show all X' instead of 'show allx', removes some margin to account for this. Didn't see anywhere else this would be visually affected. 

## How To Test:
* Paginate to the end of your recent projects list with the mouse or keyboard; focus should return to the "Previous" button when you get to the end of the list. If running voiceover or another screenreader, you should hear 'Page x of y' on every page after you press the button. You should also hear 'Showing all projects' after you click the 'show all' button.

## Feedback I'm looking for:
- Reducer use look correct?


src/components/containers/projects-list.js
```diff
@@ -48,16 +49,54 @@ const ProjectsUL = ({ collection, projects, sortable, onReorder, noteOptions, la
...
+      return {
+        page: oldState.page + 1,
+        totalPages: oldState.totalPages,
+        announce: `Showing page ${oldState.page + 1} of ${oldState.totalPages}`,
+      };
+    case 'previous':
+      return {
+        page: oldState.page - 1,
+        totalPages: oldState.totalPages,
+        announce: `Showing page ${oldState.page - 1} of ${oldState.totalPages}`,
```
sarahzinger: i could see an argument to remove total pages and announce out of state and move that into the view. BUT ALSO I don't know if that really matters? this seems pretty clear to me as is

src/components/containers/projects-list.js
```diff
@@ -48,16 +49,54 @@ const ProjectsUL = ({ collection, projects, sortable, onReorder, noteOptions, la
...
+  const [state, dispatchState] = useReducer(paginationReducer, {
+    page: 1,
+    totalPages: numPages,
+    announce: '',
+  });
+  const prevButtonRef = useRef();
+
+  const canPaginate = enabled && !state.expanded && projectsPerPage < numProjects;
+
+  const onNextButtonClick = () => {
```
sarahzinger: do we need a similar one for onPreviousButtonClick so that it moves focus to the next button when the previous button is disabled because you've moved all the way back to the beginning?

clottman: oh good catch! Yes, I'll add that.

PR #582, opened 2019-06-04
Oceanic ornament 1: preventing tabs from outside of the new stuff dialog

sarahzinger:
## Links
* https://oceanic-ornament-1.glitch.me/
* https://glitch.manuscript.com/f/cases/3328202/

## Changes:
* Previously when you opened the "new stuff" pupdate and you hit tab you'd tab through the modal, but then you'd also tab through the other elements behind the modal. This pr makes it so that when the new stuff update modal is open you tab through all of the focusable elements available in a circle and to exit you hit escape. 

## How To Test:
* you can open the pupdate by logging in and clicking on your avatar on the upper right hand side and clicking "new stuff" in the dropdown, then try to tab around, at the moment you should be stuck on the checkbox as that's the only focusable element there. but if you add another button you should see that it's tab-able. to exit hit escape. 

## Feedback I'm looking for:
- does this functionality seem right? I implemented it because that's what was outlined in the case and in [this article](https://bitsofco.de/accessible-modal-dialog/) and sounded right but as I play around with it I wonder if instead of going in a circle of focusable elements it should instead auto-close the modal and return your focus to the pup once you've gone through all the focusable elements? I can see an argument for a circle if it had some kind of required acknowledgement in it like a terms of service, but in this case maybe this is making it more difficult to use? what do we think?
- do we have other dialog boxes like this where we'd want this kind of functionality? i did a brief search and it seemed like no?
- am I doing anything funky with useEffect/useRef/useState? I'm still getting a feel for them. 
- I could be wrong but in my testing it seemed like the other accessibility issues with the new stuff dialog outlined in the cases had been fixed already, but definitely let me know if I missed anything!

- [x] try chrome
- [x] test firefox
- [x] test safari
- [x] test edge

### DIFF
--- a/src/presenters/overlays/new-stuff.js
+++ b/src/presenters/overlays/new-stuff.js
@@ -16,48 +16,10 @@ import newStuffLog from '../../curated/new-stuff-log';
 
 const latestId = Math.max(...newStuffLog.map(({ id }) => id));
 
-const useRestrictKeyboardFocusToDialog = () => {
-  const [focus, setFocus] = React.useState(0);
-  const ref = React.useRef();
-
-  React.useEffect(() => {
-    const dialog = ref.current;
-    if (dialog) {
-      const focusableElements =
-        'a:not([disabled]), button:not([disabled]), input:not([disabled]), [tabindex]:not([disabled]):not([tabindex="-1"]), select:not([disabled]), textarea:not([disabled])';
-      const focusableDialogElements = dialog.querySelectorAll(focusableElements);
-      const focusableItems = [...focusableDialogElements];
-      const keyHandler = (event) => {
-        // user tabs only within the elements of the dialog in a circle (not anything outside of it)
-        if (['Tab'].includes(event.key)) {
-          event.preventDefault();
-          event.stopPropagation();
-
-          let newFocus;
-          if (event.shiftKey) {
-            // tab backwards
-            newFocus = focus - 1 >= 0 ? focus - 1 : focusableItems.length - 1;
-          } else {
-            // tab forwards
-            newFocus = focus + 1 < focusableItems.length ? focus + 1 : 0;
-          }
-
-          focusableItems[newFocus].focus();
-          setFocus(newFocus);
-        }
-      };
-      window.addEventListener('keydown', keyHandler);
-      return () => window.removeEventListener('keydown', keyHandler);
-    }
-    return () => {};
-  }, [focus]);
-
-  return ref;
-};
-
-export function usePreventTabOut(first, last) {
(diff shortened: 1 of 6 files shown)


clottman:
One way we could make this even better is if we added a visible & focusable button to the overlay that represented 'close' and did the same thing as pressing the escape key does. (Dismisses the dialog). 

https://www.w3.org/TR/wai-aria-practices/#dialog_modal 

A few more things from the above guideline: 

- set aria-modal=true on the <dialog> element
- set a value for aria-labelledby on the <dialog> that refers to the visible title. (the h1 that says "New Stuff")

src/presenters/overlays/new-stuff.js
```diff
@@ -15,20 +15,53 @@ import newStuffLog from '../../curated/new-stuff-log';
...
+    }
+    return () => {};
+  }, [focus]);
+
+  return ref;
+};
+
+const NewStuffOverlay = ({ setShowNewStuff, showNewStuff, newStuff }) => {
+  const [focus, setFocus] = React.useState(0);
+  const newStuffOverlayRef = useRestrictKeyBoardFocusToDialog(focus, setFocus);
```
clottman: What is the `focus` in state doing? It looks like we're keeping it set but not ever reading from it.

sarahzinger: we do! it's an index that represents which of the navigable elements in the dialog we're in so it knows what the next element is to focus on, and cycles through (check out focus/setFocus in "useResetrictKeyBoardFocusToDialog"

although now that I look at this again it seems like if we are to continue with this route we have to think about navigating backwards... I think I might have missed that use case 🤦‍♀

clottman: I think if we're just using it in the other hook, that's probably where the state of which element is in focus should live.

sarahzinger:
to my knowledge I don't think we really have an `x` to close example anywhere to go off of...kept looking for something that might work that to match the styles and didn't find anything... But then I thought maybe our Back to Glitch button that we use in the editor might be appropriate?
Screen Shot 2019-06-06 at 2 13 45 PM

Feel free to let me know if you'd rather have something else though! @ttseng mind taking a look and lending me your thoughts?

sarahzinger:
@clottman thanks for sharing @chancedickson 's solution, liked a lot more than my own so I added it here! Gotta get that component library going lol :) Unless I hear otherwise gunna assume this is good to go and try shipping out on monday.

ttseng:
The Back to Glitch button looks good!  Also a great reminder that we have so many pupdates to write!

PR #585, opened 2019-06-05
sweet-ethernet: don't pass props to DOM elements

taravancil:
## Links
* [sweet-ethernet.glitch.me](https://sweet-ethernet.glitch.me)

## Changes:
This removes `{...props}` from an `<article>` DOM element. I think this is leftover from refactoring, we shouldn't need to pass props to a native DOM element, and it was throwing a console error that you can see at [lavish-laborer.glitch.me/@glitch](https://lavish-laborer.glitch.me/@glitch)

## How To Test:
Open [sweet-ethernet.glitch.me/@glitch](https://sweet-ethernet.glitch.me/@glitch), you shouldn't see any errors in the console

## Feedback I'm looking for:
Does this unexpectedly break anything? Is there actually a good reason to pass `props` to DOM elements?

### DIFF
--- a/src/components/containers/projects-list.js
+++ b/src/components/containers/projects-list.js
@@ -158,7 +158,7 @@ function ProjectsList({
   return (
     <FilterController enabled={enableFiltering} placeholder={placeholder} projects={projects}>
       {({ filterInput, renderProjects }) => (
-        <article {...props} className={classNames(styles.projectsContainer)}>
+        <article className={classNames(styles.projectsContainer)}>
           <div className={styles.header}>
             {title && <Heading tagName="h2">{title}</Heading>}
             {filterInput}
--- a/src/components/containers/projects-list.js
+++ b/src/components/containers/projects-list.js
@@ -153,11 +153,12 @@ function ProjectsList({
   collection,
   noteOptions,
   projectOptions,
+  dataCy,
 }) {
   return (
     <FilterController enabled={enableFiltering} placeholder={placeholder} projects={projects}>
       {({ filterInput, renderProjects }) => (
-        <article className={classNames(styles.projectsContainer)}>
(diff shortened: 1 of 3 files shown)


sarahzinger:
@taravancil I could be wrong but I think we're doing this for our cypress tests? @oliviacpu whatchu think?

clottman:
If it's for our cypress tests, is it because cypress needs some props but not all? We should maybe just be passing the ones we want (if we can identify those) and not the ones that we don't need and aren't valid.

src/components/containers/projects-list.js
```diff
@@ -153,12 +153,11 @@ function ProjectsList({
   collection,
   noteOptions,
   projectOptions,
-  ...props
 }) {
   return (
     <FilterController enabled={enableFiltering} placeholder={placeholder} projects={projects}>
       {({ filterInput, renderProjects }) => (
-        <article {...props} className={classNames(styles.projectsContainer)}>
+        <article className={classNames(styles.projectsContainer)}>
```
oliviacpu: `presenters/pages/user.js` uses `data-cy="pinned-projects"` to distinguish the different uses of `ProjectsList`

We could set `data-cy={cypressTag}` or allow the individual uses of these to be identified in another way. I just never cleaned this up once the tests were kind of working.

oliviacpu: Related Cypress documentation: https://docs.cypress.io/guides/references/best-practices.html#Selecting-Elements

clottman: Ok awesome - so sounds like we just need to pass in data-cy={cypressTag} and the rest of the props can get left out.

taravancil: ok cool! @oliviacpu can we pair on this for a few minutes today or tomorrow?

PR #586, opened 2019-06-05
Hexagonal gecko: color contrast warnings

sarahzinger:
## Links
* https://hexagonal-gecko.glitch.me/
* https://glitch.manuscript.com/f/cases/3328637/

## GIF/Screenshots:
Screen Shot 2019-06-05 at 2 59 12 PM

## Changes:
* shows a warning when the user picks a color with not great color contrast
* picks a different color if random color picker picks a low contrast color
* note: it still lets you set not high contrast colors, it won't warn you that a color is not great until you edit the input field

## How To Test:
* go to your collections and try picking different colors of varying contrasts

## Feedback I'm looking for:
- is this error message too ugly? is there a terser/friendlier/clearer message you'd prefer?
- I recognize that recursively calling getRandomColor sounds a little dangerous but it's pretty rare for it to find a color that doesn't pass the test, it took like 30+ tries for me to hit a bad color, feel free to suggest a better way around this tho. 
- any other bugs with this thing you notice while we're looking at it? 

PS: hilarious that this the project is HEX-agonal gecko right?


src/presenters/pop-overs/edit-collection-color-pop.js
```diff
@@ -14,14 +16,20 @@ const validHex = (hex) => {
   return false;
 };
 
+const isGoodColorContrast = (hex) => {
+  const contrastHex = isDarkColor(hex) ? '#fff' : '#222';
```
sarahzinger: i do wonder if there's a way to just update isDarkColor in someway to ensure we always pick a high contrast color and then we don't need this? I'm guessing no? I don't actually know enough about colors to feel confident I know the answer to this one

oliviacpu:
`is there a terser/friendlier/clearer message you'd prefer?` - We could just not let people use bad contrast ratios, but I still don't know a nice message to convey that.

taravancil:
Oh this is cute! Spitballing some error messages:

"Color contrast too low. Try another!"
"Color contrast too low. Try another color!"

clottman:
Maybe just "The contrast between this color and the default text color may be hard for some folks to see." as the error message? Makes it clear why low contrast is a problem.

taravancil:
@clottman Looking at the screenshot I think we may want to aim for as few letters as possible

sarahzinger:
ok gang, @oliviacpu @taravancil @clottman 
what do y'all think of this:
Screen Shot 2019-06-05 at 4 22 54 PM

I also took olivia's suggestion and prevented the user from selecting colors that are hard to read, like with invalid hex, we early return before setting it. Seem ok?

src/presenters/pop-overs/edit-collection-color-pop.js
clottman: What if we tried both `#fff` and `#222` and use the one that has high-enough contrast if one does, otherwise show the error and prevent submitting?

src/presenters/pop-overs/edit-collection-color-pop.js
```diff
@@ -38,26 +46,36 @@ class EditCollectionColorPop extends React.Component {
...
-      } else {
-        this.setState({ error: true });
+      const hexIsValid = validHex(query);
+      if (!hexIsValid) {
+        this.setState({ error: 'Invalid Hex' });
+        return;
+      }
+      const hexIsGoodColorContrast = isGoodColorContrast(query);
+      if (!hexIsGoodColorContrast) {
+        this.setState({ error: 'This color might make things hard to read, try another!' });
```
clottman: Instead of 'things' how about 'notes' or 'text'?

src/presenters/pop-overs/edit-collection-color-pop.js
sarahzinger: ohhh good thought! will try that thanks :)

sarahzinger:
Thanks for the feedback y'all this is what I have now:
Screen Shot 2019-06-06 at 10 29 08 AM
(also fixed this bug where using back-ticks on a dark background was nearly impossible to read)
Screen Shot 2019-06-06 at 10 24 07 AM
We now use .dark styles depending on whether the background color has a higher contrast with it or white... presumably this works exactly the same as what we have in real life but it's possible the calculations are somewhat different, if you have good test colors let me know!

ttseng:
Nice work on this!  I'd suggest changing the error message to two sentences, since the comma introduces a fragment.  `This color might make text hard to read. Try another!`

PR #589, opened 2019-06-06
tundra-beat: Two-Factor Authentication

oliviacpu:
## Links
* Remix link: https://tundra-beat.glitch.me/
* Manuscript link
* Specs/Notion docs: https://www.notion.so/glitch/Authentication-UI-2c0a7259704b486f83c525e9ab33e1d6

## GIF/Screenshots:
image
image
image
image
image
image

## Changes:
* Adds "User Passwords" feature flag to `/secret`
* Adds "Account Settings" to User Options popover
* Adds "Set Password" to Account Settings if no password is set
* Adds "Change Password" to Account Settings if a password is set
* Adds "Reset Password" to Account Settings and Sign In popover
* Adds "Enable Two-Factor Authentication" to Account Settings if not enabled
* Adds "Disable Two-Factor Authentication" to Account Settings if enabled
* Adds support for signing in with two-factor authentication enabled

## How To Test:
* Currently, `https://tundra-beat.glitch.me/secret` and enable "User Passwords"
* Check that you can set, update, and reset passwords
* Check that you can enable and disable two-factor authentication

## Feedback I'm looking for:
- ?

- [ ] Review and test?
- [ ] Remove endpoint check for password enabled and use property on user once merged


src/components/account-settings-overlay/password-settings.js
```diff
@@ -0,0 +1,94 @@
...
+
+const PasswordSettings = () => {
+  const api = useAPI();
+  const { currentUser, reload } = useCurrentUser();
+
+  const [oldPassword, setOldPassword] = useState('');
+  const [newPassword, setNewPassword] = useState(null);
+  const [passwordVersion, setPasswordVersion] = useState(1);
+
+  const [state, setState] = useState(null);
```
sarahzinger: would it make sense to to combine all of these setStates into one setState with an object? Or maybe useReducer? 

alternatively maybe we could name this updateStatusState or something rather than a generic setState?

src/components/account-settings-overlay/password-settings.js
```diff
@@ -0,0 +1,94 @@
...
+  const resetPassword = async (event) => {
+    event.preventDefault();
+    setResetState('working');
+    try {
+      await api.post('email/sendResetPasswordEmail', {
+        emailAddress: primaryEmail.email,
+      });
+      setResetState('done');
+    } catch (error) {
+      console.error(error);
```
sarahzinger: maybe we should use captureException from sentry instead?

src/components/buttons/button.js
```diff
@@ -98,6 +97,8 @@ Button.propTypes = {
   type: PropTypes.oneOf(TYPES),
   /** size of button */
   size: PropTypes.oneOf(SIZES),
+  /** button submits a form */
+  submit: PropTypes.bool,
```
sarahzinger: I think I'm missing why this is needed and a distinction from a regular onclick?

src/components/buttons/sign-in-button.js
```diff
@@ -0,0 +1,77 @@
...
+    <div style={{ marginBottom: '10px' }}>
+      <Button href={href} onClick={onClick} size="small" emoji={emoji}>
+        Sign in with {name}
+      </Button>
+    </div>
+  );
+};
+
+SignInButton.propTypes = {
+  company: PropTypes.oneOf(Object.keys(companies)).isRequired,
```
sarahzinger: this whole file is super easy to follow, also I love this way of doing proptypes, never thought to do that!

src/components/layout/auth-layout.styl
```diff
@@ -0,0 +1,43 @@
...
+    padding-right: 20px
+
+.container
+  display: flex
+  align-items: center
+  justify-content: center
+  flex-direction: column
+  margin: 20px 0
+
+.content {
```
sarahzinger: maybe no `{}` and no `;`  actually I can't remember what we normally do lol. seems weird we don't lint these files to have a consistent look? lol a ticket for another day.

src/components/new-password-input/index.js
```diff
@@ -0,0 +1,115 @@
...
+  if (!weakPWs.includes(password)) {
+    const hasCapScore = /^(?=.*[A-Z])/.test(password) ? 1 : 0;
+    const hasNumScore = /^(?=.*\d)/.test(password) ? 1 : 0;
+    const hasCharScore = /[!@#$%^&*(),.?":{}|<>]/.test(password) ? 1 : 0;
+    pwStrength = hasCapScore + hasNumScore + hasCharScore;
+  } else {
+    weakPasswordError = true;
+  }
+
+  const [didFirstEffect, setDidFirstEffect] = React.useState(false);
```
sarahzinger: why do we need to run the effect once before calling the onChange?

src/components/new-password-input/index.js
```diff
@@ -0,0 +1,115 @@
...
+        type="password"
+        labelText="password"
+        placeholder="new password"
+        onChange={setPassword}
+        disabled={disabled}
+        error={weakPasswordError ? weakPWErrorMsg : null}
+      />
+
+      {password.length >= pwMinCharCount ? (
+        <PasswordStrength strength={pwStrength} />
```
sarahzinger: I wonder if we should wrap this in an aria-live or something so folks with screenreaders would get this message or if that would be disruptive/annoying?

src/components/sign-in/two-factor-form.js
```diff
@@ -0,0 +1,73 @@
...
+import { useCurrentUser } from 'State/current-user';
+
+import styles from './styles.styl';
+
+const TwoFactorSignIn = ({ initialToken, onSuccess }) => {
+  const api = useAPI();
+  const { login } = useCurrentUser();
+  const [code, setCode] = React.useState('');
+  const [token, setToken] = React.useState(initialToken);
+  const [status, setStatus] = React.useState({});
```
sarahzinger: I see we do this in a variety of ways in our components, but I think this format is my fave

src/presenters/pages/login.js
```diff
@@ -50,20 +51,23 @@ const LoginPage = ({ provider, url }) => {
...
+      if (data.tfaToken) {
+        setTwoFactor(data.tfaToken);
+      } else if (!data.id || data.id <= 0) {
         throw new Error(`Bad user id (${data.id}) after ${provider} login`);
-      }
-
-      console.log('LOGGED IN', data.id);
-      login(data);
+      } else {
+        console.log('LOGGED IN', data.id);
```
sarahzinger: is it ok to leave these console.logs in? I guess we already do?

styles/overlays.styl
```diff
@@ -40,4 +40,19 @@ details.overlay-container[open] > summary::after
...
   .wistia_embed
     height: 100%
-    width: 100%
\ No newline at end of file
+    width: 100%
+  .button
+    display: inline
+
+// user settings overlay
+.account-settings-overlay
```
sarahzinger: is there no other way than to add these styles here?

sarahzinger:
I tried creating a new test account by signing in because I saw `New to Glitch? Create an account by signing in.` along with a new email/password section, and it gave me a 401 error, that seems not desirable yeah?

src/components/buttons/button.js
GregWeil: `<Button submit>` doesn't need an onclick at all, since the event handling lives on an outer form. This also sets the html `<button type="submit">`

src/components/new-password-input/index.js
GregWeil: The effect will always get called once on component mount. This ignores that and waits for the next call, when the value has actually changed.

src/components/buttons/button.js
sarahzinger: ahhh yeah ok now it clicks for some reason thanks!

src/components/new-password-input/index.js
GregWeil: I don't think there's any real danger to calling onChange early, I'll try removing it

GregWeil:
Thanks for all the feedback!. I didn't hit everything yet, but got some of them

package.json
```diff
@@ -72,6 +72,7 @@
     "group-by-time": "^1.0.0",
     "helmet": "^3.16.0",
     "html-truncate": "^1.2.2",
+    "intersection-observer": "^0.7.0",
```
sarahzinger: are we using this?

PR #595, opened 2019-06-07
Pool sweatpants: resend invites for team users, delete invited user

whimsicallyson:
## Links
* Remix link: https://pool-sweatpants.glitch.me/
* Manuscript link:

 related to https://glitch.manuscript.com/f/cases/3326592/teams-persistently-see-avatars-of-people-that-have-been-invited-to-a-team
* Specs/Notion docs: https://www.notion.so/glitch/87bda09f85184c1d81e1659aa62e4392?v=9f7963becc9341ed9af502682c6e9eef&p=7de6515edd654496af164018eb597c0a

## GIF/Screenshots:
Screen Shot 2019-06-07 at 2 55 37 PM
## Changes:
* Clicking an invited user will now get you a popup (previously went to their profile)
* That popup has link to profile, button to resend invite email, button to remove user from team

## How To Test:
* Add members to your favorite team, try the buttons to resend invites and revoke access
* Note that 429 rate limiting is still in effect for the emails

## Feedback I'm looking for:
* Does it work?
* Did I component it right? (if so, all accolades should go to Tara)
* Is anything missing?

- have a dance party? IDK...


sarahzinger:
rad stuff! ran into some small bugs: http://recordit.co/jbPTj2V6rG
- I think it's a little odd the green success message is over my avatar right? should it be on the upper right? I don't actually know what we usually do but seemed off
- when I removed myself from the team my avatar was still there?

If you want to practice component-ing you could see if you could trying writing a story for storybook in for your component (unless maybe someone's already working on it? not sure!) I tried to run `npm run storybook` in your branch and it hit an error, but maybe if you pull in master the bug would be fixed not sure?

taravancil:
Ah yeah, we have the concept of an inline notification, which I think the green one is. Maybe @ttseng has opinions about where each of these notifications should go?

Whoops @whimsicallyson, I totally forgot about adding this to storybook when we paired, but we should def add it there

whimsicallyson:
updates: 
- Moved notification when adding a user to team to the upper-right corner, to match with other notifications. Also updated it to use a `Notifications` component.
- Avatars should now disappear when invitees are removed from a team
- Added some error messages around adding users, since I noticed we didn't have them yet
- Started adding `TeamUsers` storybook component, but as notifications currently aren't in the storybook, it doesn't work (much like the header storybook components don't work)

sarahzinger:
I'm noticing a difference in behavior between prod and your remix, when I invite a new user with a throw away email address I found on prod I get this nice suggestion to invite the new user, when I use your remix I do not, I'll post some screenshots:
Screen Shot 2019-06-12 at 2 54 10 PM
Screen Shot 2019-06-12 at 2 54 26 PM
might just be that you need to pull to get the latest code? not sure if these are different popovers at this point?

PR #599, opened 2019-06-10
delicious-pulsar: Move notifications provider to src/state, componentize notifications

taravancil:
## Links
* [delicious-pulsar.glitch.me](https://delicious-pulsar.glitch.me)
* [Manuscript](https://glitch.fogbugz.com/f/cases/3328583/)

## Changes:
* Hooks-ify the previous implementation of notifications
* Move Notifications provider to `src/state`
* Convert instances of `<div class="notification">` to `<Notification>`
* Update API for creating notifications imperatively:
  * I removed `createPersistentNotification`, and added an `opts` argument with a `persistent` property
  * I kept `createErrorNotification` for the [convenience of having a fallback value for the `content` argument](https://github.com/FogCreek/Glitch-Community/compare/delicious-pulsar?expand=1#diff-253edf4bc0c7c040aa8856a4309af3caR27), but maybe we could get rid of that too?
  * The new API is:
    `createNotification(content, { type: 'success' | 'error' | 'info', persistent: Boolean, inline: Boolean })`

## How To Test:
Trigger some notifications! I'll run through a thorough list of all notifications before deploy, but if you want to do some basic testing try these:

- Add a project to a collection from a project page
- Turn your wifi off
- Log in via email
- Invite someone to a team

## Feedback I'm looking for:
- Is there anything I missed in moving the provider to `/src/state`
- On manuscript, I'm not entirely sure what @modernserf means by "incorporate error handlers", can anyone clarify what that might involve?

- [ ] Thoroughly QA all notifications


clottman:
Once this is in and so is my other PR that adds the react-aria-live plugin, next step should be to put these notifications in a <LiveMessage> wrapper so they get announced to screenreader users when they show up!

taravancil:
@clottman Yes! I thought about figuring out how to make notifications a11y-compliant as I was working on this, but I realized it was already a big PR and might become too difficult to review

src/components/notification/styles.styl
```diff
@@ -0,0 +1,51 @@
...
+    border: 0
+    background-color: primary-background
+  progress
+    margin-left: 5px
+  .loader
+    width: 20px
+    height: 19px
+    margin: 0
+    margin-right: 5px
+  &:hover,
```
clottman: Do we need &:focus on this too?

src/state/notifications.js
```diff
@@ -0,0 +1,74 @@
...
+
+  const remove = (id) => {
+    setNotifications((prevNotifications) => prevNotifications.filter((n) => n.id !== id));
+  };
+
+  const create = (content, opts = {}) => {
+    const { type, inline, persistent } = opts;
+    const notification = {
+      id: `${Date.now()}{Math.random()}`,
+      type: type || 'info',
```
clottman: This is like setting the default value of the param twice, right? Maybe we should just let the <Notification> component handle it?

src/state/offline-notice.js
```diff
@@ -1,14 +1,17 @@
...
+const PersistentNotification = ({ children }) => {
+  const { createNotification } = useNotifications();
 
-  useEffect(() => {
-    const { removeNotification } = createPersistentNotification(children, className);
-    return removeNotification;
-  }, [children, className]);
+  useEffect(
+    () => {
+      const { removeNotification } = createNotification(children, { type: 'error', persistent: true });
```
clottman: It's not obvious to me that because a notification is Persistent it should be type='error'. Do we want to allow for passing in the type?

taravancil: That's just for this specific notification, the one that shows when you lose connectivity

src/state/notifications.js
taravancil: Yep! Thanks for the catch

src/state/offline-notice.js
clottman: Ahh - I was loking at the name of this component, which is PersistentNotification so I was thinking it was more general. I see now that it's not exported. Do we need a separate component or can we combine this component with the OfflineNotice and reduce some confusion? (Or rename this to OfflinePersistentNotification or something if we want to keep them separate?)

taravancil: If that's something we want to do let's do that in a separate PR. This just ports over what was already there, and I want to try to keep this PR focused on updating the Notifications provider

PR #604, opened 2019-06-11
Long museum: collection avatar a11y improvement

sarahzinger:
## Links
* https://long-museum.glitch.me
* https://glitch.manuscript.com/f/cases/3328691/Link-text-for-collection-avatar-should-be-collection-name-not-url-slug

## Changes:
* Before when you navigated to the homepage and used a screenreader on featured collections it would read out the collection avatar as the href to the collection, which is hard to hear. I've added some text that is only accessible with a screenreader so that when you tab through it reads out the name of the collection

## How To Test:
* I used chrome and voiceover and it seemed to work ok! Feel free to do your own better testing tho!

## Feedback I'm looking for:
- anything I might be missing?

### DIFF
--- a/src/components/collection/defaultAvatar.js
+++ b/src/components/collection/defaultAvatar.js
@@ -1,8 +1,8 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
-const DefaultAvatar = ({ backgroundFillColor }) => (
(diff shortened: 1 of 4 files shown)


sarahzinger:
talked it over aria-labels vs hidden text and we decided to go with aria-labels instead! tested, and still works

PR #606, opened 2019-06-12
mighty-queen: Update packages in package.json to newest versions

keithk:
elizabethi

## Links
package.json: https://glitch.com/edit/#!/mighty-queen?path=package.json:1:0
https://mighty-queen.glitch.me/

- [ ] ...


package.json
```diff
@@ -34,64 +34,64 @@
...
-    "cache-loader": "^2.0.1",
+    "babel-preset-modern-browsers": "^14.0.0",
+    "body-parser": "^1.19.0",
+    "c3": "^0.7.1",
+    "cache-loader": "^3.0.1",
     "cheerio": "^0.22.0",
     "classnames": "^2.2.6",
-    "clean-webpack-plugin": "^2.0.0",
-    "compression": "^1.7.3",
+    "clean-webpack-plugin": "^3.0.0",
```
clottman: This is a major version change of something that's been tough to get right in our build process - how confident are you that this won't break anything?

clottman: The only thing I notice is that the default export has been removed, so we'll need to modify our webpack.config.js where we require this plugin for it to continue working. https://github.com/johnagan/clean-webpack-plugin/releases

keithk: oh oops I did that but forgot to push up, doing it now!

keithk:
Taking a look at this tomorrow, the error is now:
 Dependency cycle detected  import/no-cycle in current-user.js

I think this is because current-user.js is including a function from api.js which then includes a function from current-user.js 🙃

keithk:
@modernserf Curious actually what you think about this -
https://glitch.com/edit/#!/mighty-queen?path=src/state/current-user.js:8:16 - this is the line that refers to api.js which then calls current-user.js
Do you see any good way to split those things out, or another way to handle this conflict? (I guess this is a new eslint rule, so we could also just ignore it?)

modernserf:
that's a good rule to follow in general, but we can add an `eslint-disable-line` for this particular case.

keithk:
merging this in first thing tmrw!

PR #607, opened 2019-06-12
Nova chance: Fix the project is still building page

GregWeil:
## Links
- https://nova-chance.glitch.me/
- https://glitch.manuscript.com/f/cases/3328664/The-the-site-is-not-ready-yet-page-doesn-t-work

## GIF/Screenshots:
image

## Changes:
- Delete the stats.json file on webpack startup so the site knows that there isn't any build output ready
- Don't wipe the storybook build when webpack/the server gets restarted
- Bumped prod webpack memory from 384mb to 512mb, because it was crashing sometimes when I made a local change. This means that non perf-boosted projects are more likely to get limited, but only if they're running in dev mode.

## How To Test:
Join the project via perf boost and run `refresh` in the console. It should show the not ready page, and storybook should stay built. Changing a js file should also keep storybook around.

## Feedback I'm looking for:
Why was stats.json excluded from cleanup?

### DIFF
--- a/server/webpack.js
+++ b/server/webpack.js
@@ -1,7 +1,7 @@
 function webpackBackgroundProcess() {
   // Launch webpack in a separate process because it blocks a bit
   const { spawn } = require('child_process');
-  const env = { ...process.env, NODE_OPTIONS: '--max-old-space-size=384' };
+  const env = { ...process.env, NODE_OPTIONS: '--max-old-space-size=512' };
   spawn('webpack', ['--watch', '--info-verbosity', 'verbose'], { env, stdio: 'inherit' });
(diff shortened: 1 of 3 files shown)


keithk:
@GregWeil It wasn't included because if stats.json doesn't exist (during a rebuild, if the production site rebuilds unexpectedly for example) it will take the site down until new assets are built, where if stats.json isn't deleted it'll still point to the old (cached) assets

GregWeil:
Thanks for that context.

Now it'll only remove files not referenced by stats.json, so the entire previous build will stick around and keep getting served until the initial build finishes.

PR #611, opened 2019-06-12
Clover bead: componentize ProjectAvatar

modernserf:
## Links
* Remix link: https://clover-bead.glitch.me

## Changes:
* componentize ProjectAvatar

## How To Test:
* visit https://clover-bead.glitch.me
* check the new project menu
* check the search autocomplete results
* check any of the popover menus that have a project search (e.g. add project to team, add project to collection)


src/components/images/avatar.js
```diff
@@ -95,3 +96,26 @@ UserAvatar.defaultProps = {
   hideTooltip: false,
   withinButton: false,
 };
+
+export const ProjectAvatar = ({ project, withinButton }) => (
```
clottman: I think for accessibility, we need to be setting alt where the project avatar is being displayed to convey info about the project (name is only available as a tooltip) and setting alt='' where the name is displayed right next to the avatar.

clottman: Also places where the project avatar is used as a button should have alt text, but if ProjectAvatar is a button within something that's already a button leading to the project page, we should make it not a button (like in search results for a project)

src/components/team-analytics/team-analytics-project-details.js
```diff
@@ -19,7 +19,7 @@ const RECENT_REMIXES_COUNT = 100;
 const ProjectDetails = ({ projectDetails }) => (
   <article className={styles.projectDetails}>
     <ProjectLink project={projectDetails}>
-      <ProjectAvatar {...projectDetails} />
+      <ProjectAvatar project={projectDetails} hasAlt />
```
modernserf: @clottman this one _should_ have the alt, right? since it's by itself in a link?

clottman: Yep exactly!

PR #613, opened 2019-06-13
gentle-fireman: announce notifications with aria-live

taravancil:
## Links
* [gentle-fireman.glitch.me](https://gentle-fireman.glitch.me)
* [Manuscript](https://glitch.fogbugz.com/f/cases/3328726/Notifications-should-be-announced-using-aria-live)

## Changes:
Use [`react-aria-live`](https://npm.im/react-aria-live) to announce notifications. This uses `useRef` to extract the notification's text content, so that we don't announce images, progress bars, and whatever types of non-text content we might put in notifications someday.

## How To Test:
tbh I don't know how to test this. @clottman do you have any pointers about how to confirm that this works as expected?


clottman:
This doesn't seem to work. If it did, it would announce things like "Added to collection polar-medleyTake me there" which is in the inner text of current when it gets set, if you just added to a collection. I don't think we want to announce the text of a button if we're not indicating to the user in some way how to find the button on the page.

clottman:
Something else I thought of from reading https://inclusive-components.design/notifications/ - we should make sure to include what kind of message it is in the liveMessage, since color won't do it for us. So "Error: Connection lost" or "Info: Added to collection" as opposed to just "Connection lost" or "Added to collection"

sarahzinger:
I see this pr has been open a while, should we close it maybe? can I help with review or pair on how these should work?

taravancil:
We definitely need to finish this one! Got lost in the shuffle of the landing pages, but I plan to give it another attempt this afternoon.

Want to pair @sarahzinger? Last I remember I was struggling with voice over and could probably use a rubber duck.

taravancil:
@clottman do you know of a website I can use to see how an aria-live="polite" message should appear in Voice Over? I'm working on this now and struggling to understand what's going wrong, and it would be helpful to see a working example

taravancil:
Hold up, I think I got it

clottman:
@taravancil here's a "playground" you can use to test various settings. Aria-live=polite is recommended, but I have found it is perhaps too polite for most situations when the user really needs to know the info https://dequeuniversity.com/library/aria/content-feedback/liveregion-playground

taravancil:
🔈 Made some solid progress here! It actually announces things now and strips out non-text nodes like `<button>`s and `<progress>` or whatever.





**What's not working perfectly**
Sometimes some notifications don't get announced, even though the message is correctly set and you can see the `aria-live="polite"` node withe the message in the DOM. This happens most consistently on the add project to collection notification. But sometimes it's announced? I'm guessing Voice Over has a heuristic for deciding when to announce what, and for some reason this message isn't making the cut. Using `aria-live="assertive"` doesn't seem to help.

@clottman this is ready for another round of review

taravancil:
Nope! We only need to do that for notifications that also include non-text content, and I think I updated all of those cases. This works:

```js
createNotification('Something cool happened')
```

I'll make a ticket to investigate why the add project to collection notifications aren't being announced reliably.

PR #615, opened 2019-06-13
Lunar front 1: add accessible label to user profile link in recent projects

modernserf:
## Links
* Remix link: https://lunar-front-1.glitch.me
* Manuscript link: https://glitch.manuscript.com/f/cases/3328727/

## GIF/Screenshots:
Screen Shot 2019-06-13 at 10 32 11 AM

## Changes:
* add visually hidden "Your profile page" text to user avatar link
* make arrow in "Your projects" hidden to screen readers

## How To Test:
* log into https://lunar-front-1.glitch.me
* check the link around the user avatar

## Feedback I'm looking for:
Is this the text I should use for this link?

### DIFF
--- a/src/components/recent-projects/index.js
+++ b/src/components/recent-projects/index.js
@@ -4,12 +4,11 @@ import Heading from 'Components/text/heading';
 import ProjectsList from 'Components/containers/projects-list';
 import Loader from 'Components/loader';
 import CoverContainer from 'Components/containers/cover-container';
-import { UserLink } from 'Components/link';
+import { UserLink, WrappingLink } from 'Components/link';
 import Button from 'Components/buttons/button';
 import Emoji from 'Components/images/emoji';
 import SignInPop from 'Components/sign-in-pop';
-import VisuallyHidden from 'Components/containers/visually-hidden';
-import { getAvatarStyle } from 'Models/user';
+import { getAvatarStyle, getLink } from 'Models/user';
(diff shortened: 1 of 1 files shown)


src/components/recent-projects/index.js
```diff
@@ -49,14 +50,15 @@ const RecentProjects = () => {
...
-        <UserLink user={currentUser}>Your Projects →</UserLink>
+        <UserLink user={currentUser}>Your Projects <span aria-hidden="true">→</span></UserLink>
       </Heading>
       {isAnonymousUser && <SignInNotice />}
       <CoverContainer type="user" item={currentUser}>
         <div className={styles.coverWrap}>
           <div className={styles.avatarWrap}>
             <UserLink user={currentUser}>
               <div className={styles.userAvatar} style={getAvatarStyle(currentUser)} />
+              <VisuallyHidden>Your profile page</VisuallyHidden>
```
clottman: In general we should prefer setting aria-label for links instead of adding link content that's visually hidden, based on this recommendation from W3: https://www.w3.org/WAI/GL/wiki/Using_aria-label_for_link_purpose

clottman: Link text is perfect though! More helpful than just 'my name'

modernserf:
@clottman per our discussion on slack, i've made this a fake link instead, since its redundant with the (accessible) Your Projects link right before it.

PR #631, opened 2019-06-17
weak-sale: componentize user options pop and create team pop

taravancil:
## Links
* [weak-sale.glitch.me](https://weak-sale.glitch.me)
* [Manuscript](https://glitch.fogbugz.com/f/cases/3328478/Build-User-Options-Pop)

## GIF/Screenshots:
<div>

</div>

## Changes:
- Move user options pop and create team pop to `/components`
- Get rid of `:global` styles in the header that were styling the user avatar
- Use Hooks for create team pop
- Get rid of a few unused popover presenters

## How To Test:
* Open the user options pop, is it keyboard-navigable? Do all the buttons and links work?
* Open the create team pop. Type a team name that's already taken, you should see an error. Create a team, it should work.


src/components/user-options-pop/index.js
```diff
@@ -2,38 +2,24 @@ import React from 'react';
...
-            Sign in
-          </button>{' '}
-          to create teams
-        </p>
-        <Button size="small" emoji="herb" disabled>
-          Create Team
-        </Button>
-      </>
-    );
-  }
```
taravancil: Removed this after a discussion with Yasi

src/components/create-team-pop/index.js
```diff
@@ -0,0 +1,158 @@
...
+    setState({
+      teamName: newValue,
+      error: '',
+    });
+
+    await debouncedValidate(newValue);
+  };
+
+  const handleSubmit = async (event) => {
+    event.preventDefault();
```
modernserf: Can we validate on submit as well? maybe even disable the button if the team name is already in use?

PR #633, opened 2019-06-17
Like story: Clarify error states on optimistic input fields

sarahzinger:
## Links
* https://like-story.glitch.me
* https://glitch.manuscript.com/f/cases/3327685/

## Changes:
* when we hit an error show a default error message to the user with the truck (this just seemed like a good idea to me, but if we think it shouldn't work that way let me know, happy to change it! for example I do wonder if the truck should be reserved for user-errors and not show for things like server errors?)
* when we hit an error and then we blur the field, we hide the error and instead show the last saved good state so it's clear to the user what's there.
* some overall renaming as I found stuff a bit confusing here and there and am attempting to make it a little easier to read, though it's possible I made it even more confusing lol, please advise if you have feelings on it. I renamed useOptimisticText for example to usePassivelyTrimmedInput.

## How To Test:
* you can play around with these fields on storybook for now as shown in the gif, or if you're feeling into it you can go through all the places we use optimistic inputs, I believe I've identified all of them here:

- [x] Optimistic-markdown-input
    - [x] auth-description
        - [x] note
        - [x] collection page ( collection description)
        - [x] project page (project description)
        - [x] team page
        - [x] user page
- [x] Optimistic-text-input
    - [x] project-domain-input
    - [x] team-fields
    - [x] user-login-input
    - [x] user-name-input
- [x] Optimistic-wrapping-text-input
    - [x] collection-name-input

To test error states I added things like this:
```
  async updateNote({ note, projectId }) {
    note = _.trim(note);
    throw new Error("test error")
    await this.props.api.patch(`collections/${this.state.id}/project/${projectId}`, { annotation: note });
    this.updateProject({ note, isAddingANewNote: true }, projectId);
  }
```
but I imagine you could also toggle to offline or do something else to force an error

To compare with what we have on prod you may wish to play around with: https://long-museum.glitch.me
I added some alternating error/not error behavior to updating notes which you can see here: https://glitch.com/edit/#!/long-museum?path=src/state/collection.js:82:4
you'll notice that it doesn't show errors that don't have the exact format we expect from our api.

## Feedback I'm looking for:
- bugs! I found this to be trickier than I originally anticipated, very possible I missed something
- code style, again it's possible I'm doing something weird here. Originally we had a custom hook which inside of it called a custom hook. Now I have component that calls 2 hooks. I had originally thought to make it a component that called 3 hooks and transformed state along the way but that proved...hard? so now it's 2 hooks. Perhaps I should go back to the nesting we had before?
- ux: does this feel right to you? I *think* this is an improvement over what we have but I wouldn't be surprised if it jogs new ideas for how this should work so it's as seamless as possible. it's also difficult for me to tell because I've been looking at it for too long
- I added a default error message, does it seem clear/appropriate/on brand?


sarahzinger:
@GregWeil I'm going to tag you in this since I know you did a lot of the initial work on this (and boy is this thing a beast!) and would love your thoughts!

src/components/fields/use-optimistic-value.js
```diff
@@ -2,37 +2,79 @@ import React from 'react';
...
+
+  // as the user types we save that as state.value, later as the user saves, we reset the state.value to undefined and instead show whatever value is passed in
+  const optimisticOnChange = (newValue) => {
+    setState((prevState) => ({ ...prevState, value: newValue, shouldShowLastSaved: false, error: null }));
+  };
+
+  let optimisticValue = value;
+  if (state.shouldShowLastSaved) {
+    optimisticValue = state.lastSaved;
+  }
```
GregWeil: What's the purpose of `shouldShowLastSaved`? It seems to store the 'committed' value of the field, which should be the same as the value passed into the hook, and used by setting `state.value` to undefined like the `onChange().then` call does.

src/components/fields/use-optimistic-value.js
```diff
@@ -2,37 +2,79 @@ import React from 'react';
...
+/*
+
+Use Optimistic Value:
+
+- takes in an initial value for the input
+- takes in a way to update the server
+
+as users type (on change):
+- we ping the server with a trimmed version of the text
+- we display an untrimmed version
```
GregWeil: This hook doesn't trim the value, so it can be used for numbers/checkboxes alongside text.

src/components/fields/use-passively-trimmed-input.js
```diff
@@ -0,0 +1,15 @@
...
+
+  const displayedInputValue = rawInput === untrimmedValue.trim() ? untrimmedValue : rawInput;
+
+  const wrapAsyncUpdateWithTrimmedValue = (value) => {
+    setUntrimmedValue(value);
+    return asyncUpdate(value.trim());
+  };
+
+  return [displayedInputValue, wrapAsyncUpdateWithTrimmedValue];
+}
```
GregWeil: Could this component do a similar onBlur callback to remove the uncommitted trailing whitespace when the input loses focus?

src/components/fields/optimistic-markdown-input.js
```diff
@@ -1,17 +1,19 @@
...
 
-const OptimisticMarkdownInput = ({ value, onChange, ...props }) => {
-  const [optimisticValue, optimisticOnChange, optimisticError] = useOptimisticText(value, onChange);
-  return <MarkdownInput {...props} value={optimisticValue} error={optimisticError} onChange={optimisticOnChange} />;
-};
+function OptimisticMarkdownInput({ value, onChange, ...props }) {
+  const [untrimmedValue, onChangeBoundWithTrimmedInputs] = usePassivelyTrimmedInput(value, onChange);
+  const [optimisticValue, optimisticOnChange, optimisticOnBlur, optimisticError] = useOptimisticValue(untrimmedValue, onChangeBoundWithTrimmedInputs);
 
+  return <MarkdownInput {...props} value={optimisticValue} onChange={optimisticOnChange} onBlur={optimisticOnBlur} error={optimisticError} />;
```
GregWeil: This will eat an onBlur passed into OptimisticMarkdownInput, which I think collection notes use.

src/components/fields/use-optimistic-value.js
sarahzinger: so I think this is what's happening:
- user hits error
- while the user is still focused on the input tag, we want to show the optimistic code even though it didn't work, incase there's something they can do about it (ex: a name is already being used)
- but when they leave their focus we want to show what has last been saved before the error (ex: user is renaming a project name from `cool-project` to `pikachu`, they see that `pikachu` has already been taken and say eh `cool-project` was fine and leave the input field, not realizing that their project has now been saved as `pikach` because that what was last successfully saved. 

so we need shouldShowLastSaved to know when to show the optimistic input (before blur) and when to show the real deal (after blur) BUT you're totally right I think in calling out that the `value` which is passed in through the hook is probs going to be the same in this instance as `state.lastSaved` so I'm going to attempt to rewrite with that knowledge thanks!!

src/components/fields/use-optimistic-value.js
sarahzinger: ah right it doesn't trim that was in `user-optimistic-text` aka `use-passively-trimmed-input` and i left this comment in here from when I was following the original flow. cool good call out about why these are separate hooks, didn't think about that!

src/components/fields/use-passively-trimmed-input.js
sarahzinger: ohhh I bet it could! I did always find that weird, let me try to update.

src/components/fields/optimistic-markdown-input.js
sarahzinger: ahh thanks!! this was exactly the kind of thing I knew I'd miss :)

src/components/fields/use-optimistic-value.js
sarahzinger: wow wait as I look at this again you are 💯 we really don't need any of this, we just need the onblur, I think I over thought this a lot more than I needed 😭 thanks again for a great code review!

sarahzinger:
@GregWeil want to take another look? I think if the user hasn't seen the error message, better to show the error message and the old text with it, once they focus and focus away it'll go back to the last saved state.

src/components/fields/use-optimistic-value.js
```diff
@@ -2,37 +2,75 @@ import React from 'react';
...
-      };
+      const setStateIfStillRelevant = (newState) => setState((prevState) => (prevState.value === debouncedValue ? newState : prevState));
+      
       // this scope can't be async/await because it's an effect
-      setValueAsync(debouncedValue).then(
-        () => setStateIfMatches({ value: undefined, error: null }),
+      onChange(debouncedValue).then(
+        () => {
+          setStateIfStillRelevant({ value: undefined, error: null });
+          return debouncedValue;
```
GregWeil: Does this return go anywhere?

src/components/fields/optimistic-wrapping-text-input.js
```diff
@@ -2,11 +2,19 @@ import React from 'react';
...
-import useOptimisticText from './use-optimistic-text';
 
-const OptimisticWrappingTextInput = ({ value, onChange, ...props }) => {
-  const [optimisticValue, optimisticOnChange, optimisticError] = useOptimisticText(value, onChange);
-  return <WrappingTextInput {...props} value={optimisticValue} error={optimisticError} onChange={optimisticOnChange} />;
+import usePassivelyTrimmedInput from './use-passively-trimmed-input';
+import useOptimisticValue from './use-optimistic-value';
+
+const OptimisticWrappingTextInput = ({ value, onChange, onBlur, ...props }) => {
+  const [untrimmedValue, onChangeWithTrimmedInputs, onBlurWithTrimmedInputs] = usePassivelyTrimmedInput(value, onChange);
```
GregWeil: This one still needs onBlur passed in

src/components/fields/use-optimistic-value.js
sarahzinger: nope :) thanks!

sarahzinger:
@GregWeil cool thanks! typos fixed I believe!

PR #635, opened 2019-06-18
Quixotic rocket: Deleted a ton of old css that isn't used anywhere

GregWeil:
## Links

## GIF/Screenshots:
🤷‍♂ it looks the same

## Changes:
- styles.css is about 25% smaller

## How To Test:
- Skim through the classes I removed, either do a find all or confirm that the relevant page still looks fine


src/components/report-abuse-pop/index.js
```diff
@@ -69,7 +69,7 @@ const Failure = ({ value }) => (
       </InfoDescription>
     </PopoverInfo>
     <PopoverActions>
-      <textarea className={`content-editable ${styles.textArea}`} value={value} readOnly />
```
GregWeil: Also a real refactor, since this was the last content-editable and it isn't even editable anyway.

styles/buttons.styl
```diff
@@ -37,18 +37,6 @@ button
```
taravancil: I think soon we'll be able to get rid of `buttons.styl` entirely!

PR #636, opened 2019-06-18
first-adapter: add ids to headings in READMEs so users can write fragment links

taravancil:
## Links
* [first-adapter.glitch.me](https://first-adapter.glitch.me)

## GIF/Screenshots:
No visible changes

## Changes:
* Add [`markdown-it-github-headings`](https://github.com/Flet/markdown-it-github-headings) package
* Add a hidden anchor within README headings so users can link to headings (some of our users are already doing this, expecting our READMEs to work like GitHub's)
* Add a `linkifyHeadings` prop to the `<Markdown>` component

## How To Test:
* Click links in the table of contents on [~tmcw-big-demo](https://first-adapter.glitch.me/~tmcw-big-demo)

## Feedback I'm looking for:
* How concerned should be we about `id` collision? I didn't add any prefixes to the`id`s, so a user can add a heading that collides with an existing `id`


src/components/text/markdown.js
```diff
@@ -17,6 +18,10 @@ const md = ({ allowImages }) => {
   if (!allowImages) {
     mdIt.disable('image');
   }
+  if (linkifyHeadings) {
+    const headingOpts = { enableHeadingLinkIcons: false, prefixHeadingIds: false };
```
sarahzinger: yeah I think it might make sense to add a prefix heading because I know for example there are accessibility issues if  a page has duplicate ids. Fortunately I think we have a nice hook to give us unique ids: https://github.com/FogCreek/Glitch-Community/blob/8b421f0ef5843ade7707fb7fea099fc173bea43c/src/components/inputs/text-input.js#L6

sarahzinger:
one random thought that comes to mind, is it possible to auto unfurl the readme when the url contains a link to a header maybe? if there was that would be cool, people could share links and it would auto scroll to the right place, for example clicking: https://first-adapter.glitch.me/~tmcw-big-demo#writing-a-presentation does not bring you to that spot since the readme is stuck behind a show more button

src/components/text/markdown.js
taravancil: The plugin has an option built-in, so we won't need that hook, it'll be as simple as flipping a boolean ✨

Where this gets tricky is the user writing links. As this PR stands, you can do
```md
Check out [some heading](#some-heading)
```

But if we add prefixes, you'll have to do
```md
Check out [some heading](#123-prefix-some-heading)
```
Which is not acceptable, so we'd need to write some tooling to hook into fragment links so that `#some-heading` actually goes to `#prefixed-some-heading`

taravancil:
Yeah! We can do an automatic unfurl, but let's save it for a separate PR.

src/components/text/markdown.js
taravancil: My thinking is this: we should decide if we're ok with the risk of a user potentially mucking up their own project page. If not, I think we'll probably want to move this to a WIP until it's officially on the roadmap—it might take more time than I originally planned.

taravancil:
Ok! Made some changes to prevent duplicate `id`s:

We're now using the `useUniqueId()` hook for DOM nodes that have `id`s:
  - This includes updating the `TooltipContainer` component to generate unique `id`s for each tooltip instead of expecting the `id` to be passed in via props

src/components/deleted-projects/index.js
```diff
@@ -27,7 +27,6 @@ const DeletedProject = ({ id, domain, onClick }) => {
         <div className={styles.projectName}>{domain}</div>
         <TooltipContainer
           type="action"
-          id="undelete-project"
```
sarahzinger: I clicked to undelete a project and the site broke, I think you'll want to keep this in or refactor it to work the way you have it

sarahzinger:
I see this pr has been open a while, do we still want to work on this? can I help? or should we close?

taravancil:
Ah, sorry. It's done IIRC or in-progress, just got pushed to the side during landing pages work. Let me finish it up by the end of this week.

taravancil:
I need to think about this some more. Now that Cassey has added skip links, we have a number of components that she worries will look tacky if we use `useUniqueId`, like glitch.com#123abc, so I need to rethink how we're going to avoid id collisions.

taravancil:
Conversation with Cassey:

> tara 1:17 PM
> there may be other options, but i see two right now:
> 1. don't worry about readme links colliding with our ids
> 2. rewrite readme urls

> Cassey 1:19 PM
> I think 1 is probably a better idea, after all we don’t set ids on anchors very often anyways

This is an imperfect solution, but I think I agree with her. The other option is to rewrite README URLs, which is more work than this deserves right now. I'll keep using `useUniqueId` for `id`s where I can, but I argue we accept the risk of id collision with elements like `<main id="main">`

src/components/deleted-projects/index.js
taravancil: I can't reproduce this anymore, can you?

sarahzinger: not anymore! haha, nice :)

src/presenters/pages/router.js
```diff
@@ -58,11 +58,24 @@ const PageChangeHandler = withRouter(({ location }) => {
...
+  let linkedEl = null;
+  useEffect(() => {
+    if (!linkedEl && location.hash) {
+      linkedEl = document.getElementById(location.hash.substr(1));
+      if (linkedEl && !scrolledToLinkedEl) {
+        linkedEl.scrollIntoView();
+        setScrolledToLinkedEl(true);
+      }
+    }
+  });
```
taravancil: This is to handle when you first visit the page or refresh the page. Click [this link](https://first-adapter.glitch.me/~tmcw-big-demo#features) and it should scroll down to the "Features" heading.

src/components/new-stuff/index.js
```diff
@@ -52,7 +55,7 @@ const NewStuffOverlay = ({ setShowNewStuff, showNewStuff, newStuff, closePopover
         <div className={styles.newStuffAvatar}>
           <NewStuffPup />
         </div>
-        <OverlayTitle id="newStuff">New Stuff</OverlayTitle>
+        <OverlayTitle id={useUniqueId()}>New Stuff</OverlayTitle>
```
sarahzinger: is it ok that on every render it'll make a new id? is that desirable?

taravancil: Probably not. I'm going to revert it to newStuff in case we want to be able to link to it.

taravancil:
Ah yeah, looks like he wrote the link wrong. It should be #quickstart-installation-glitch. Cool, thanks for all the reviews!

PR #637, opened 2019-06-18
Encouraging girdle: fix bug in search bar results

modernserf:
## Links
* Remix link: https://encouraging-girdle.glitch.me/

## Changes:
* add an `openPopover` method to PopoverContainer passed props
* use openPopover, not togglePopover, on search input focus

## How To Test:
* verify that search bar results work with both keyboard and click navigation
* verify that the "how it works" video works with both keyboard and click navigation

### DIFF
--- a/src/components/search-form/index.js
+++ b/src/components/search-form/index.js
@@ -127,8 +127,9 @@ const AlgoliaSearchController = withRouter(({ history, visible, openPopover, def
     if (!query) return;
     if (selectedResult) {
       history.push(urlForItem[selectedResult.type](selectedResult, query));
+    } else {
+      history.push(`/search?q=${query}`);
     }
-    history.push(`/search?q=${query}`);
   };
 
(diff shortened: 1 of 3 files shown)


src/components/search-form/index.js
```diff
@@ -100,20 +95,15 @@ const { actions, reducer } = createSlice({
...
-function AlgoliaSearchController({ visible, togglePopover, children, defaultValue }) {
+const AlgoliaSearchController = withRouter(({ history, visible, openPopover, defaultValue }) => {
   const initialState = {
     selectedResult: null,
     query: defaultValue,
-    redirect: null,
     results: [],
   };
-  const [{ query, results, selectedResult, redirect }, dispatch] = useReducer(reducer, initialState);
+  const [{ query, results, selectedResult }, dispatch] = useReducer(reducer, initialState);
```
clottman: Destructuring the results of useReducer like this is such a great idea.

PR #644, opened 2019-06-18
chalk-laugh: refactor Text component

taravancil:
## Links
* [chalk-laugh.glitch.me](https://chalk-laugh.glitch.me)
* [Manuscript](https://glitch.fogbugz.com/f/cases/3328263/Text-component-margins-and-font-size-adjustments)

<div style="display: flex">
  
  

</div>

## Changes:
* Add `size`, `weight`, `defaultMargin` props to Text component
* Allow Text component to accept `className`
* Update all instances of Text component to take advantage of these updates

## How To Test:
Try out a bunch of places where the Text component is used:
- Project count on [collection page](https://chalk-laugh.glitch.me/@taravancil/glitch)
- [Error page](https://chalk-laugh.glitch.me/404)
- [Home page "of course this site was made on glitch too"](https://chalk-laugh.glitch.me)

## Feedback I'm looking for:
- Instead of the `defaultMargin` prop, would we prefer `marginTop` and `marginBottom` props? This would eliminate some cases where we're using a `className` just to specify a margin
- As far as I can tell, there's nowhere the `.p + .p` selector is applied. Do we want to eliminate that and just use margin props? ^

### DIFF
--- a/src/components/text/text.styl
+++ b/src/components/text/text.styl
@@ -5,7 +5,6 @@
   font-size: var(--size);
   font-weight: var(--weight);
   line-height: 1.35
-  line-height: 20px
   margin: 0
   + .p
     margin-top: 1em
--- a/src/components/text/text.styl
+++ b/src/components/text/text.styl
@@ -3,3 +3,7 @@
 .p
   font-family: sansserif
   line-height: 20px
+  margin: 0
+
+.defaultMargin
+  margin: initial
--- a/src/components/text/text.styl
+++ b/src/components/text/text.styl
@@ -4,6 +4,7 @@
   font-family: sansserif
   font-size: var(--size);
   font-weight: var(--weight);
+  line-height: 1.35
   line-height: 20px
   margin: 0
   + .p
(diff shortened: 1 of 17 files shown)


src/components/text/text.styl
```diff
@@ -2,4 +2,11 @@
 
 .p
   font-family: sansserif
-  line-height: 20px
\ No newline at end of file
+  font-size: var(--size);
+  font-weight: var(--weight);
+  line-height: 20px
```
modernserf: should line-height be proportional to --size?

src/components/text/text.styl
```diff
@@ -2,4 +2,11 @@
...
+  margin: 0
+  + .p
+    margin-bottom: 1em
```
modernserf: i think you want `margin-top` here, since it applies to the _second_ `p`

taravancil: Whoops, good catch

src/components/text/text.styl
taravancil: Ah, yep. Will update later after this workshop

PR #660, opened 2019-06-21
Butter sunspot: Replace system prompts with happy popovers

whimsicallyson:
## Links
* https://butter-sunspot.glitch.me
* https://glitch.manuscript.com/f/cases/3328669/Delete-collection-alert-dialog-doesn-t-match-delete-team-pop

## Changes:
* On a collection page, clicking "delete collection" will bring up a happy dumpster popover instead of a system prompt
* On a collection icon (ie the list on your profile page), clicking the arrow and choosing "delete collection" will also bring up a happy dumpster
* ~~(in progress, kinda stuck) On a project icon, clicking the arrow and choosing "leave project" ought to bring up a popover explaining that you won't be able to get back in without an invite. Currently it does nothing :(~~ change of plans, I'm going to wait till Justin's work in `abstracted-droplet` is good and then work from that, I think it will be much simpler.

## How To Test:
* Delete collections
* ~~Remove yourself from a project that isn't part of a team (since you can re-add yourself to those, there's no confirmation needed) and where you aren't an admin (can't remove from those)~~

## Feedback I'm looking for:
* ~~I feel like I've got things set up properly for the "leave project" popover to work, but it doesn't. What am I missing?~~

Note: opening a WIP pull request is not a good way to get eyes on a project. I'll just post in Slack in the future :)

- Get an OK
- Wait till Monday :)


GregWeil:
Deleting a collection through the chained pop looks odd, I think it should use a multipopover like removing someone from a team does
image

src/components/collection/delete-collection-pop.js
```diff
@@ -0,0 +1,67 @@
...
+  const api = useAPI();
+  const { createNotification } = useNotifications();
+  const [collectionIsDeleting, setCollectionIsDeleting] = useState(false);
+  const illustration = 'https://cdn.glitch.com/c53fd895-ee00-4295-b111-7e024967a033%2Fdelete-team.svg?1531267699621';
+
+  async function deleteThisCollection() {
+    if (collectionIsDeleting) return;
+    setCollectionIsDeleting(true);
+    try {
+      if (window.location.pathname !== getOwnerLink(collection)) {
```
GregWeil: This should check if you're on the collection page, rather than if you're not on the owner's page.

whimsicallyson: agreed. do we have a reference of how to check things like that? I used this pattern because it was one I was aware of :)

whimsicallyson:
@GregWeil since there's only one action in the pop, I removed the extra button altogether. Think that works?

src/components/collection/delete-collection-pop.js
```diff
@@ -0,0 +1,65 @@
...
+  const api = useAPI();
+  const { createNotification } = useNotifications();
+  const [collectionIsDeleting, setCollectionIsDeleting] = useState(false);
+  const illustration = 'https://cdn.glitch.com/c53fd895-ee00-4295-b111-7e024967a033%2Fdelete-team.svg?1531267699621';
+
+  async function deleteThisCollection() {
+    if (collectionIsDeleting) return;
+    setCollectionIsDeleting(true);
+    try {
+      if (window.location.pathname === getLink(collection)) {
```
GregWeil: This can get `location` as a prop since the component is wrapped in `withRouter`, rather than pulling it off of the window

src/components/collection/delete-collection-pop.js
```diff
@@ -0,0 +1,65 @@
...
+
+import { getOwnerLink, getLink } from 'Models/collection';
+import Image from 'Components/images/image';
+import Loader from 'Components/loader';
+import { PopoverDialog, PopoverActions, PopoverTitle, ActionDescription, PopoverMenuButton } from 'Components/popover';
+import { deleteCollection as deleteCollectionViaState } from 'State/collection';
+import { useNotifications } from 'State/notifications';
+import { useAPI } from 'State/api';
+
+const DeleteCollectionPop = withRouter(({ history, collection, animateAndDeleteCollection }) => {
```
GregWeil: It also needs location as a prop here so the component can pull it from react router

PR #662, opened 2019-06-21
Melted scorpion: fix pagination bug

sarahzinger:
## Links
* Remix link: https://melted-scorpion.glitch.me/
* Manuscript link: https://glitch.manuscript.com/f/cases/3328769/pagination-bug

## GIF/Screenshots:
This gif is tinier than I thought but if you click it should be more understandable

## Changes:
* resets the current page to 1 if the number of projects has changed

## How To Test:
* navigate to a user (or team) page or anywhere else we might show paginated and filtered projects
* click forward through a few pages
* find something to filter by that has fewer than that amount of pages
* notice that the the pagination starts at the beginning again (rather than keeping you on a non existent page)

## Feedback I'm looking for:
- I believe I'm using useEffect/useReducer correctly but feel free to tell me if I'm missing something! 
- let me know if you see other bugs this causes! I couldn't find any but it's def possible they are there!


src/presenters/pages/user.js
```diff
@@ -226,8 +226,6 @@ UserPage.propTypes = {
     projects: PropTypes.array.isRequired,
     teams: PropTypes.array.isRequired,
     collections: PropTypes.array.isRequired,
-    _cacheCover: PropTypes.number.isRequired,
```
sarahzinger: this is unrelated just noticed we were getting console.log errors about this, looks like we now fetch this data with a hook after we load the component so these props don't seem to be every passed in as far as I can tell

PR #669, opened 2019-06-25
Defiant-Okra: Update package.json and fix conflicts

keithk:
Easy-Baked-Okra-5-500x500

## Links
* Remix link: https://defiant-okra.glitch.me


src/state/current-user.js
```diff
@@ -5,7 +5,7 @@ import { getSingleItem, getAllPages, allByKeys } from 'Shared/api';
 import { sortProjectsByLastAccess } from 'Models/project';
 import { configureScope, captureException, captureMessage, addBreadcrumb } from 'Utils/sentry';
 import useLocalStorage from './local-storage';
-import { getAPIForToken } from './api';
+import { getAPIForToken } from './api'; // eslint-disable-line
```
GregWeil: Could this only disable the specific rule being broken, rather than all of eslint?

keithk: oops yes, I did that and just didn't push it one sec

taravancil:
Cool, thanks @keithk, going to deploy it!

PR #674, opened 2019-06-26
Gaudy report: hookify current user

modernserf:
## Links
* Remix link: https://gaudy-report.glitch.me/

## Changes:
* hookify CurrentUser container
* also, hookify Popover container -- now there are no more components to hookify 

## How To Test:
* open two tabs with https://gaudy-report.glitch.me/
* in one tab, login with email + code
* your account should load in both windows
* logout in one tab
* your account should unload in both windows


src/state/current-user.js
```diff
@@ -79,227 +79,166 @@ function identifyUser(user) {
...
 
     // Check if we have to clear the cached user
-    if (!usersMatch(sharedUser, this.props.cachedUser)) {
-      this.props.setCachedUser(undefined);
+    if (!usersMatch(sharedOrAnonUser, cachedUser)) {
+      setCachedUser(undefined);
     }
 
-    const newCachedUser = await this.getCachedUser();
+    const newCachedUser = await getCachedUser(api, sharedUser);
```
GregWeil: I think this should be using sharedOrAnonUser

src/state/current-user.js
```diff
@@ -79,227 +79,166 @@ function identifyUser(user) {
...
+
     if (newCachedUser === 'error') {
       // Looks like our sharedUser is bad, make sure it wasn't changed since we read it
       // Anon users get their token and id deleted when they're merged into a user on sign in
-      // If it did change then quit out and let componentDidUpdate sort it out
-      if (usersMatch(sharedUser, this.props.sharedUser)) {
+      // If it did change then quit out and let useEffect sort it out
+      if (usersMatch(sharedOrAnonUser, sharedUser)) {
```
GregWeil: sharedUser will get bound when this function is defined/run, rather than matching the current value from localstorage. It probably needs to be in a ref or something so the function can get the latest value.

src/state/current-user.js
```diff
@@ -79,227 +79,166 @@ function identifyUser(user) {
...
-        // 401 means our token is bad, 404 means the user doesn't exist
-        return 'error';
-      }
-      throw error;
-    }
-  }
+const useDebouncedAsync = (fn) => {
+  const [working, setWorking] = useState(false); // Used to prevent simultaneous loading
+  return async (...args) => {
+    if (working) return;
```
GregWeil: This should also use a ref, since the function is called in a setTimeout and it's possible for working to change in that time.

GregWeil:
Hmm, now if I open two signed-in tabs and sign out of one, they both get left in a broken state where they never generate a new anon user. It seems like getCachedUser is called and returns 'error' even though the net requests worked.

src/state/current-user.js
```diff
@@ -79,227 +79,176 @@ function identifyUser(user) {
...
-          level: 'info',
-          message: `Fixed shared cachedUser. Was ${JSON.stringify(sharedUser)}`,
-        });
-        addBreadcrumb({
-          level: 'info',
-          message: `New shared cachedUser: ${JSON.stringify(newSharedUser)}`,
-        });
-        captureMessage('Invalid cachedUser');
+        setFetched(false);
+        const newSharedUser = await getSharedUser(api, persistentToken);
```
GregWeil: This should grab an updated persistentToken from sharedOrAnonUser in case it just generated a new anon user

src/state/current-user.js
```diff
@@ -79,227 +79,176 @@ function identifyUser(user) {
...
-    let { sharedUser } = this.props;
+  const load = useDebouncedAsync(async () => {
+    let sharedOrAnonUser = sharedUser;
 
     // If we're signed out create a new anon user
-    if (!sharedUser) {
-      sharedUser = await this.getAnonUser();
-      this.props.setSharedUser(sharedUser);
+    if (!sharedOrAnonUser) {
+      sharedOrAnonUser = await getAnonUser(api);
```
GregWeil: Finally figured out why making new anon users wasn't working! After getting a new anon user the persistent token has changed, so the api needs to be recreated or else the getCachedUser will fail because the auth header wasn't be sent.

PR #675, opened 2019-06-26
Vivacious shirt: Fix /vscode-auth multi-prompt

dluxemburg:
## Links
* Remix link: https://vivacious-shirt.glitch.me/

## Changes:
* Before bouncing users to VS Code to complete the authentication process, navigate to `/` without a page refresh . This prevents the confirmation prompt from being re-triggered when the browser deals with going to a custom URI (`vscode://`) that is handled by the system by redirecting back.
* Also update some copy to reflect that signing in with email is no longer its own command in the extension.

## How To Test:
* Set a `persistentToken`
* Go to https://vivacious-shirt.glitch.me/vscode-auth
* Observe that dismissing whatever prompt appears does not result in another prompt

## Feedback I'm looking for:

Things I did wrong? (This is my first pull request on this project and the last time I wrote React code was 2017.)


taravancil:
Stylistically the code looks good!

Is this PR intended to address this multi-popup situation? If so, I'm still seeing it:

Screen Shot 2019-06-26 at 3 37 35 PM

taravancil:
And as discussed in-person, also seeing this error in dev tools:

```
Warning: Failed prop type: Invalid prop `insiders` of type `string` supplied to `withRouter()`, expected `boolean`
```

oliviacpu:
:eyes: Checking this out now.

dluxemburg:
@taravancil a less severe (only one extra dialog) version of the bug in Chrome is successfully fixed by this, so it looks like the issue might manifest differently across browsers and I'm just addressing the Chrome-specific symptoms here.

oliviacpu:
💁 I tried out a change with `useEffect` on `luxurious-whimsy` (my best remix name yet)

https://luxurious-whimsey.glitch.me/vscode-auth

How does that work for you both? @taravancil @dluxemburg

dluxemburg:
@oliviacpu yes and that is really cool!

PR #681, opened 2019-06-27
Gigantic chair: handle and log issue with determineProjectOptionsFunctions 

sarahzinger:
## Links
* https://gigantic-chair.glitch.me
* https://sentry.io/organizations/fog-creek-software/issues/1076258278/?project=1246508&query=is%3Aunresolved

## Changes:
* for reasons that are not currently clear to me, we often get a project object that seems to be missing a permissions field. I would like to log what project is at this point in time to better understand what could be going wrong and pass an empty array for permissions so as to not break the function overall

## Feedback I'm looking for:
- if you have ideas for why we might be missing a permissions field that I can't think of at the moment that would be cool! Then we don't even need this pr lol. 
- or if you see reasons why it would be bad to continue rendering with an empty permissions field that would be good too!


modernserf:
I think this would be handled in state/project-options now.

sarahzinger:
hmm I think the bug has gone away since we moved to state/project-options :) going to close this.

PR #654, opened 2019-06-20
[WIP] Mousy termite: centralize logic for displaying project popover actions

modernserf:
## Links
* Remix link: https://mousy-termite.glitch.me
* Manuscript link: https://glitch.manuscript.com/f/cases/3328765/Make-project-options-consistent-and-available-everywhere

## Changes:
* pass all possible project options as props, instead of going section by section
* pass profile object (e.g. user, team) instead of `isAuthorized` as project options
* provide default project options for addProjectToCollection, leaveProject, deleteProject that are accessible from everywhere

## How To Test:
* visit https://mousy-termite.glitch.me
* verify project options are not available when logged out
* log in
* verify that pin / unpin / feature options are correct for your pinned vs recent projects
* verify that join / leave options are correct for your team projects
* verify that you cannot pin / unpin / feature / join / delete other users' or other teams' projects

PR #683, opened 2019-06-27
Abstracted droplet: normalize project options

modernserf:
replaces #654.

## Links
* Remix link: https://abstracted-droplet.glitch.me/
* Manuscript link: https://glitch.manuscript.com/f/cases/3328765/Make-project-options-consistent-and-available-everywhere

## Changes:
* pass all possible project options as props, instead of going section by section
* pass profile object (e.g. user, team) instead of `isAuthorized` as project options
* provide default project options for addProjectToCollection, joinTeamProject, leaveProject, deleteProject that are accessible from everywhere

## How To Test:
* visit https://abstracted-droplet.glitch.me/
* verify project options are not available when logged out
* log in
* verify that pin / unpin / feature options are correct for your pinned vs recent projects
* verify that join / leave options are correct for your team projects
* verify that you cannot pin / unpin / feature / join / delete other users' or other teams' projects


src/state/collection.js
```diff
@@ -51,38 +49,43 @@ function loadCollectionProjects(api, collections, setResponses, withCacheBust) {
   });
 }
 
+const CollectionProjectContext = createContext();
```
modernserf: I encountered some gnarly performance issues while i was working on this that were being caused by this and the project member context forcing frequent re-renders. Splitting these into separate contexts resolved this issue.

modernserf:
@whimsicallyson fixed the join/leave bugs. Can't reproduce the project options visibility bug

PR #684, opened 2019-06-28
Pond powder: fix notification bug

sarahzinger:
## Links
* https://pond-powder.glitch.me
* https://glitch.manuscript.com/f/cases/3328730/

## Changes:
* fix notification id's so that they provide a random number. For reasons I don't totally understand this actually fixes a bug where the popover closes when the notification closes. It's weird because I can't figure out why the popover would close? But at least when I'm qa-ing it that's what I'm seeing

## How To Test:
* add a project to a collection, 
* see the notification at the top right hand side of the screen, 
* while the notification is still present,
* open the add project to collection popover again
* notice that the popover stays open when the notification auto-dismisses

## Feedback I'm looking for:
- do you see that this fixes the bug too?  
- do you understand why fixing this bug would help?

### DIFF
--- a/src/components/collection/add-collection-project-pop.js
+++ b/src/components/collection/add-collection-project-pop.js
@@ -67,10 +67,6 @@ function AddCollectionProjectPop({ collection, togglePopover, addProjectToCollec
     { origin: 'Add Project collection' },
   );
 
-  const onSubmit = (event) => {
-    console.log("event", event)
-  }
-
   /* eslint-disable no-shadow */
   const { visibleProjects, excludingExactMatch } = useMemo(
     () => {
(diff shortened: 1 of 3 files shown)


sarahzinger:
@whimsicallyson hmm maybe it's not on every page? I'm seeing it on project pages for example: http://recordit.co/A6ipqsL2JM

whimsicallyson:
@sarahzinger hm, on a project page I'm still seeing the same issue in the remix :( (tried to create a screen recording but that's not working either? ugh computers.)

sarahzinger:
@whimsicallyson maybe it's different browsers? I'm using chrome. Also for whatever it's worth, even if we can't consistently recreate the original bug, this code change I'm recommending will fix other bugs, if the use has multiple notifications they won't have a unique id otherwise.

whimsicallyson:
I'm perfectly ok with merging this, as it is definitely a typo in the original code being fixed. I'm not convinced it fixes the issue claimed, though.

sarahzinger:
@whimsicallyson you're very right, lol I just tried again and it did not fix the bug :/ I guess I must have gotten a false positive before. I'll probably still merge this, but I'll keep looking into it.

sarahzinger:
ok got to the bottom of this going to close it out and reopen with a different solution

PR #695, opened 2019-07-01
Pond powder: random bug fixes related to adding a project to a collection

sarahzinger:
## Links
* https://pond-powder.glitch.me/

## Changes:
This PR started as a way to [fix an issue](https://glitch.manuscript.com/f/cases/3328730/) where popovers were closing when notifications faded away. BUT in the time it took me to figure it out, it got fixed by another pr. And new totally unrelated bugs appeared so I fixed those.

The reason popovers were closing when notifications dismissed was because we were remounting the project-embed component (which contained a add project to collection popover) in every render because I had written code like this:
```
const ParentComponent = () => {
    const ChildComponent = () => <div>something</div>
    return (
        <div>
            <ChildComponent />
        </div>
    )
}
```

which means in every render it will create a new component. 🤦‍♀ 

When I pulled in the latest changes I see greg has already fixed this issue for us! 🎉  However I see more generally the add to collection popover has some issues. For example it is not adding projects to collections due to a typo, also our fonts got funky for some reason, so I put them back. Also somewhere along the way I found this notification bug that I also addressed. 

## How To Test:
* try adding a project to a collection, how's it working/looking?

## Feedback I'm looking for:
- anything I might be missing? Other bugs I'm causing?


src/components/containers/results-list.styl
```diff
@@ -38,6 +38,9 @@
   
 .active .resultName
   color: inherit
+
+.resultName
+  font-size: 1.2em;
```
sarahzinger: previously this got a font size of 14px from button but I saw that setting font sizes in px is maybe not best practice, this looked right to me but feel free to look again.

src/components/project/project-embed.js
```diff
@@ -23,23 +23,6 @@ const ProjectEmbed = ({ project, top, addProjectToCollection }) => {
     baseDomain: project.domain,
   });
 
-  const bottomLeft = isMember ? (
```
sarahzinger: this is actually totally fine to leave at this point. I just removed the abstraction since I think its more confusing than helpful (at one point having this as a bottomleft/right made more sense but at this point it doesn't really give us anything

src/components/containers/results-list.styl
modernserf: `em` is context-sensitive -- you would want `rem` instead -- and we're using px literally everywhere else. we should do the px -> rem transition all at once, not piecemeal

sarahzinger: ah cool thanks that makes sense to me, putting it to 14px then.

PR #697, opened 2019-07-02
Mewing owner: clicking 'Create Team' should ... create team

modernserf:
## Links
* Remix link: https://mewing-owner.glitch.me

## Changes:
* make Create Team button a submit button
* fix setState bugs in create team pop

## How To Test:
* log into https://mewing-owner.glitch.me
* click on the user menu
* click "create team"
* enter a team name
* click "create team"
* you should be redirected to the new team page

### DIFF
--- a/src/components/create-team-pop/index.js
+++ b/src/components/create-team-pop/index.js
@@ -1,6 +1,5 @@
 import React, { useState, useEffect } from 'react';
-import PropTypes from 'prop-types';
-import _ from 'lodash';
+import { kebabCase, debounce } from 'lodash';
 import { withRouter } from 'react-router-dom';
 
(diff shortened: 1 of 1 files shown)


src/components/create-team-pop/index.js
```diff
@@ -66,7 +65,7 @@ function CreateTeamPopBase(props) {
     }
   };
 
-  const debouncedValidate = _.debounce(validate, 200);
+  const debouncedValidate = debounce(validate, 200);
```
GregWeil: Not a new bug it seems, but this debounced function will get recreated every render so it will just delay calls without batching them at all. I think it needs to be in a useState or useRef so it only gets created once.

PR #700, opened 2019-07-02
Ninth berry: fix preview projects

modernserf:
## Links
* Remix link: https://ninth-berry.glitch.me/

## Changes:
* store sampled collections instead of re-sampling on every render

## How To Test:
* visit https://ninth-berry.glitch.me/
* verify that featured collection projects do not constantly re-render


sarahzinger:
looks good to me!

PR #703, opened 2019-07-02
Nova chance: Fix question reloading, hide name/description for suspended projects

GregWeil:
## Links
- https://nova-chance.glitch.me/@csglitch1
- https://glitch.manuscript.com/f/cases/3328714/Projects-that-are-suspended-still-show-pre-suspension-description-avatar-on-user-profile-pages
- https://app.zenhub.com/workspaces/community-5d07d1359dee6059a9688176/issues/fogcreek/glitch-community-work/50

## Changes:
- Add a cache param to the questions endpoint and removed the loading spinner
- Replace project item name/descriptions with 'suspended project' (todo avatars in a separate pr)

## How To Test:
Try the questions page and let it reload a few times, look at the profile linked above

## Feedback I'm looking for:
Should suspended projects not link to their project page? the link gives away their actual name


src/components/questions/questions.styl
```diff
@@ -54,73 +54,3 @@
   margin-right: 5px
   display: inline-block
   margin-bottom: 6px
-
```
taravancil: This is a nice-sized chunk of CSS to be able to get rid of!

taravancil:
> Should suspended projects not link to their project page? the link gives away their actual name

It seems reasonable to disable the "button" and remove the link in that case. But maybe we should keep it if the current user is a superuser.

taravancil:
Can we also make the buttons use the disabled styles? They still look clickable for suspended projects.

PR #704, opened 2019-07-02
Pond powder: fix `Take me there` links

sarahzinger:
## Links
* https://pond-powder.glitch.me
* https://app.zenhub.com/workspaces/community-5d07d1359dee6059a9688176/issues/fogcreek/glitch-community-work/45

## GIF/Screenshots:
fFzZI20LqE

## the bug:
to reproduce:
- go to a collection page
- add a project to a different collection
- click the `Take me there` link in green notification on the top right
- notice the url changes, but the collection never changes

a related bug I believe (this one happens intermittently):
- on a project page
- add a project to a collection
- click the `take me there` link in the green notification
- get redirected to the collection page, but notice the project you just added is not there

Why does this happen:
- the way we render our components with react router we don't force a component mount/dismount. So we currently only see this issue for links to collection pages made with <Link> as opposed to just a regular <a> tag. 
- in theory this is a good thing, we can let react figure out what to update rather than remount the same type of component
- the issue is that our data loader is currently set up to only make a new get request for data when the api changes. But our api will not change between collection page changes. 

## Changes:
* `DataLoader` now takes an optional argument of args. Args can be anything but I'd recommend it be an object. 
- When `args` updates it will refresh the data. 
- I also updated `useCollectionEditor` to ensure that when it is passed a new `initialCollection` that collection overwrites whatever is currently in state. I believe `initialCollection` should only change when we're loading a new collection.

## How To Test:
* follow the bug steps above and see if things work as expected
- try playing around with other places we use the data loader to ensure this change does not impact the way that data loads. In theory we could set all of our data loaders to work in this way by passing in their args, but I figured until there's a bug perhaps it's not necessary. I also worried about changing how we load our data all at once, could be more of a gradual refactoring.

## Feedback I'm looking for:
- do you see bugs? am I misunderstanding stuff?
- should i just implement this strategy everywhere? am I being overly cautious in thinking we should be gradual in implementing?
- would it be better to just unmount/mount our <Route>'s ?

### DIFF
--- a/src/components/collection/add-collection-project-pop.js
+++ b/src/components/collection/add-collection-project-pop.js
@@ -61,7 +61,7 @@ function AddCollectionProjectPop({ collection, togglePopover, addProjectToCollec
       togglePopover();
       // add project to page if successful & show notification
       await addProjectToCollection(project, collection);
-      createNotification(<AddProjectToCollectionMsg projectDomain={project.domain} />, { type: 'success',  });
+      createNotification(<AddProjectToCollectionMsg projectDomain={project.domain} />, { type: 'success' });
     },
     'Project Added to Collection',
     { origin: 'Add Project collection' },
--- a/src/components/collection/add-collection-project-pop.js
+++ b/src/components/collection/add-collection-project-pop.js
@@ -61,7 +61,7 @@ function AddCollectionProjectPop({ collection, togglePopover, addProjectToCollec
       togglePopover();
       // add project to page if successful & show notification
       await addProjectToCollection(project, collection);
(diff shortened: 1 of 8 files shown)


modernserf:
Instead of requiring the get function to be memoized, pass additional args to DataLoader, like 
```
<DataLoader 
  get={(api, userId) => ....} 
  args={[user.id]}>

modernserf:
oh, here's the previous discussion on DataLoader https://github.com/FogCreek/Glitch-Community/pull/511

sarahzinger:
@modernserf you predicted the future! :) nice! Thanks for the suggestion, it seems to work pretty well imo, but obviously anything data related can touch a lot of stuff, so would love another set of 👀

PR #711, opened 2019-07-08
Profuse tea: Fix a race condition in user loading

GregWeil:
## Links

## GIF/Screenshots:
image

## Changes:
- Adds a little whitespace to the email sign in pop between the title and the emoji
- If a call to a useDebouncedAsync-ed function is blocked, it'll be flagged so the function gets called again when it finishes. This fixes an occasional issue where a user isn't fully loaded after using the console to write their token to localstorage

## How To Test:
Run `localStorage.cachedUser = JSON.stringify({ persistentToken: "TOKEN" })` in the console and refresh the browser. On glitch.com it'll only partially load your user (evidenced by a missing resume coding button). This happens most consistently on pages without embeds. On profuse-tea it'll consistently load everything.

### DIFF
--- a/src/state/current-user.js
+++ b/src/state/current-user.js
@@ -159,19 +159,11 @@ const logSharedUserError = (sharedUser, newSharedUser) => {
   captureMessage('Invalid cachedUser');
 };
 
-const useValueRef = (value) => {
-  const ref = useRef(value);
-  useEffect(() => {
-    ref.current = value;
-  }, [value]);
-  return ref;
-};
-
 const sleep = (ms) => new Promise((resolve) => setTimeout(resolve, ms));
 
 const useDebouncedAsync = (fn) => {
-  const [working, setWorking] = useState(false); // Used to prevent simultaneous loading
-  const [pending, setPending] = useState(false); // Set true if fn is called while working
+  const [working, setWorking] = useState(false); // tracks if fn is currently running
+  const [pending, setPending] = useState(false); // run fn again whenever it finishes
   const debouncedFn = async () => {
     if (working) {
       setPending(true);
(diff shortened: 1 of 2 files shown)


src/state/current-user.js
```diff
@@ -159,18 +159,37 @@ const logSharedUserError = (sharedUser, newSharedUser) => {
   captureMessage('Invalid cachedUser');
 };
 
+const useValueRef = (value) => {
```
sarahzinger: random but if we're going to put this hook outside of the component can we move it closer to where it's used? Also is there a more specific name we can give it? like useSharedUserRef maybe?

src/state/current-user.js
```diff
@@ -159,18 +159,37 @@ const logSharedUserError = (sharedUser, newSharedUser) => {
...
+  return ref;
+};
+
 const sleep = (ms) => new Promise((resolve) => setTimeout(resolve, ms));
 
 const useDebouncedAsync = (fn) => {
   const [working, setWorking] = useState(false); // Used to prevent simultaneous loading
-  return async (...args) => {
-    if (working) return;
+  const [pending, setPending] = useState(false); // Set true if fn is called while working
```
sarahzinger: what's the differences between pending and working in this case?

src/state/current-user.js
GregWeil: Moved it back, I split it out to share it across a few places but ended up not needing it

src/state/current-user.js
GregWeil: Updated the comments to hopefully be more clear. pending means that it'll run the async function again as soon as it finishes, because someone requested a run while it was already working.

src/state/current-user.js
```diff
@@ -162,15 +162,26 @@ const logSharedUserError = (sharedUser, newSharedUser) => {
 const sleep = (ms) => new Promise((resolve) => setTimeout(resolve, ms));
 
 const useDebouncedAsync = (fn) => {
-  const [working, setWorking] = useState(false); // Used to prevent simultaneous loading
-  return async (...args) => {
-    if (working) return;
+  const [working, setWorking] = useState(false); // tracks if fn is currently running
+  const [pending, setPending] = useState(false); // run fn again whenever it finishes
```
sarahzinger: ahh this makes a ton of sense! it's like the thing isWorking and hasPending things to work on got it 👍

PR #714, opened 2019-07-08
Wakeful scent: fix bugs related to project deletion

sarahzinger:
## Links
* https://wakeful-scent.glitch.me/
* https://app.zenhub.com/workspaces/community-5d07d1359dee6059a9688176/issues/fogcreek/glitch-community-work/81

## Changes:
* when a user deletes a project from the collection page, reset what projects are on that page (right now there's a weird blank area that's still clickable to the deleted project, and the other projects don't slide back to replace)
* remove the ability to delete a project from the homepage, right now it doesn't work, and seems sort of odd to me that we're encouraging folks to remove their apps from our curated collections? if we want this behavior tho, we'll want to rethink how we pass around deleteItem, so that the collection adjusts what projects it has in state. 

## How To Test:
* add a project to a collection page and try to delete it from that collection page
* add a test collection with a project you own to the list of curated collections that show on the home page, see that you can no longer delete it

## Feedback I'm looking for:
- any bugs this could cause that I'm not thinking of?
- should you actually be able to delete a project from the homepage?


src/state/collection.js
```diff
@@ -56,23 +56,27 @@ export const CollectionContextProvider = ({ children }) => {
...
   const api = useAPI();
 
-  const getCollectionProjects = useCallback((collection) => {
-    if (responses[collection.id] && responses[collection.id].projects) {
-      return responses[collection.id].projects;
-    }
-    loadCollectionProjects(api, [collection], setResponses);
-    return loadingResponse;
-  }, [responses, api]);
+  const getCollectionProjects = useCallback(
```
sarahzinger: this should just be prettier

src/state/collection.js
```diff
@@ -56,23 +56,27 @@ export const CollectionContextProvider = ({ children }) => {
...
+      loadCollectionProjects(api, [collection], setResponses);
+      return loadingResponse;
+    },
+    [responses, api],
+  );
 
-  const reloadCollectionProjects = useCallback((collections) => {
-    loadCollectionProjects(api, collections, setResponses, true);
-  }, [api]);
+  const reloadCollectionProjects = useCallback(
```
sarahzinger: prettier

src/state/collection.js
```diff
@@ -169,6 +173,13 @@ export function useCollectionEditor(initialCollection) {
     }, handleError),
 
     deleteCollection: () => deleteItem({ collection }).catch(handleError),
+    deleteProject: withErrorHandler(async (project) => {
```
sarahzinger: everything in this file that has been changed is prettier except for this block btw. It ensures that when we delete a project from a collection page that we update the collection

src/state/project-options.js
```diff
@@ -17,7 +17,7 @@ const bind = (fn, ...args) => {
 const withErrorHandler = (fn, handler) => (...args) => fn(...args).catch(handler);
 
 const useDefaultProjectOptions = () => {
-  const { addProjectToCollection, joinTeamProject, removeUserFromProject, deleteProject } = useAPIHandlers();
```
sarahzinger: if we want to allow people to delete projects from the homepage then this should be deleteItem, but then we also have to do some work elsewhere to ensure that the curated collection gets reset as well so there's not a blank spot for it

taravancil: We're not showing collection projects on the new homepage at all, just the collection title and description. I can see a world where we want deleteItem to be a default project option, but like you said there's more related work so we should do that in a separate PR.

PR #716, opened 2019-07-08
Chiseled sawfish: remove old homepage

modernserf:
## Links
* Remix link: https://chiseled-sawfish.glitch.me/

## Changes
* remove old homepage components & styles
* remove old /curated files that are now handled by the CMS


keithk:
oops meant for that to just be a comment not an approval 🙇

PR #717, opened 2019-07-09
Tundra beat: VSCode auth design changes

oliviacpu:
## Links
* Remix link: https://tundra-beat.glitch.me/vscode-auth

## Changes:
* Redesign of the VSCode auth page
* Supports OAuth and Magic Code options currently

## How To Test:
* Try the OAuth and Magic Code options and make sure it opens VSCode


src/components/layout/sign-in-layout.js
```diff
@@ -0,0 +1,95 @@
...
+
+import useDevToggle from 'State/dev-toggles';
+
+import styles from './sign-in-layout.styl';
+
+const keyImageUrl = 'https://cdn.glitch.com/8ae9b195-ef39-406b-aee0-764888d15665%2Foauth-key.svg?1544466885907';
+
+const TermsAndConditions = () => (
+  <div className={styles.termsAndConditions}>
+    By signing into Glitch, you agree to our <Link to="/legal/#tos">Terms of Services</Link> and <Link to="/legal/#privacy">Privacy Statement</Link>
```
clottman: Should probably be *Terms of Service* (not Services)

clottman: We call `legal/#privacy` our Privacy Policy, not Statement

src/components/sign-in/password-login.js
```diff
@@ -0,0 +1,97 @@
...
+import Notification from 'Components/notification';
+
+import useDebouncedValue from 'Hooks/use-debounced-value';
+
+import { useAPI } from 'State/api';
+import { useCurrentUser } from 'State/current-user';
+
+import styles from './styles.styl';
+
+function useEmail() {
```
clottman: Looks like the same function is being used in get-magic-code.js - can we move it to a helper/shared file?

src/components/sign-in/use-magic-code.js
```diff
@@ -0,0 +1,81 @@
...
+        <Notification persistent type="success">
+          Success!
+        </Notification>
+      )}
+      {status === 'error' && (
+        <>
+          <Notification persistent type="error">
+            Error
+          </Notification>
+          <div>Code not found or already used. Try signing in with email.</div>
```
clottman: Does this need to be `<div>`, or can it be `<p>`?

taravancil: If `<p>`, we should use our `<Text>` component

oliviacpu:
The copy changes and extracting the function will be part of work after this to unify all the sign-in efforts. I think I fixed the redirect and used the `<Text>` component thing.

oliviacpu:
`This redirects to Glitch.com when I test it in my browser`

Setting `destinationAfterAuth` on Glitch.com local storage to the value from the remix local storage before you do the OAuth flow should "fix" this until it is deployed I think

PR #724, opened 2019-07-10
Fallacious slice: Shuffled around analytics code so it doesn't block the initial render

GregWeil:
## Links
https://app.zenhub.com/workspaces/community-5d07d1359dee6059a9688176/issues/fogcreek/glitch-community-work/119

## Changes:
- Updated the segment snippet to the latest version, and pushed the `analytics.load()` call to the bottom of the page. This ensures that the analytics script doesn't load before the client code, which forces it to wait until the analytics are fully parsed/initialized
- Moved the browser checking code after the client js, and combined it together so `window.bootstrap()` is not called if we're showing an error page or redirecting

## How To Test:
- Try loading the page and check that analytics are still going through (POST to api.segment.io)
- Use lighthouse and compare with community.glitch.me (avoid glitch.com for a more even test)


taravancil:
Nice! This doesn't change our lighthouse score much, but that's to be expected. This change is more about perceived perf, and it does feel a tiny bit snappier ⚡️ 

I'll pull this into my Wistia remix, should make a noticeable difference

PR #725, opened 2019-07-10
Neighborly pull: Fetch missing permissions for search page

sarahzinger:
## Links
* https://neighborly-pull.glitch.me

## GIF/Screenshots:
image

## Changes:
* Previously for project items on the search page we were passing the members array, but that array did not include any information regarding admin vs not admin status so our project options popover was giving incorrect options to the user. For example users were permitted to leave projects when they were the last admin on a project, (which is not something we want to encourage as we don't want orphaned projects). 
* Now if we try to render a project item with an empty permissions array, we'll try to fetch information on that project to get the permissions. 

## How To Test:
* search for a project that you are the last admin on, make sure you can't leave the project. 

## Feedback I'm looking for:
* arguably you should be able to delete the project from this page, but that's a current issue in prod, could be solved in another pr
* I'm a little worried we're making too many network calls. Perhaps the data calls for this should be at a higher level and then pulled in with something like ProjectContext ? I was mostly following the format we have with `<ProfileListLoader>` but maybe now is the time to rethink this? Feel free to tell me your thoughts. Arguably this is more of a backend issue? Perhaps I should ask our api team to change the project details we get back from algolia? I wasn't sure if that was some kind of limitation.


taravancil:
Yeah I think you're right, it's probably time to rethink this. Let's see if we can get permissions included in the Algolia projects index.

In the past when I've asked Mads to update the Algolia indices, it's been a quick and easy change. The projects index does take almost 24h to rebuild, but that's just a waiting game. I'm not sure the best way to get this on infrastructure's roadmap but maybe DM Mads? Could be a super fast fix that he can just do.

sarahzinger:
@taravancil just reached out and he says it's not possible unfortunately, the search results don't take in  data on the logged in user I guess. 

Arguably I could move this data call higher up? I feel like that's how we used to do it, but I can't remember. I know this did work at one time :(

sarahzinger:
alright after thinking more about this/talking with tara I'm going to move this data call higher up in our search results data.

PR #727, opened 2019-07-10
tundra-beat: Create useEmail hook

oliviacpu:
## Links
* Remix link: https://tundra-beat.glitch.me

## Changes:
* So far just extracting this hook

## How To Test:
* Do the email fields still validate and show errors?

## Feedback I'm looking for:
* Generally what looks like it could be reasonably shared versus what looks like it could be shared but is misleading us and actually needs to remain separate

- [ ] Review please


src/components/sign-in/password-login.js
```diff
@@ -5,23 +5,13 @@ import Button from 'Components/buttons/button';
...
 
 import { useAPI } from 'State/api';
 import { useCurrentUser } from 'State/current-user';
 
 import styles from './styles.styl';
 
-function useEmail() {
-  const [email, setEmail] = useState('');
-  const debouncedEmail = useDebouncedValue(email, 500);
-  const validationError = useMemo(() => {
```
clottman: Is useMemo still being used somewhere else in this file, or can it be removed from imports?

oliviacpu: Oh it can maybe be removed. It might have shown up in the last lint failure that got reported. I'll look!

PR #730, opened 2019-07-11
hazel-judge: Load Wistia scripts early, so landing page videos load faster

taravancil:
## Links
* [hazel-judge.glitch.me/create](https://hazel-judge.glitch.me/create)
* [#122](https://github.com/FogCreek/Glitch-Community-Work/issues/122) 

## Changes:
**Load Wistia scripts before our bundle and before analytics**
Once our app is done loading/parsing and renders `<WistiaVideo>` components, they are playable ASAP

This means that when we add a Wistia video, we'll need to specify the video's id in the route on the server like this:

```js
// routes.js
await render(res, { title, socialTitle, description, image, wistiaVideoId: '2vcr60pnx9'});
```
  
**Lazy load the embeds on /create**
Even though the Wistia scripts were being _loaded_ earlier, they weren't being _parsed_ until after the embeds and the embeds' scripts were finished. Lazy loading these embeds makes /create feel so much snappier!

**Update render function in routes.js to take an `opts` param instead of named paramaters**
This was to accommodate the growing list of parameters. Was becoming unwieldy and weird to specify each of them. 

## How To Test:
- Visit [hazel-judge.glitch.me/create](https://hazel-judge.glitch.me/create), it should feel fast! You should be able to play the Wistia video!
- Visit the [homepage](https://hazel-judge.glitch.me/), it should work and you should be able to play the Wistia video
- Click around a bunch of different page types. I updated how our render function in routes.js takes arguments, so we should confirm I didn't break anything. I already tested a ton of page types in metatags.io, so we should be good.


src/components/wistia-video/index.js
```diff
@@ -1,28 +1,11 @@
...
-    document.head.appendChild(script);
-    loadedScripts.add(src);
-  }
-}
-
-const WistiaVideo = ({ videoId }) => {
-  useEffect(() => {
-    loadScript(`//fast.wistia.com/embed/medias/${videoId}.jsonp`);
-    loadScript('//fast.wistia.com/assets/external/E-v1.js');
-  }, []);
```
GregWeil: Removing this breaks the homepage if you go to it by clicking the fish from a different page, since the server never inserts the wistia scripts.

taravancil: I updated this to check if these script tags exist (because they were included on the server), and if not, to load them in this component

server/routes.js
```diff
@@ -229,21 +234,29 @@ module.exports = function(external) {
...
   });
+  
 
   app.get('*', async (req, res) => {
     const title = 'Glitch';
     const socialTitle = 'Glitch: The friendly community where everyone builds the web';
     const description = 'Simple, powerful, free tools to create and use millions of apps.';
     const image = `${CDN_URL}/0aa2fffe-82eb-4b72-a5e9-444d4b7ce805%2Fsocial-banner.png?v=1562683795781`;
-    await render(res, { title, description, image, socialTitle });
+    await render(res, {  description, image, socialTitle });
```
GregWeil: This lost the title

src/components/wistia-video/index.js
```diff
@@ -1,21 +1,33 @@
...
+  const loadedScripts = new Set();
+  function loadScript(src) {
+    if (!loadedScripts.has(src)) {
+      const script = document.createElement('script');
+      script.src = src;
+      script.async = true;
+      document.head.appendChild(script);
+      loadedScripts.add(src);
+    }
   }
```
GregWeil: I think the set and function should live in the effect. Keeping them here looks like they're state/event handlers

PR #731, opened 2019-07-11
Cassey-community: a11y small fixes

clottman:
## Links
* cassey-community.glitch.me
* FogCreek/Glitch-Community-Work#139
* FogCreek/Glitch-Community-Work#134
* FogCreek/Glitch-Community-Work#162
* FogCreek/Glitch-Community-Work#140
* FogCreek/Glitch-Community-Work#142

## Changes:
* Make sure the More Ideas section is consistently outside of `<main>` and in an `<aside>`
* Add more descriptive alt text to the verified badge that shows next to a team/in search results
* Add a visually hidden header on the secrets page since the text is in the background image
* Make the dev toggles buttons on the secrets page into toggle buttons for assistive tech users

## How To Test:
* Go to the dev toggles page and toggle some stuff to make sure I didn't break anything there; make sure it looks good too

## Feedback I'm looking for:
Any unexpected effects of these changes?

### DIFF
--- a/src/components/buttons/button.js
+++ b/src/components/buttons/button.js
@@ -113,7 +113,7 @@ Button.propTypes = {
   imagePosition: PropTypes.oneOf(['left', 'right']),
   /** emoji name */
   emoji: PropTypes.string,
-  ariaPressed: PropTypes.bool,
+  ariaPressed: PropTypes.string,
 };
 
 Button.defaultProps = {
--- a/src/components/buttons/button.js
+++ b/src/components/buttons/button.js
@@ -15,7 +15,7 @@ export const SIZES = ['small'];
  */
 
 const Button = React.forwardRef(({
-  onClick, href, disabled, type, size, submit, matchBackground, hover, children, active, decorative, newTab, image, imagePosition, emoji,
+  onClick, href, disabled, type, size, submit, matchBackground, hover, children, active, decorative, newTab, image, imagePosition, emoji, ariaPressed,
 }, ref) => {
   const className = cx({
     btn: true,
@@ -66,7 +66,7 @@ const Button = React.forwardRef(({
   }
 
   return (
-    <button ref={ref} onClick={onClick} className={className} disabled={disabled} type={submit ? 'submit' : 'button'}>
+    <button ref={ref} onClick={onClick} className={className} disabled={disabled} type={submit ? 'submit' : 'button'} aria-pressed={ariaPressed}>
       {content}
     </button>
   );
@@ -113,6 +113,7 @@ Button.propTypes = {
   imagePosition: PropTypes.oneOf(['left', 'right']),
   /** emoji name */
   emoji: PropTypes.string,
+  ariaPressed: PropTypes.bool,
 };
 
 Button.defaultProps = {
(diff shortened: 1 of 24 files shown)


clottman:
Made some changes to this branch instead of the one I meant to; they are all undone now so the only changes here are the ones listed in the PR description.

src/components/buttons/button.js
```diff
@@ -15,7 +15,7 @@ export const SIZES = ['small'];
  */
 
 const Button = React.forwardRef(({
-  onClick, href, disabled, type, size, submit, matchBackground, hover, children, active, decorative, newTab, image, imagePosition, emoji,
+  onClick, href, disabled, type, size, submit, matchBackground, hover, children, active, decorative, newTab, image, imagePosition, emoji, ariaPressed,
```
sarahzinger: cool hadn't heard of this one before, thanks for sharing do we only have one toggle in the app?

sarahzinger: could be a good quick one to add to your a11y tips btw!

src/components/more-ideas/index.js
```diff
@@ -8,7 +8,7 @@ import categories from '../../curated/categories';
 import styles from './more-ideas.styl';
 
 const MoreIdeas = () => (
-  <section className={styles.container}>
+  <aside className={styles.container}>
```
sarahzinger: ah yeah that does make sense! Would give another +1 for a11y tips on this :) do you have a favorite cheatsheet you like to use when deciding which tags to use btw?

src/presenters/pages/secret.js
```diff
@@ -42,18 +44,19 @@ const Secret = () => {
...
   return (
-    <section className={styles.secretPage}>
+    <main className={styles.secretPage}>
       <Helmet title="Glitch - It's a secret to everybody." />
+      <VisuallyHidden as={Heading} tagName="h1">Glitch - It's a secret to everybody</VisuallyHidden>
       <ul>
         {toggleData.map(({ name, description }) => (
           <li key={name} className={isEnabled(name) ? styles.lit : ''}>
-            <Button title={description} onClick={() => toggleTheToggle(name)}>
+            <Button title={description} ariaPressed={isEnabled(name) ? 'true' : 'false'} onClick={() => toggleTheToggle(name)}>
```
sarahzinger: won't this always return a truthy string rather than a boolean? or is it smart enough to know? Maybe we should just pass isEnabled(name) directly? I'm getting a console log error here

src/components/buttons/button.js
clottman: I _think_ it's the only one, we use checkboxes in other places (like the private/not private toggle)

src/components/more-ideas/index.js
clottman: No cheatsheet unfortunately, I typically refer to the WCAG guides which are basically specifications.

src/presenters/pages/secret.js
clottman: Good catch, the error was that I defined the prop type incorrectly. Aria-pressed (sadly) takes a string that is supposed to equal either 'true' or 'false'. :( [ctrl+f 'aria-pressed' on this page to see](https://www.w3.org/TR/wai-aria-practices/examples/button/button.html)

PR #732, opened 2019-07-11
Foregoing-spell: move focus back to popover-opening-button after a popover is closed with keyboard

clottman:
## Links
* foregoing-spell.glitch.me
* https://github.com/FogCreek/Glitch-Community-Work/issues/148 (see issue for context on how this is useful for accessibility - basically, it's to prevent an unexpected change in context for users navigating with the keyboard

## GIF/Screenshots:
gif showing feature

## Changes:
* adds a prop to popover container for a ref to the button that is used to open the popover; if that prop is passed in, it's used to set focus back to said button when the popover is closed

## How To Test:
* open a popover basically anywhere on the site (both with keyboard and by clicking the button). Close it by clicking outside the popover; it should work normally. Close it by pressing the escape key; keyboard focus should return to the button you used to open the popover

## Feedback I'm looking for:
Is the ref passing around as clear as it could be?

### DIFF
--- a/src/components/popover/container.js
+++ b/src/components/popover/container.js
@@ -18,9 +18,6 @@ const usePopoverToggle = ({ startOpen, onOpen, triggerButtonRef }) => {
   };
   const closePopover = () => {
     setStatus('closed');
-    if (triggerButtonRef && triggerButtonRef.current) {
-      triggerButtonRef.current.focus();
-    }
   };
 
   const togglePopover = (event) => {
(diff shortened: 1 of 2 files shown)


clottman:
The new stuff overlay is why the toggleButtonRef isn't a required prop - focus goes back to `body` right now and that's not ideal but probably okay for now. Going back to the user menu button would be nice, but kind of breaks how we separate our component.

PR #735, opened 2019-07-12
tourmaline-jade: Add <Video> component

taravancil:
## Links
* [tourmaline-jade.glitch.me](https://tourmaline-jade.glitch.me)
* [#157: 
Only load /create videos if they're used for the current viewport width](https://github.com/FogCreek/Glitch-Community-Work/issues/157)

## GIF/Screenshots:
**Multiple video tags, but only one with a source**

## Changes:
This was motivated by improving performance /create, which uses different videos on different viewports, so we only want to load the videos we actually need on a given viewport.

- Add a `<Video>` component that mimics srcset behavior:
```jsx
<Video
  autoPlay
  controls
  track="captions.vtt"
  sources={[
    { src: 'myvideo.mp4', minWidth: 0, maxWidth: 599 }, 
    { src: 'mybigvideo.mp4', minWidth: 600 },
  ]}
/>
```
- Update non-Wistia videos on /create to use the component
- Update CDN URLs to use `CDN_URL` on /create

## How To Test:
Visit [/create](https://tourmaline-jade.glitch.me/create)
- the videos should load
- you should only see the `-small.mp4` videos on small viewports and the big ones on big viewports
- the videos should update when you resize the window


taravancil:
Forgot to add to storybook, brb doing that now

src/components/text/markdown.js
```diff
@@ -39,8 +39,13 @@ const Markdown = ({ children, length, allowImages, renderAsPlaintext }) => {
   if (renderAsPlaintext) {
     rendered = stripHtml(rendered);
     className = '';
+
+    // in many cases we use the renderAsPlaintext prop to put markdown in a paragraph
+    // <div> can't be a descendant of a <p>, so use span
+    return <span className={className}>{rendered}</span>;
   }
 
+  // use <div> here, because <p> and other markup can't be a descendant of a <span>
```
taravancil: These are pulled in from my [plastic-climb PR](https://github.com/FogCreek/Glitch-Community/pull/734), ignore

taravancil:
Added a story, but you can't test it out yet because storybook is broken. Sarah is fixing in a separate PR, and we'll just have to trust that my story works 😄

src/presenters/pages/create.js
```diff
@@ -19,9 +20,12 @@ import { useTracker } from 'State/segment-analytics';
 import { getRemixUrl } from 'Models/project';
 import { getLink as getTeamLink } from 'Models/team';
 import { emojiPattern } from 'Shared/regex';
+import { CDN_URL } from 'Utils/constants';
 
 import styles from './create.styl';
 
+const emptyCaptionTrack = `${CDN_URL}d495650d-c651-418d-a4bc-d0574c3a3816%2Fempty-caption-track.vtt?v=1562940632157`;
```
clottman: What's the reasoning behind the emptyCaptionTrack? For accessibility we need real captions on every video, and it should be easy to see if we've forgotten to include one. (This seems like it kind of hides that there's no captions?)

src/presenters/pages/create.js
```diff
@@ -409,20 +408,30 @@ function VSCode() {
...
+            },
+          ]}
           muted
           autoPlay
           loop
-          playsInline
-        >
-          <source src="https://cdn.glitch.com/50f784d9-9995-4fa4-a185-b4b1ea6e77c0%2Fvscode-small.mp4?v=1562184049096" />
-        </video>
+          track={emptyCaptionTrack}
```
modernserf: extra prop

src/presenters/pages/create.js
taravancil: Ah that was to get around the linter was telling us we needed captions even though the video had the `muted` attribute, but now that I've [disabled this linter rule](https://github.com/FogCreek/Glitch-Community/pull/735/files#diff-476706c09215654c267e50b33f7db532R23), this isn't relevant.

I think what I'll do instead is require a caption track if the video isn't muted. Does that sound ok to you @clottman?

clottman: Makes sense to me!

src/components/video/index.js
```diff
@@ -0,0 +1,72 @@
...
+
+            // if a video has words, it needs a complete caption track
+            // if the video has background music but no lyrics or words, use a .vtt file that describes the mood of the music:
+            //
+            // WEBVTT
+            //
+            // 00:01.00 --> 00:14.00
+            // (CHEERFUL POLKA MUSIC)
+          }
+        }
```
taravancil: @clottman this seem ok to you? I'm throwing an error if a video is *not muted*, and provide inline docs about what kind of caption tracks different videos need.

clottman: Yep that sounds great!

src/components/video/index.js
```diff
@@ -0,0 +1,72 @@
...
+      }
+      setVisibleVideos(videos);
+    },
+    [windowWidth],
```
clottman: muted is a prop, should go in the useEffect dependencies array, right? This also depends on `sources` indirectly via filterVideos() so I think it should go in there too.

taravancil: Hm, if we plan to eventually use this component in a way that videos change dynamically, then yes. I wrote this to fit how we currently use videos, which is that they are static so their props don't change.

clottman: I think it'd be safer to throw them in the array anyways - so if someone reuses the component it will work out of the box. Since those components don't change anyways, your component won't have much of a penalty for adding those as a dependency on the hook

src/components/video/index.js
taravancil: Actually I think to be resilient for the future I should pull out the stuff related to `muted` into a separate effect. Good call!

src/components/video/index.js
taravancil: Actually I think to be resilient for the future I should pull out the stuff related to muted into a separate effect. Good call!

src/components/video/index.js
modernserf: I feel like it would make more sense to do this checking in proptypes than at runtime?

src/components/video/index.js
```diff
@@ -0,0 +1,72 @@
...
+          {video.track && <track kind="captions" src={video.track} srcLang="en" />}
+          <source key={video.src} src={video.src} />
+        </React.Fragment>
+      ))}
+    </video>
+  );
+}
+
+Video.propTypes = {
+  sources: PropTypes.array.isRequired,
```
modernserf: something like:
```js
sources: (props, propName, componentName) => {
  for (const source of props.sources) {
    if (!source.track && !props.muted) {
      return new Error(`No caption track provided for asset ${source.src}`);
    }
  }
  return undefined;
},
```

src/components/video/index.js
modernserf: Also, for muted videos, do we want/need to have some description of the _visual_ content?

modernserf: actually -- why is the track associated with the video size at all? why would a 1080p video have different captions than a 480p video?

src/components/video/index.js
```diff
@@ -0,0 +1,72 @@
...
+  muted: PropTypes.bool,
```
modernserf: instead of checking for either `muted` / `sources[index].track`, do something like
```js
track: PropTypes.string.isRequired,
```
and
```js
<video muted={props.track === "muted"}>
  {props.track !== "muted" && <track kind="captions" src={props.track} srcLang="en" />}
</video>
```

src/components/video/index.js
taravancil: > for muted videos, do we want/need to have some description of the visual content?

I'm not sure what the best-practices are here. @clottman if you think that's appropriate, would you like me to make a ticket and put it in the backlog?

taravancil: > actually -- why is the track associated with the video size at all?

In most cases the tracks will be the same, but you never know. Let's say we do user research that suggests a user's attention is limited on mobile, so we decide to show a truncated version of the desktop video. I figured there's no guarantee the captions will be the same, so why not account for that possibility?

src/components/video/index.js
taravancil: Hm, this will work if we assume the tracks are always the same for each source, do we feel comfortable making that assumption @modernserf?

taravancil: On second thought, it's very unlikely the tracks will be different, and if they are, we can cross that bridge when we get there

PR #736, opened 2019-07-12
Foregoing-spell: also add focus management to user options pop

clottman:
## Links
* foregoing-spell.glitch.me
* https://github.com/FogCreek/Glitch-Community-Work/issues/178

## Changes:
* Missed a pop! Now when you close the user options pop by pressing escape, focus returns to the pop button instead of body.

### DIFF
--- a/src/components/popover/container.js
+++ b/src/components/popover/container.js
@@ -42,9 +42,9 @@ const usePopoverToggle = ({ startOpen, onOpen, triggerButtonRef }) => {
       if (['Escape', 'Esc'].includes(event.key)) {
         event.preventDefault();
         setStatus('closed');
-      }
-      if (triggerButtonRef && triggerButtonRef.current) {
-        triggerButtonRef.current.focus();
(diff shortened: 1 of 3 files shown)


clottman:
Also fixes https://github.com/FogCreek/Glitch-Community-Work/issues/180 which was reported after the first foregoing-spell was deployed

src/components/popover/container.js
```diff
@@ -42,9 +42,9 @@ const usePopoverToggle = ({ startOpen, onOpen, triggerButtonRef }) => {
       if (['Escape', 'Esc'].includes(event.key)) {
         event.preventDefault();
         setStatus('closed');
-      }
-      if (triggerButtonRef && triggerButtonRef.current) {
-        triggerButtonRef.current.focus();
+        if (triggerButtonRef && triggerButtonRef.current) {
```
clottman: Focus should move if we're closing the popover by pressing the escape key, not just any time a key is pressed while in a popover. 😱

keithk: 👍🏼

PR #737, opened 2019-07-12
Marred socks

dluxemburg:
## Links
* Remix link: https://marred-socks.glitch.me

## GIF/Screenshots:
(It's a little slow...)

vscodium

## Changes:
- Use the `scheme` parameter now passed by the VS Code extension when present (but check it against a list of approved ones so it can be used to redirect to any old app).
- Remove the `openProject` parameter. We're not using it.
- `insiders` is kind of also obsolete, but there may be some pre-releases floating around. I'll remove it later.

## How To Test:

While already authed and the relevant apps installed:

- https://marred-socks.glitch.me/vscode-auth?scheme=vscode should ask to open VS Code
- https://marred-socks.glitch.me/vscode-auth?scheme=vscode-insiders should ask to open VS Code Insiders
- https://marred-socks.glitch.me/vscode-auth?scheme=vscodium should ask to open VSCodium

`vscodium-insiders` [doesn't seem to exist yet](https://github.com/VSCodium/vscodium-insiders). Seems better to only add it once it does.

## Feedback I'm looking for:

- Passing a URI scheme we are going to tell the browser to open in a query param is a little scary. The check should be sufficient through, right?
- Is defaulting to VS Code on a bad scheme the right thing to do? It's what happens if there is no scheme (or insiders flag).

- [x] Test with extension flow in VSCodium


src/presenters/pages/vscode-auth.js
```diff
@@ -10,7 +10,16 @@ import SignInLayout from 'Components/layout/sign-in-layout';
...
 
-const VSCodeAuth = ({ insiders, openProject }) => {
+const KNOWN_DISTRIBUTION_SCHEMES = new Set([
+  'vscode',
+  'vscode-insiders',
+  'vscodium',
+]);
+
+const VSCodeAuth = ({ insiders, scheme }) => {
+  if (!scheme) { scheme = insiders ? 'vscode-insiders' : 'vscode'; }
```
oliviacpu: Curious about how the extension is formatting this now? If we're always sending the scheme, which should be one of the known ones in that set, do we still need to have the `insiders` property to determine that first part?

oliviacpu: Oh I just saw your note about removing it later. 💯 👍

dluxemburg: No, we should take it out. Theoretically someone could have one of the preview copies installed, but in the unlikely even this comes up we can tell them to update.

dluxemburg: Updated!

PR #712, opened 2019-07-08
Understood basin: leave project popover

whimsicallyson:
## Links
* Remix link https://understood-basin.glitch.me
* Manuscript link https://glitch.manuscript.com/f/cases/3328669/Delete-collection-alert-dialog-doesn-t-match-delete-team-pop (related)

## GIF/Screenshots:
Screen Shot 2019-07-08 at 3 57 39 PM

## Changes:
* Uses `MultiPopover` to confirm you want to leave a project, instead of a system prompt

## How To Test:
* Leave a project that's part of a team you're in, you should leave without a popover.
* Leave a project that isn't, you should get a popover warning you about the consequences of leaving and asking if you're sure

## Things left to do before deploying:
- release the new homepage :)

PR #744, opened 2019-07-15
Understood basin: separate button visibility from onClick

whimsicallyson:
## Links
Remix link https://understood-basin.glitch.me
Manuscript link https://glitch.manuscript.com/f/cases/3328669/Delete-collection-alert-dialog-doesn-t-match-delete-team-pop (related)

**Previous PR** https://github.com/FogCreek/Glitch-Community/pull/712

## Changes:
* issue with PR 712 short-circuited around popover. initial change of line 120 made popover work, but showed button even if "leave project" wasn't a legit option. Adding the visible flag gets us the button with the popover at the correct time! 

## How To Test:
* Leave projects you didn't create, and are not part of one of your teams. There should be a popover warning you about the consequences of your action.

## Feedback I'm looking for: 
Everything look good?


src/components/project/project-options-pop.js
```diff
@@ -96,10 +110,15 @@ export default function ProjectOptionsPop({ project, projectOptions }) {
...
+            leaveProject: () => <LeaveProjectPopover project={project} leaveProject={projectOptions.leaveProject} togglePopover={togglePopover} />,
           }}
         >
-          {({ addToCollection }) => (
-            <ProjectOptionsContent project={project} projectOptions={toggleBeforeAction(togglePopover)} addToCollectionPopover={addToCollection} />
+          {({ addToCollection, leaveProject }) => (
+            <ProjectOptionsContent
+              projectOptions={toggleBeforeAction(togglePopover)}
+              addToCollectionPopover={addToCollection}
+              leaveProjectPopover={leaveProject}
```
modernserf: why not `leaveProjectPopover={projectOptions.leaveProject && leaveProject}`?

whimsicallyson: because it never occurred to me? its much cleaner, though, and I like it!

PR #745, opened 2019-07-16
Marked art: assets and button for "my stuff"

sarahzinger:
## Links
* https://marked-art.glitch.me/storybook/index.html?path=/story/button--bookmarkbutton
* https://marked-art.glitch.me/storybook/index.html?path=/story/bookmarkavatar--regular
* my stuff specs: https://www.notion.so/glitch/My-Stuff-Default-Collection-f9eb3c8196c6431596fba526fc5b91c2#cf4e25097f004ceeb6a324f1355140cf
* original animation designs: https://equable-sword.glitch.me/rainbow.html
* zenhub: https://app.zenhub.com/workspaces/community-5d07d1359dee6059a9688176/issues/fogcreek/glitch-community-work/164 and https://app.zenhub.com/workspaces/community-5d07d1359dee6059a9688176/issues/fogcreek/glitch-community-work/168

## Changes:
* add assets for my stuff (might need more in the future but I believe this is most of what we need
* create a BookmarkButton component that has this cute little animation tiff made, might modify it once it's hitting a backend but for now I'm porting over what she had into react
* start moving over our storybook stories into their individual components (specifically Button and Image since I was touching those stories anyway)
* created a BookmarkAvatar component to be shown in places where we have a collection avatar 🖼 when it's part of the "my stuff" collection

## How To Test:
*`npm run storybook` look at the stories listed above in links, I haven't yet incorporated things into the app

## Feedback I'm looking for:
- quick styling suggestions to make it more polished (don't worry too much tho, we're still a fair bit away from releasing this feature, and tiff will get a chance to give it all a once over)
- a11y bugs
- interface/code suggestions, while I wouldn't be that surprised if the the svgs change, the general interface of a BookmarkButton and BookmarkAvatar probably won't too much, so if something seems wrong there let me know


src/components/buttons/bookmark-button.js
```diff
@@ -0,0 +1,87 @@
...
+      className={`${styles.halo} ${isBookmarked ? styles.haloAnimated : ''}`}
+      width="54px"
+      height="29px"
+      viewbox="0 0 54 29"
+      version="1.1"
+      xmlns="http://www.w3.org/2000/svg"
+      role="presentation"
+    >
+      <g id="Bookmark-v5" stroke="none" stroke-width="1" fill="none" fill-rule="evenodd">
+        <g id="Artboard" transform="translate(-57.000000, -14.000000)">
```
clottman: These will appear in multiple places on the page, right? If so these ids will end up being duplicated, which can be an accessibility issue. 

on the collection default avatar, we handle this by putting metadata that's just for our own info in the attribute `data-svg`

src/components/buttons/bookmark-button.js
```diff
@@ -0,0 +1,87 @@
...
+  const onClick = () => {
+    setIsBookmarked(!isBookmarked);
+    if (action) action();
+  };
+
+  return (
+    <button className={styles.bookmarkButton} onClick={onClick} aria-pressed={isBookmarked ? 'true' : 'false'} aria-label="Add project to My Stuff">
+      <Halo isBookmarked={isBookmarked} />
+      <Image src={isBookmarked ? FILLED_BOOKMARK : EMPTY_BOOKMARK} alt=""/>
+      <Image className={`${styles.check} ${isBookmarked ? styles.checkAnimated : ''}`} src={CHECKMARK} alt="" width="10px" height="10px"/>
```
clottman: How do the two images handle the page being zoomed in or out? It's breaking for me in storybook, but I don't know if that's because of how storybook handles zoom with their zoom buttons compared to how a normal site would

src/components/buttons/bookmark-button.js
sarahzinger: whoops yeah we don't need these, I can take em out, thought I got all the ids but apparently not, thanks!

src/components/buttons/bookmark-button.js
sarahzinger: hmm good point, I bet I can chat with tiff about just getting one svg for this.

sarahzinger: err... oh wait this is the button, I thought we were looking at the regular avatar. yeah for the button animation to work we'll need multiple svgs I think... maybe I can style them differently though so they respond better to zoom in/outs?

sarahzinger: you know it's interesting when I use the browser zoom in/out (command +) it works great, it's just storybook's +/- that messes things up for me, is that your experience as well?

clottman: Yeah that's my experience too!

sarahzinger: do you think the storybook zoom button is similar to what people might use to zoom in/out? or do you think we can trust that if it works in the browser that's good enough?

clottman: I think we can trust for now that if it works with browser zoom it's probably fine. There might be some issues with tools like ZoomText that are more magnifiers, but we don't have a way to test that for sure right now.

PR #746, opened 2019-07-16
Superb guide: Several changes to support server side rendering

GregWeil:
## Links
https://superb-guide.glitch.me
https://app.zenhub.com/workspaces/community-5d07d1359dee6059a9688176/issues/fogcreek/glitch-community-work/186

## Changes:
- Pushed ZINE_POSTS and HOME_CONTENT into a globals react context, which gets provided along with the BrowserRouter in client.js. Both of these will be replaced with server friendly versions when running in node.
- Replace project._avatarCache with project.avatarUpdatedAt, which will provide slightly better caching on the project page
- Wait until an effect to start reading from localStorage, rather than reading in the initial render
- The api queues up get requests when there is no persistentToken, so pages can make requests before localStorage is ready and get the right data back

## How To Test:
Try navigating around the site, signing in and out, and clearing localStorage

## Feedback I'm looking for:
Is there a better way to handle any of this? The initial render on the client must be identical to the initial render from the server, so I think it has to start out rendering as if localStorage doesn't exist.


taravancil:
> I think it has to start out rendering as if localStorage doesn't exist

Yep, I think you're right and your approach looks sound. Do we rely on any other browser APIs on initial render? We use `.click()`, `.play()`, `alert()`, and probably others, but I can't think of any cases where they're in the initial render.

I clicked around the site, logged in and out, and cleared localStorage. It all looks good to me!

GregWeil:
We inject some stuff into the Promise prototype in the uploader code, I'm setting up a separate pr to refactor it out so it just returns a promise and an onProgress listener, rather than messing with prototypes

PR #748, opened 2019-07-17
Only enforce https if not running locally (based on env variable that doesn't exist on glitch.com containers)

clottman:
## Links
* foregoing-spell.glitch.me
* Issue-tracker link: n/a

## Changes:
* If the env variable RUNNING_LOCALLY is set, don't enforce HTTPS. This enables local development (on your machine vs on glitch.com) without code changes that could get accidentally committed

## How To Test:
* Try going to a http version of foregoing-spell, you should get redirected to https
* Run `RUNNING_LOCALLY=true node server/server.js` in your local development environment if you have one. You should be able to go to localhost and see the site running. (Without setting the env variable, you'd get redirected to `https://localhost` and told that your connection wasn't secure, therefore you couldn't connect. 

## Feedback I'm looking for:
* Any considerations for why we shouldn't do this that I'm overlooking?
* Any possible way this will break sitemaps?

### DIFF
--- a/server/server.js
+++ b/server/server.js
@@ -2,6 +2,7 @@ const express = require('express');
 const compression = require('compression');
 const constants = require('./constants');
 const moduleAlias = require('module-alias');
+const dotenv = require('dotenv');
 
 moduleAlias.addAliases(require('../shared/aliases'));
 
@@ -9,6 +10,9 @@ const sentryHelpers = require('Shared/sentryHelpers');
 
 // https://docs.sentry.io/error-reporting/quickstart/?platform=node
 const Sentry = require('@sentry/node');
+
+// required for using .env when not on glitch; no-op when running on glitch
+dotenv.config();
 
 try {
   Sentry.init({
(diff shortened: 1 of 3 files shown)


server/server.js
```diff
@@ -10,6 +11,9 @@ const sentryHelpers = require('Shared/sentryHelpers');
 // https://docs.sentry.io/error-reporting/quickstart/?platform=node
 const Sentry = require('@sentry/node');
 
+// required for using .env when not on glitch; no-op when running on glitch
+dotenv.config();
```
clottman: This lets you run with a .env file, so you can use `npm run start` normally instead of trying to set the env variable when starting the server. I've used it in other projects where I'm alternating between glitch.com and local development and it doesn't seem to have any ill effects.

GregWeil:
Does `dotenv` not need to be added to `package.json`?

clottman:
@GregWeil it is rather inexplicably already there.

PR #750, opened 2019-07-17
Woozy shoemaker: Adding `My Stuff` to User Pages

sarahzinger:
## Links
* https://woozy-shoemaker.glitch.me/
* Specs/doc links: https://www.notion.so/glitch/My-Stuff-Default-Collection-f9eb3c8196c6431596fba526fc5b91c2

## GIF/Screenshots:
Screen Shot 2019-07-17 at 5 13 34 PM

## Changes:
* BEHIND A DEV TOGGLE: start work on creating a `MyStuffItem` that shows in the collections section on a User page. Assets and the backend API for this are in another pr yet to be deployed, so this is mostly just the scaffolding for it:
    * when we load the page we check to see if the user already has this special kind of collection
        * if they do it'll fetch all the projects within that collection
            * if the user is authorized (looking at their own page) we'll show that collection as the first collection
            * if the user is not authorized we only show the collection if there are projects in it.
        * if the user doesn't have a `My Stuff` collection yet and the user is authorized, it will show a placeholder image. When the user clicks on that placeholder image, it'll create an empty `My Stuff` Collection.
* some centralization of our randomColor picking logic (just seemed like a good clean up task since I was going to use it anyway)

## How To Test:
* go to https://woozy-shoemaker.glitch.me/ everything should continue to work as normal (you should not see any references to `My Stuff` anywhere, selecting a random color for a collection should work as normal.
* go to https://woozy-shoemaker.glitch.me/secret and enable `My Stuff` and go to your user page, you should see a `My Stuff` placeholder section in collections if you're logged in (and not if you're not logged in). 

## Feedback I'm looking for:
- what do you think? see any bugs? 

## Things left to do before deploying (or might do this in another pr):
- [x] investigate why unnecessary re-renders are happening (in the light of day I can't recreate this issue...? I tried installing why-did-you-render in a remix of this remix to see issues and saw nothing 🤷‍♀ )
- [x] qa changes I did to randomColor
- [x] teams shouldn't get a new default `my stuff`
- [ ] pull in images once those are in prod (might do this in another pr)
- [ ] add api calls once those are in prod (will probably do this in another pr)
- [ ] clean up, add new components to storybook? (will do in another pr)


src/presenters/pages/user.js
```diff
@@ -67,6 +79,47 @@ NameAndLogin.defaultProps = {
   login: '',
 };
 
+function MyStuffCollectionLoader({ collections, myStuffCollection, isAuthorized, children }) {
+  const { value: projects } = useCollectionProjects(myStuffCollection);
+
+  React.useEffect(() => {
+    if (projects && (projects.length > 0 || isAuthorized)) {
```
clottman: What does this effect do?

sarahzinger: At the moment this code doesn't run because the API hasn't been implemented yet, but in the future:

this effect adds projects to the MyStuffCollection, and adds that collection to the beginning of the collections list under certain conditions:
- if there are projects inside of it, it always shows
- if there user is looking at their own user page it always shows

It won't show a my stuff collection:
- if I'm looking at someone else's user page and their My Stuff collection is empty

although now that I'm looking at this code again I think I'm using find and not filter to get the "My Stuff Collection" out so I'll have to add that.

PR #751, opened 2019-07-18
Quixotic rocket: Fix link components using hash/relative urls

GregWeil:
## Links

## Changes:
Use withRouter to map urls relative to the current url and not just the origin

## Feedback I'm looking for:
TODO in a separate pr: make our own `useRouter()` thing so we don't have to mess around with HOCs


GregWeil:
Update: `history` and `location` are now exposed by `useGlobals()`. Potential danger in that `location` is just a URL object, so you can't use it for navigation like with `window.history` or the functions in the history object.

PR #752, opened 2019-07-18
Cassey-community: skip links

clottman:
## Links
* cassey-community.glitch.me
* https://github.com/FogCreek/Glitch-Community-Work/issues/154

## GIF/Screenshots:
medium screen size
image

full width screen size 
image

mobile size
image

#main on homepage 
image

## Changes:
* Skip links! https://axesslab.com/skip-links/ 
* On any page on the community site, the first tab stop on the page should be a button that appears when it receives focus, that when clicked, moves keyboard focus to the first header tag inside the `<main>` section. 

## How To Test:
* **If you're testing in Firefox or Safari**: enable tabbing to all controls on your mac. [how-to](https://www.scottohara.me/blog/2014/10/03/link-tabbing-firefox-osx.html)


.gitconfig
```diff
@@ -1,2 +1,4 @@
 [push]
 	default = simple
+[core]
```
sarahzinger: is this intentional?

clottman: 😠 yes but I thought it was git ignored :(

src/components/buttons/button.js
```diff
@@ -14,63 +14,91 @@ export const SIZES = ['small'];
...
+      unstyled: type === 'dropDown',
+      hasImage: emoji || image,
+      hasNarrowEmoji: ['balloon', 'index', 'policeOfficer'].includes(emoji),
+      hasSunglassesEmoji: emoji === 'sunglasses',
+      padLeft: (image || emoji) && imagePosition === 'left',
+      matchBackground: matchBackground === true,
+      active,
+      hover,
+      decorative,
+    }, className);
```
sarahzinger: I wonder if it's more inline with our styles to pass in a prop rather than a class name so the prop is like "isVisibleOnFocus: true" or something. That said I don't have super strong opinions on that, not blocking!

clottman: In this case it'd be `isVisibleOnFocusAndAbsolutelyPositionedAtTopLeft`, and I can't think of any other times we'd want to use that.

src/components/search-results/index.js
```diff
@@ -124,16 +119,12 @@ function SearchResults({ query, searchResults, activeFilter, setActiveFilter })
...
-      {!ready && (
-        <>
-          <Loader />
-          <h1>All results for {query}</h1>
-        </>
-      )}
+    <main className={styles.page} id="main">
       {ready && searchResults.totalHits > 0 && (
-        <FilterContainer filters={filters} setFilter={setActiveFilter} activeFilter={activeFilter} query={query} />
+        <FilterContainer filters={filters} setFilter={setActiveFilter} activeFilter={activeFilter} />
```
sarahzinger: I think this breaks stuff. when I search for stuff it works but if I refresh the search page with the url it shows the loader forever (ex: https://cassey-community.glitch.me/search?q=louie)

src/presenters/pages/project.js
```diff
@@ -127,13 +129,13 @@ const ProjectPage = ({ project: initialProject }) => {
   const [project, { updateDomain, updateDescription, updatePrivate, deleteProject, uploadAvatar }] = useProjectEditor(
     initialProject,
   );
-
+  useFocusFirst();
```
sarahzinger: would it make sense to do this in something like app.js so you don't have to do it in ever /pages file? Or does it need to happen here for some reason?

clottman: That's a smart idea, let me see if it works!

src/components/search-results/index.js
clottman: I think something's broken but I don't think it's from this code - will investigate!

.gitconfig
sarahzinger: yeah now the delete is committed. I think the trick is to put it as it was, then commit git ignore file with .gitconfig in there, then add any changes you want :)

src/components/search-results/index.js
clottman: Actually it's happening in prod as well

src/presenters/pages/project.js
clottman: So I tried that, and it doesn't work because we need the hook to run every time the page updates. app.js renders once but the stuff inside it changes, same with router.js. So this has to live on each page.

PR #759, opened 2019-07-23
Seasoned cushion: fix storybook

sarahzinger:
## Links
* https://seasoned-cushion.glitch.me/storybook/index.html?path=/story/profilelist--loading

## Changes:
* when we upgraded a bunch of stuff we broke storybook, this fixes it so our storybook config matches our prod config better

## How To Test:
* `npm run storybook`

### DIFF
--- a/storybook-config/.babelrc
+++ b/storybook-config/.babelrc
@@ -11,7 +11,7 @@
           "edge": "16"
         },
         "useBuiltIns": "usage",
-        "corejs": 2
+        "corejs": 3
       }
     ]
   ],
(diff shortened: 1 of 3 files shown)


taravancil:
Yeah this was my bad, I didn't even think to check storybook 😖 Reviewing now!

sarahzinger:
@taravancil no worries, it should be just be a check in our circle ci stuff really, there's no way we're all going to remember to check each time tbh.

taravancil:
>  it should be just be a check in our circle ci stuff really

Totally agree. We'll be getting rid of storybook anyway once the shared components library is ready, so eventually we won't need to remember

sarahzinger:
I'm going to close this since it turns out we're just going to remove storybook all together it sounds like

taravancil:
If it's ok with you, I'm going to pull these changes into #735 because I had added a story for the `<Video>` component, and I don't want reviewers to be confused

sarahzinger:
@taravancil sure go for it

PR #760, opened 2019-07-23
Cassey-community-2: Adds server/ to files that can fail the build if they have linting errors

clottman:
## Links
* cassey-community-2.glitch.me
* FogCreek/Glitch-Community-Work#123

## GIF/Screenshots:
n/a

## Changes:
* Adds server/ to files that use eslint-loader; don't add shared/ because weird errors happen when you do that probably aren't worth fixing since those files don't change as much. 

## How To Test:
* Builds should build normally, everything should work.


GregWeil:
The server's prestart script runs eslint against server/ shared/ and webpack.config.js . If this is doing the same thing, can that be removed?

clottman:
@GregWeil We could remove server/ from that list but not shared/ or webpack.config.js, neither of which are covered by the eslint loader right now. I tried adding in shared/ but it seems to not be finding the babel config in that directory for some reason, and I wasn't sure it was worth sinking more time into.

GregWeil:
Ok. Is it possible to go in the other direction and have ci run eslint against those folders separate from webpack?

clottman:
We could, though I'd rather wait until we have Circle CI in place rather than trying to hack it into our current test runner which we won't be using for very much longer.

clottman:
When I test in my branch with the eslint from package.json removed, I'm not seeing the errors show up that I introduced in a server file. So I think this wasn't actually working when I thought it was. :( 

Closing for now, let's revisit it once we have our new CI in place. 

CC: @oliviacpu

PR #762, opened 2019-07-24
solstice-poinsettia: Make tooltips hoverable and text within tooltips selectable

taravancil:
## Links
* [solstice-poinsettia.glitch.me](https://solstice-poinsettia.glitch.me)
* [Ticket #152](https://github.com/FogCreek/Glitch-Community-Work/issues/152)

## GIF/Screenshots:
<table>
<tr>
  <td><strong>Button w/tooltip</strong></td>
  <td></td>
</tr>
<tr>
  <td><strong>Link w/tooltip</strong></td>
  <td>
  
  </td>
</tr>
<tr>
  <td><strong>Non-clickable element w/tooltip</strong></td>
  <td>
    
  </td>
</tr>
</table>

## Changes:
This makes tooltips hoverable and the text within tooltips selectable. It puts an invisible hover target between the tooltip container and the tooltip, so when you hover outside of the visible part of the container, you are actually still moused over:
Screen Shot 2019-07-24 at 1 39 59 PM

In order to support clicking to select text on draggable elements like links, I added a `draggable` prop on our `<Link>` component. When `draggable="true"` on links that wrap avatars on tooltips, clicking to select tooltip text drags the whole link instead of selecting text.

I also fixed a few related bugs along the way:
- [Fix alignment of top-aligned tooltips](https://github.com/FogCreek/Glitch-Community/compare/solstice-poinsettia?expand=1#diff-846903f0917429fd5c7d18d1ddc2894aL26)
- [Remove `font-weight: bold` on persistent tooltips](https://github.com/FogCreek/Glitch-Community/compare/solstice-poinsettia?expand=1#diff-b5aae4c7c5e782249f3da51d7d90c042L88)
- [Disabled tooltips on a few avatars where the tooltip was redundant and unecessary](https://github.com/FogCreek/Glitch-Community/compare/solstice-poinsettia?expand=1#diff-afd691baf38d3892516b58a12626d4e3R29)
- [Get rid of unecessary use of `cx`](https://github.com/FogCreek/Glitch-Community/compare/solstice-poinsettia?expand=1#diff-846903f0917429fd5c7d18d1ddc2894aL26)
- Not related, but was logging an error in the console [`filter-controller/index.js L48`](https://github.com/FogCreek/Glitch-Community/compare/solstice-poinsettia?expand=1#diff-9149377f287eb70f34d0e65e70bf0be7R48)

## How To Test:
* Open tooltips and hover them, select their text! Does it work?
* Can you still close tooltips with `ESC`?

## Feedback I'm looking for:
- Should I find a way to make the invisible hover target the same width as the tooltip itself? Please say no, I am so over tooltips lol


src/components/tooltips/tooltip-container.js
```diff
@@ -71,26 +70,34 @@ function TooltipContainer({ type, tooltip, target, align, persistent, children,
...
   const shouldShowTooltip = tooltip && (tooltipIsActive || persistent);
 
+  function cancelClick(e) {
+    e.preventDefault();
+    e.stopPropagation();
+  }
+
   let tooltipNode = null;
+  /* eslint-disable jsx-a11y/click-events-have-key-events */
+  /* eslint-disable jsx-a11y/no-static-element-interactions */
```
taravancil: I discussed disabling these rules with @clottman, she said it's ok because the click handler prevents stuff from happening, so we don't need to expose that NotStuff to keyboard users.

src/components/tooltips/tooltip-container.js
```diff
@@ -71,26 +70,34 @@ function TooltipContainer({ type, tooltip, target, align, persistent, children,
...
+      <div onFocus={() => setTooltipIsActive(true)} onBlur={() => setTooltipIsActive(false)}>
         {extendedTarget}
       </div>
+      {align.includes('top') && (
+        <span aria-hidden="true" onClick={cancelClick} className={cx({ invisibleHoverTarget: true, top: align.includes('top'), persistent })} />
+      )}
       {tooltipNode}
+      {!align.includes('top') && (
+        <span aria-hidden="true" onClick={cancelClick} className={cx({ invisibleHoverTarget: true, top: align.includes('top'), persistent })} />
+      )}
```
taravancil: Invisible hover target goes on top if the tooltip is top-aligned, and goes on bottom if it's bottom-aligned

src/components/tooltips/tooltip.styl
```diff
@@ -14,7 +14,6 @@ tooltip-background = rgba(0,0,0,0.95)
   border: none
   box-sizing: inherit
   display: inline-block
-  pointer-events: none
```
taravancil: We need pointer events in order to select text

src/components/filter-controller/index.js
```diff
@@ -45,7 +45,7 @@ function FilterController({ matchFn, enabled, placeholder, items, children, sear
         value={filter}
       />
     ),
-    filterHeaderStyles: enabled && styles.header,
+    filterHeaderStyles: enabled ? styles.header : undefined,
     renderItems: (renderFn) => {
```
taravancil: This fixes an unrelated console error

clottman:
RE: "Should I find a way to make the invisible hover target the same width as the tooltip itself" 
I think it's close enough? I'm not noticing much of a difference testing it at 100% zoom and at bigger zoom levels

src/components/tooltips/tooltip-container.js
clottman: Perhaps we should add a comment here that this is to prevent dragging the tooltip, so we remember in 6 months when someone revisits this?

taravancil:
> I think it's close enough?

That's music to my ears 🙏🏻

PR #765, opened 2019-07-25
tundra-beat: Add .circleci/config.yml

oliviacpu:
## Links
n/a

## Changes:
* Adds CircleCI/config.yml file, which I will probably need to update at least once

## How To Test:
* Merge and let CircleCI build it

## Feedback I'm looking for:
* Is anything obviously wrong before I merge this?

It should be...
- Ensuring a directory for artifacts
- Using a CircleCI docker image (same as Editor uses)
- Updating npm
- Restoring and checking cache of packages
- Doing integration tests (using our cypress script and NPX, we still can't put cypress in our package.json)
- Storing artifacts

It is very similar to Editor except it doesn't deploy or run other tests

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,79 @@
+version: 2
+jobs:
+  build_and_test:
+    docker:
+      - image: circleci/node:10-browsers
+    working_directory: ~/FogCreek/Glitch-Community
+    environment:
+      CIRCLE_ARTIFACTS: /tmp/circleci-artifacts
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -0,0 +1,45 @@
...
+          command: npm ci
+      - save_cache:
+          key: dependency-cache-v2-{{ checksum "package.json" }}
+          paths:
+            - ~/.npm
+      - run:
+          name: eslint
+          command: npm run lint
+      - run:
+          name: integration tests
```
keithk: 😭

PR #774, opened 2019-07-30
ninth-winter: add documentation about running ~community locally

taravancil:
## Links
* [ninth-winter.glitch.me](https://ninth-winter.glitch.me)

## Changes:
* Document how to run the app locally

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -89,7 +89,7 @@ RUNNING_LOCALLY=true
 ```
 4. Start the server:
 ```bash
-node server/server.js
+npm start
 ```
 
 ### How do I add a question to the FAQ?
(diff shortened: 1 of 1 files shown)


FAQ.md
```diff
@@ -69,7 +69,28 @@ In production, our code goes through cloudfront. This is why there's a delay aft
...
+```
+npm install
+```
+3. Create a file called `.env` in the root directory, and populate it with the `RUNNING_LOCALLY` variable, which tells our build process to allow the app to run on `localhost`
+```yaml
+RUNNING_LOCALLY=true
+```
+4. Start the server:
+```bash
+node server/server.js
```
oliviacpu: I've been doing `npm start` for local development, which will watch `server` and `shared` for changes and restart the server if you're editing them. I believe the server is building the client files after they change, so it isn't watched by the `start` script.

I'm doing something like this for CI though, since we don't really need to watch for changes there.

clottman: Seconded, this should be `npm start`

taravancil:
Updated to `npm start`, thanks!

PR #780, opened 2019-07-30
Fuschia canoe: Filter out "My Stuff" from "More by" and "Included in" sections

sarahzinger:
## Links
* https://fuschia-canoe.glitch.me
* https://app.zenhub.com/workspaces/community-5d07d1359dee6059a9688176/issues/fogcreek/glitch-community-work/244

## Changes:
* removes "My Stuff" collections from the "More by" and "Included in" sections on the project and collection pages. 

## How To Test:
* include a project in a "My Stuff" collection, you should never see that collection show up on the project page under the Included In section, similarly when you go to your collection pages you should never see the "My Stuff" collection show up under the "More by" section


whimsicallyson:
When My Stuff is the only collection a project is in, can we filter out the empty heading as well?
Screen Shot 2019-07-31 at 1 47 31 PM

src/presenters/pages/project.js
```diff
@@ -39,7 +39,7 @@ function syncPageToDomain(domain) {
   history.replaceState(null, null, `/~${domain}`);
 }
 
-const filteredCollections = (collections) => collections.filter((c) => c.user || c.team);
+const filteredCollections = (collections) => collections.filter((c) => (c.user || c.team) && !c.isMyStuff);
```
whimsicallyson: I think we want to use filteredCollections on line 47 as well :)

sarahzinger: oh nice catch! I bet that's an existing bug lol

whimsicallyson: that fixed it!

PR #782, opened 2019-07-31
Coherent nylon: My Stuff bookmark icon focus styling

ttseng:
## Links
* Remix link: https://coherent-nylon.glitch.me/

## GIF/Screenshots:
pr

## Changes:
* Custom focus styling for bookmark icon button.  The focus styling should appear when you use keyboard to navigate to the button.  
Note that in order to capture the shape of the button, the highlight styling is a path within the SVG itself, so the icons (filled and empty) are now inline SVG.

## How To Test:
* go to https://coherent-nylon.glitch.me/secret and turn on `my stuff`
* go to a project page (https://coherent-nylon.glitch.me/~coherent-nylon) and try interacting with the My Stuff button via keyboard and mouse navigation

## Feedback I'm looking for:
* Does the utilization of onFocus and onBlur to override default focus behavior make sense?


src/components/buttons/bookmark-button.js
```diff
@@ -74,16 +62,75 @@ Halo.propTypes = {
...
-      setState({ isAnimating: true, isBookmarked: true });
+      setState({ isFocused: false, isAnimating: true, isBookmarked: true });
     } else {
-      setState({ isAnimating: false, isBookmarked: false });
+      setState({ isFocused: false, isAnimating: false, isBookmarked: false });
     }
     if (action) action();
   };
+  const onFocus = () => {
+    setState({ isFocused: true, isAnimating: state.isAnimating, isBookmarked: state.isBookmarked });
```
sarahzinger: doesn't matter in the slightest but something I do here to save time is to write `setState({ ...state, isFocused: true })` (no need to change just letting ya know for the future)

sarahzinger:
I wonder if the blue outline should always be there if it's focused? seems weird to me that we lose it. Alternatively if we're going to add the tooltips maybe it's ok to lose it? I see with the privacy toggle we don't have an outline but we do show a tooltip explaining the toggle state, which maybe is more useful anyway. I defer to Cassey's expertise tho!

Also I'm not sure what you did that made the halo less choppy but thank you it looks great! cc-ing @whimsicallyson to let her know I think that's fixed

src/components/buttons/bookmark-button.js
ttseng: yeah! that's definitely better

ttseng:
Updated so my stuff btn retains blue focus highlight as long as it's focused via keyboard input.  Also added tooltip.

my-stuff-btn

PR #787, opened 2019-07-31
Fallacious slice: Server rendering for project pages with new api cache

GregWeil:
## Links
https://fallacious-slice.glitch.me
https://app.zenhub.com/workspaces/community-5d07d1359dee6059a9688176/issues/fogcreek/glitch-community-work/191

## Changes:
- Introduced ApiCacheProvider and useCachedProject. useCachedProject will load the project when it is called, and return `{ status, value, error }`. It'll return a value immediately if one is cached, then update with a new value when the request completes.
- The project editor watches the argument and replaces its state whenever the argument changes
- The server renders public project pages. Private and nonexistent projects render as a loading spinner, so the client doesn't flicker a not found page before showing a private project.

## How To Test:
Refresh on project pages, public and private, and try making changes to them.

## Feedback I'm looking for:
- This caches either an arbitrary url or a full project entity with users and teams. If it gets used more than at the root level of pages it will start caching redundant data. Should it move towards a normalized data store type thing? Alternatively, is this over doing it? It could probably scale back to something a bit simpler that only applies to the initial render.
- Is there a better way to handle private/not found projects? Maybe it could rope in the hasLogin cookie, so authed users get a loading spinner, while google/link previews get a not found.


GregWeil:
Another option is to shelve this for a bit in the hope that react suspense releases soon and solves this for us, but I don't know how soon that is

GregWeil:
before:
image

after:
image

keithk:
I added @taravancil  and @modernserf to this to think through if those gains are worth the added complexity to making this work, and if we should continue on this path

taravancil:
Hm, it looks like it'll be a while longer before Suspense is released and stable, so I'm not sure if we should wait for it:

> We’ve started work on a new asynchronous server renderer that will support Suspense, but it’s a large project and will take a good chunk of 2019 to complete.

I'll review these changes today to see if the added complexity makes me change my mind!

taravancil:
Double checking that metadata for project pages still works, it all looks good!

One question is should we treat suspended projects like we do 404s and private projects? We should probably at least not use their avatar. This is a problem on production too so we don't need to fix it in this PR.

<table>
<tr>
<td><strong>404</strong></td>
<td>
Screen Shot 2019-08-08 at 11 04 47 AM
</td>
</tr>

<tr>
<td><strong>private</strong></td>
<td>
Screen Shot 2019-08-08 at 11 05 46 AM
</td>
</tr>

<tr>
<td><strong>suspended</strong></td>
<td>
Screen Shot 2019-08-08 at 11 04 00 AM
</td>
</tr>
</table>

taravancil:
We might want to wait to deploy this until Sentry is fixed

PR #788, opened 2019-08-01
charming-memory: open the app in default browser if `RUNNING_LOCALLY` is set

taravancil:
## Links
* [charming-memory.glitch.me](https://charming-memory.glitch.me)

## Changes:
* Open `http://localhost:123whatever` in the default browser when`RUNNING_LOCALLY` is set.

## How To Test:
* Pull this branch
* Run the app locally ([instructions](https://github.com/FogCreek/Glitch-Community/blob/master/FAQ.md#can-i-run-this-app-locally))
* Running `npm start` should open the app in a new tab in your default browser

## Feedback I'm looking for:
* Is this helpful or annoying?
* @oliviacpu is this going to cause problems on CircleCI?

### DIFF
--- a/server/server.js
+++ b/server/server.js
@@ -94,5 +94,11 @@ app.use(Sentry.Handlers.errorHandler());
 
 // Listen on App port
 const listener = app.listen(process.env.PORT, () => {
-  console.log(`Your app is listening on port ${listener.address().port}.`);
+  const port = listener.address().port;
+  console.log(`Your app is listening on port ${port}.`);
+
+  if (process.env.RUNNING_LOCALLY) {
+    const open = require('open');
+    open(`http://localhost:${port}`);
+  }
 });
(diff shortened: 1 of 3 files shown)


server/server.js
```diff
@@ -94,5 +94,11 @@ app.use(Sentry.Handlers.errorHandler());
 
 // Listen on App port
 const listener = app.listen(process.env.PORT, () => {
-  console.log(`Your app is listening on port ${listener.address().port}.`);
+  const port = listener.address().port;
```
clottman: Something we might consider adding to the "how to run Glitch locally" guide - maybe we should remind folks they'll have an easier time if they set PORT in .env?

taravancil:
I removed `open` and just left the documentation about setting `PORT` in `.env`. Setting `PORT` eased most of the pain I felt when I was updating server code locally.

PR #790, opened 2019-08-01
Eight truck: My stuff button on project items, notifications

whimsicallyson:
## Links
* https://eight-truck.glitch.me
* https://app.zenhub.com/workspaces/community-5d07d1359dee6059a9688176/issues/fogcreek/glitch-community-work/165
* https://app.zenhub.com/workspaces/community-5d07d1359dee6059a9688176/issues/fogcreek/glitch-community-work/248
* https://www.notion.so/glitch/My-Stuff-Default-Collection-f9eb3c8196c6431596fba526fc5b91c2#8349aaa7d9a8486fa8f3d34709929fc0

## GIF/Screenshots:
Screen Shot 2019-08-01 at 12 15 05 PM
Screen Shot 2019-08-01 at 12 15 15 PM
Screen Shot 2019-08-01 at 12 15 25 PM
Screen Shot 2019-08-01 at 12 16 05 PM

## Changes:
* Added a bookmark button to project items, clicking it toggles the project being in My Stuff, does a fun animation and shows a notification to alert you of the change
* note button shows only on hover unless you're at a mobile-sized screen

## How To Test:
* Sign in
* Add and remove projects using the button

## Feedback I'm looking for:
All good?
You can only attach a PR to one zenhub issue?


ttseng:
Nice!

I made some tweaks to the icons in /src/components/buttons/bookmark-button.js to account for focus behavior - do you mind incorporating those changes here?

Also, this might be in a follow up PR, but I think we should hide the bookmark button on your My Stuff collection page, so it doesn't feel redundant. 

image

(Users can remove projects from My Stuff by removing the project from the collection in the project dropdown).

sarahzinger:
Screen Shot 2019-08-01 at 1 20 57 PM

maybe we can remove them on the home page too? they look a little out of place there to me

@ttseng is this a bug or as designed to have the project options popover show up on the home page like this:
Screen Shot 2019-08-01 at 1 22 00 PM

(if it is a bug we can fix in another pr just figured I'd ask in case it's as designed in which case maybe we want the bookmark buttons to work similarly)

src/components/project/project-item.js
```diff
@@ -41,8 +47,25 @@ const ProfileListLoader = ({ project }) => (
...
 
 const ProjectItem = ({ project, projectOptions: providedProjectOptions }) => {
-  const projectOptions = useProjectOptions(project, providedProjectOptions);
+  const myStuffEnabled = useDevToggle('My Stuff');
   const { currentUser } = useCurrentUser();
+  const isAnonymousUser = !currentUser.login;
+  const api = useAPI();
+  const { addProjectToCollection, removeProjectFromCollection } = useAPIHandlers();
+  const { createNotification } = useNotifications();
+  const [hasBookmarked, setHasBookmarked] = useState(project.authUserHasBookmarked);
```
sarahzinger: on the search page I don't see authUserHasBookmarked as coming through do you?

src/components/project/project-item.js
```diff
@@ -67,6 +90,15 @@ const ProjectItem = ({ project, projectOptions: providedProjectOptions }) => {
                   <div className={classnames(styles.userListContainer, { [styles.spaceForOptions]: !!currentUser.login })}>
                     <ProfileListLoader project={project} />
                   </div>
+                  {myStuffEnabled && !isAnonymousUser && (
+                    <div className={styles.bookmarkButton}>
+                      <TooltipContainer
+                        type="info"
+                        target={<BookmarkButton action={bookmarkAction} initialIsBookmarked={hasBookmarked} />}
+                        tooltip="Add to My Stuff"
```
sarahzinger: I noticed in Tiffs pr we toggle between the tooltip saying add vs remove might be cool to have that here, I think she has the tooltip at the bookmark button level as opposed to at the project item level, I think when you pull in master it should be there for free and you can remove this https://github.com/FogCreek/Glitch-Community/pull/782/files#diff-325ca00b0ace128adcb0ab7292d1f95aR120

ttseng:
hm yeah, it's implemented as intended in that the idea was for you to be able to access the my stuff button on any project item, but I think it's confusing here and on your own profile page, because you might consider your own projects to be your own stuff anyway.

I think we could either 1) be consistent and show the bookmark button on all project items, or 2) only show it in places that are not guaranteed to only contain your own content (search results, project pages, team pages, collections - but not in Your Projects or for projects on your profile page).

if our goal is to help people bookmark other people's content, then I'd say we go for #2.

sarahzinger:
hmmm I'm cool to show it on user/home pages if we want just thought I'd ask!

Maybe we can get the hover functionality for the project-options dropdown to match that of the bookmark button (or vice versa)

src/components/project/project-item.js
whimsicallyson: nice! pulled it in :)

src/components/project/project-item.js
whimsicallyson: I don't. but...why?

sarahzinger: I suspect it's because this project is coming from our algolia results (which don't have a concept of a logged in user) vs pinging our own api for project details, if you want to pair on it let me know!

src/components/project/project-item.styl
```diff
@@ -102,4 +105,13 @@ a.linkBody:hover
   width: 42px
   height: 42px
   border-radius: 5px
-    
\ No newline at end of file
+    
+.bookmarkButton
+  display: block
+  margin: -17px 28px 0 0
```
ttseng: suggest tweaking this to `margin: -20px 28px 0 0` so the focus border doesn't overlap with the project item container:
image

sarahzinger:
@whimsicallyson, @ttseng and I did a little bit of work on this on Friday. We now have some slide-y animation action on hover of a project item. Also animation should now continue even if you hover away. Although I still see 2 things remaining:
- seems like bookmark buttons are not going away until you click away now (not sure why or when that started happening but I'm sure we can fix, don't think that's intended behavior!)
- bookmarkbuttons are still appearing on My Stuff collections which we should probs fix. 

we could fix in this pr or in another, whatever makes the most sense! happy to pair with ya or if you want to take the pr back feel free!

src/components/buttons/bookmark-button.styl
```diff
@@ -1,161 +1,167 @@
...
+    display: none
   }
   
   &:focus{
-    outline: none;
+    outline: none
   }
 }
-
+.visible{
```
whimsicallyson: this works, but I feel bad about it. (cc @sarahzinger @ttseng )

sarahzinger: just recapping what we spoke about on zoom, I think the thing to do here (or at least one of many solutions!) is to move this CSS from here: https://github.com/FogCreek/Glitch-Community/pull/790/files#diff-6e3c1a1f72edb1824e53651323c6f7cbR114 into BookmarkButton component directly and pass in a prop that's like "hasFlyInAnimation" (or some other better name because I'm terrible at naming things :sob:)

This gives us a few benefits like:
- it ensures we're always styling components where they are rather than where they're called, always nice to know where to find stuff and prevents us from having to style tags directly :)
- we get to avoid !important!
- we can more clearly document what the transforms are doing, making it a little bit easier to help someone else make their next button swoop in like this one does. 

it's coming together! 🎉

whimsicallyson: ok, updated so that it works without `!important`s!

PR #793, opened 2019-08-02
Superb guide: Use the server rendered page instead of the fallback in some cases

GregWeil:
## Links
https://github.com/FogCreek/Glitch-Community-Work/issues/190

## Changes:
- If js is disabled prefer to show the server rendered page over the no js error page
- If running in google translate ( https://translate.google.com/translate?hl=&sl=auto&tl=fr&u=https%3A%2F%2Fsuperb-guide.glitch.me%2F ) only redirect out if not server rendered
- The new project/sign in buttons wait until you have an anon user (aka rendering in the browser)

## How To Test:
Load up the site and check the homepage and a non ssr page
- In a modern/supported browser
- With js disabled (you can do it from dev tools > settings > advanced in ff)
- In google translate

## Feedback I'm looking for:
- Are there any dangers to leaving the ssr content visible in these two cases? Anything interactive won't work but navigating around the site should be ok.
- I opted keep ie/not built/js failed to parse blocked because they'll probably have broken css even in ssr.
- The sign in and new project buttons sometimes take a second to appear the first time you load the site in an incognito window because it has to wait until an anon user gets created. Is that ok?

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -18,7 +18,7 @@ dayjs.extend(relativeTimePlugin);
 dayjs.extend(convertPlugin);
 
 // This function is used in index.ejs to set up the app
-window.bootstrap = () => {
+window.bootstrap = (container) => {
   if (location.hash.startsWith('#!/')) {
     window.location.replace(EDITOR_URL + window.location.hash);
     return;
@@ -44,7 +44,6 @@ window.bootstrap = () => {
       </GlobalsProvider>
     </BrowserRouter>
   );
-  const container = document.getElementById('main');
 
   if (container.hasChildNodes()) {
     hydrate(element, container);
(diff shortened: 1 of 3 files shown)


whimsicallyson:
In google translate, I see a bunch of misaligned images
Screen Shot 2019-08-07 at 2 52 51 PM
and when I go to my user page, I get a magikarp error page and am redirected to the homepage (glitch.com, in English).

I don't *think* either of those are in the scope of this PR, but I wanted to point them out in case they were.

GregWeil:
Google translate mangles the dom quite a bit. I think having it translated but kind of broken is an improvement over the 'nope, english only' thing it does now, though

The user page magikarp is because user pages aren't server rendered, so I left the current prod behavior

PR #794, opened 2019-08-02
Quixotic rocket: Fix add to collection title styling, fix a proptypes warning

GregWeil:
## Links
https://quixotic-rocket.glitch.me/@glitch

## GIF/Screenshots:
image

## Changes:
- Split the emoji out of the title on team page pinned projects
- Fixed the size of the project avatar in the add project to collection pop
- Use TeamAvatars in the user options pop rather than Image

## How To Test:
Look at both of those things and see that they are the same or better

## Feedback I'm looking for:
Both `size` and `type` already have meaning for avatars, so I opted for a boolean `tiny` prop. Should avatars support every size they get rendered at?

### DIFF
--- a/src/components/images/avatar.js
+++ b/src/components/images/avatar.js
@@ -15,9 +15,9 @@ import { FALLBACK_AVATAR_URL, getProjectAvatarUrl } from 'Models/project';
 import styles from './avatar.styl';
 
 // UserAvatar
-export const Avatar = ({ name, src, color, srcFallback, type, hideTooltip, withinButton }) => {
+export const Avatar = ({ name, src, color, srcFallback, type, width, hideTooltip, withinButton }) => {
   const contents = (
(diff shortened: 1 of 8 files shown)


src/components/images/avatar.js
```diff
@@ -15,9 +16,11 @@ import { FALLBACK_AVATAR_URL, getProjectAvatarUrl } from 'Models/project';
 import styles from './avatar.styl';
 
 // UserAvatar
-export const Avatar = ({ name, src, color, srcFallback, type, hideTooltip, withinButton }) => {
+export const Avatar = ({ name, src, color, srcFallback, type, tiny, hideTooltip, withinButton }) => {
+  const size = tiny ? '16px' : '32px';
+  const className = classNames(styles[type], { [styles.tiny]: tiny });
```
sarahzinger: since there's going to be a tiny class, do you want to move the width/height stuff there? or does it have to be inline for some reason?

GregWeil:
Moved width/height from attributes to css. According to mdn those attributes are actually supposed to be the original size of the image and not the display size

PR #797, opened 2019-08-05
Nova-chance: Fix the site in the vivaldi browser

GregWeil:
## Links
https://github.com/FogCreek/Glitch-Community-Work/issues/256

## Changes:
Change window.origin references to window.location.origin

## How To Test:
Try loading up the site in vivaldi v1.6.689.40


GregWeil:
I agree. In this case we were referencing window.origin which isn't a part of the spec (but works in modern browsers I guess) so it seemed like it needed a fix either way.

GregWeil:
https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/origin
Found the spec, but it's still low support compared to the just as effective window.location.origin

PR #799, opened 2019-08-06
Nova chance: A few fixes for collection and team pages

GregWeil:
## Links
https://nova-chance.glitch.me
- https://github.com/FogCreek/Glitch-Community-Work/issues/226
- https://github.com/FogCreek/Glitch-Community-Work/issues/230
- https://github.com/FogCreek/Glitch-Community-Work/issues/229

## Changes:
- Clear the project member cache when adding/removing projects from teams, so that the team avatar shows up correctly without needing a refresh
- Provide onSubmit in the add project to collection pop so that the proptypes warning goes away
- Fixed result item highlighting when using the arrow keys to navigate search results
- Fixed using the enter button to select a search result selected using the arrow keys
- Use the underlying function rather than the useCollectionProjects hook to preload the next page's projects so react doesn't get upset when you reach the last page

## How To Test:
- Try removing and then adding a project to a team. The team should not appear in the add project popover, and should appear once the project is added back
- Go to a collection page and click the add project button. Use the arrow keys to select a project and hit enter to add it
- Go to the glitch team and click through to the last page of collections. As long as you pace yourself, the collections should all appear with their projects already in place
- Look at the result items in other places they appear and make sure none of them look wrong

## Feedback I'm looking for:
This pr fixes multiple issues, which one should I link it to?

### DIFF
--- a/src/components/pagination-controller/index.js
+++ b/src/components/pagination-controller/index.js
@@ -73,50 +73,49 @@ function PaginationController({ enabled, items, itemsPerPage, fetchDataOptimisti
     dispatchState({ type: 'previous' });
   };
 
+  useEffect(() => {
+    if (canPaginate && fetchDataOptimistically) {
+      const startIdx = state.page * itemsPerPage;
+      const nextItems = items.slice(startIdx, startIdx + itemsPerPage);
+      console.log(nextItems);
+      nextItems.forEach(fetchDataOptimistically);
+    }
+  }, [state.page, canPaginate, itemsPerPage, fetchDataOptimistically]);
+
   if (canPaginate) {
     const startIdx = (state.page - 1) * itemsPerPage;
-    const nextItems = items.slice(startIdx + itemsPerPage, startIdx + itemsPerPage * 2);
     items = items.slice(startIdx, startIdx + itemsPerPage);
-
-    if (fetchDataOptimistically) {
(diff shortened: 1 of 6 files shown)


src/components/pagination-controller/index.js
```diff
@@ -75,48 +76,46 @@ function PaginationController({ enabled, items, itemsPerPage, fetchDataOptimisti
...
           </div>
-        )}
-      </>
-    )
+          <Button data-cy="show-all" type="tertiary" onClick={() => dispatchState({ type: 'expand' })}>
+            Show all <Badge>{numItems}</Badge>
+          </Button>
+        </div>
+      )}
+    </>
```
GregWeil: This is a big whitespace change because I removed the extra nested parenthesis

src/components/popover/search.js
```diff
@@ -54,7 +55,7 @@ function useActiveIndex(items, onSelect) {
     return () => {
       window.removeEventListener('keydown', onKeyDown);
     };
-  }, [items]);
+  }, [items, activeIndex]);
```
GregWeil: Rebind the key handler when the activeIndex changes, which not as efficient but ensures that when you press enter if activates the correct button (rather than still thinking it's set to -1 like it was when the event was bound)

PR #809, opened 2019-08-09
Nova chance: Update shrinkwrap without updating every package

GregWeil:
## Links
https://nova-chance.glitch.me

## Changes:
Somehow our package.json got out of sync with shrinkwrap, which pnpm handles by ignoring the shrinkwrap entirely and updating every package to the latest version. This pr takes the last few updates to package.json and merges together the resulting shrinkwrap so they're back in sync.

## How To Test:
Try taking a remix and pulling the github project in. After a refresh and waiting for pnpm to do it's thing `git diff` will show a huge change to shrinkwrap.yaml. If you do the same with this branch, `git status` will show clean

## Feedback I'm looking for:
What does shrinkwrap do for us? If the goal is to prevent unexpected package updates we could just use exact versions in package.json.


taravancil:
Oh interesting! I wonder if this happened because the `eslint` version was changed locally?

I *think* Cassey has been working locally, and I bet she doesn't have `pnpm` installed, which is responsible for updating `shrinkwrap.yml`

GregWeil:
That looks like it, https://github.com/FogCreek/Glitch-Community/commit/4ba8c3b2ac26725f25c76581fc6eb506fd4362ab has a change to package-lock.json rather than shrinkwrap.yaml

PR #820, opened 2019-08-14
foggy-chess: Sign-in page (/login)

oliviacpu:
## Links
* Remix link: https://foggy-chess.glitch.me
* Issue-tracker link: https://app.zenhub.com/workspaces/community-5d07d1359dee6059a9688176/issues/fogcreek/glitch-community-work/265
* Specs/doc links: https://gallery.io/projects/MCHbtQVoQ2HCZZpaEieUFqhj/files/MCEJu8Y2hyDScckl30Wc2y4V2c6qLxWO8h0

## Changes:
* Organizes some of the login stuff together and adds the `SignInLayout` at `/login`

## How To Test:
* The `/login` route should render the same content as `vscode-auth` right now
* It won't match the design spec yet, there are other tickets for doing that a part at a time
* The transitions between email entry, codes, etc. are also going to be future tickets

## Feedback I'm looking for:
* The Glitch OAuth page is at `/signin` which matches what we use in the copy on the site and I'm curious if we can plan to make things consistent

- [x] ~Remove `/login` route~ Keep `/login` route
- [ ] Finish signing in and test with a branch of Editor (can they direct people here, and then "be signed in"


oliviacpu:
Reviewed most of this in our weekly hangout, but am going to make one more change to "Check for the token and give it to the Editor" so that we can try that sooner rather than later

PR #821, opened 2019-08-14
Community-side setup for targeting local API server instances (far-cushion)

dluxemburg:
We are going to make it possible to work on the API locally (without a full dev backend) and point a community remix at it to test against. The other half of this work is in [this infra repo pull](https://github.com/FogCreek/Glitch/pull/1707).

## Links
* Remix link: https://far-cushion.glitch.me/
* Issue-tracker link: in infra ZenHub [here](https://app.zenhub.com/workspaces/glitch-infrastructure-5d2db407bc23971a08b9fe2a/issues/fogcreek/glitch/1655)

## GIF/Screenshots:

Here is viewing a project on a remix:

Screen Shot 2019-08-14 at 4 14 00 PM

Here are the local server logs of it being loaded:

Screen Shot 2019-08-14 at 4 13 27 PM

## Changes:
* Add a "local" value for the `RUNNING_ON` environment variable. It tells the server and client to use the API URL that will be generated by the SSH tunnel on the infrastructure project side.
* Default to the project name, if it's not there (maybe we are running community locally too) then use the machine username (the infra project end also falls back to this). 
* `FWD_SUBDOMAIN_PREFIX` can also be set directly in `.env`.

## How To Test:
* Remix this project
* Set `RUNNING_ON` to "local"
* Clone the infrastructure repo
* Checkout the "forward" branch if it hasn't been merged yet
* Do
  * `cd api`
  * `sh/compile`
  * `bundle install`
  * (ask Daniel about getting Postgres and Redis set up)
  * `npm run local -- remix-name`
  * Verify the remix pulls data from the local API instance

## Feedback I'm looking for:
* The was this gets inserted into the constants feels weird, particularly having to do something slightly different for the server and client when the goal seems to be to keep them consistent. Maybe there is a better way?

* Once this and the other half are out we will verify that the connection between them works.
* There's plenty to do to improve the developer experience of actually using this, so there will definitely be followups.

### DIFF
--- a/server/constants.js
+++ b/server/constants.js
@@ -1,8 +1,14 @@
+const os = require('os');
 const { envs, tagline } = require('../shared/constants');
 
-// in the backend, just switch between staging and production
+// in the backend, just switch between local, staging and production
 // the client supports RUNNING_ON = development
-const currentEnv = process.env.RUNNING_ON === 'staging' ? 'staging' : 'production';
+const currentEnv = ['local', 'staging'].includes(process.env.RUNNING_ON) ? process.env.RUNNING_ON : 'production';
+const current = envs[currentEnv];
+if (currentEnv === 'local') {
+  current.API_URL = `https://${process.env.FWD_SUBDOMAIN_PREFIX}-glitch.fwd.wf/`;
+}
+
 module.exports = {
   ...envs,
   current: envs[currentEnv],
--- a/server/constants.js
+++ b/server/constants.js
@@ -5,13 +5,15 @@ const { envs, tagline } = require('../shared/constants');
 // the client supports RUNNING_ON = development
 const currentEnv = ['local', 'staging'].includes(process.env.RUNNING_ON) ? process.env.RUNNING_ON : 'production';
 const current = envs[currentEnv];
+
 if (currentEnv === 'local') {
-  current.API_URL = `https://${process.env.FWD_SUBDOMAIN_PREFIX}-glitch.fwd.wf/`;
+  const fwdSubdomainPrefix = process.env.FWD_SUBDOMAIN_PREFIX || process.env.PROJECT_NAME || os.userInfo().username;
+  current.API_URL = `https://${fwdSubdomainPrefix}-glitch.fwd.wf/`;
 }
 
 module.exports = {
   ...envs,
-  current: envs[currentEnv],
+  current,
   currentEnv,
   tagline,
 };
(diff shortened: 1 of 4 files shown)


server/constants.js
```diff
@@ -1,11 +1,19 @@
...
 
-// in the backend, just switch between staging and production
+// in the backend, just switch between local, staging and production
 // the client supports RUNNING_ON = development
-const currentEnv = process.env.RUNNING_ON === 'staging' ? 'staging' : 'production';
+const currentEnv = ['local', 'staging'].includes(process.env.RUNNING_ON) ? process.env.RUNNING_ON : 'production';
+const current = envs[currentEnv];
+
+if (currentEnv === 'local') {
+  const fwdSubdomainPrefix = process.env.FWD_SUBDOMAIN_PREFIX || process.env.PROJECT_NAME || os.userInfo().username;
```
oliviacpu: What will `os.userInfo().username` be in different contexts? Will this be for when the community site is running locally as well?

shared/constants.js
```diff
@@ -28,6 +28,15 @@ const envs = {
     FACEBOOK_CLIENT_ID: '1121393391305429',
     PROJECTS_DOMAIN: 'glitch.development',
   },
+	local: {
+    APP_URL: 'https://glitch.com',
+    API_URL: process.env.FWD_SUBDOMAIN_PREFIX ? `https://${process.env.FWD_SUBDOMAIN_PREFIX}-glitch.fwd.wf/` : null,
```
oliviacpu: Is there something other than null here we could use? Or maybe `localhost:w/e` for if we are running both locally and don't need to tunnel?

server/constants.js
dluxemburg: This is intended for running locally. It's the default on the API repo, so you run community locally and start the local API without specifying a remix name, it will just work.

shared/constants.js
dluxemburg: The reason I made it null is that picking the default happens in `server/contants.js` for the server and `webpack.config.js` for the client and I didn't want to repeat that logic again. We can't put it here anyway because this file is loaded on the client so we don't have access to the username as a fallback (we get that because we are injecting the `FWD_SUBDOMAIN_PREFIX ` env var into the compilation context). That `null` case shouldn't happen and if it does it means something upstream has gone wrong.

oliviacpu:
Thank you for clarifying that! I think this is good to go whenever. Happy to help "deploy" if you want too!

PR #822, opened 2019-08-14
Polyester wallet: bookmark button should update when we add a project to My stuff from popover

sarahzinger:
## Links
* Remix link: https://polyester-wallet.glitch.me

## Changes:
* wraps the `addProjectToCollection` function in project-item and project page to set the bookmark to filled if relevant. 

## How To Test:
* enable my stuff
* add a project to my stuff via the add to project collection popover on a collection page, user page, team page, search page or project page.
* see the button light up

## Feedback I'm looking for:
- i did start going down the path of putting this data into the project provider in state/project.js instead, but it was feeling like a lot of refactoring and maybe not worth it? but maybe you feel like it is worth it and maybe I just needed more coffee. I'm open to suggestions here.


src/components/project/project-item.js
```diff
@@ -86,8 +86,14 @@ const ProjectItem = ({ project, projectOptions: providedProjectOptions }) => {
     setIsHoveringOnProjectItem(false);
   };
 
+  const addProjectToCollectionAndSetHasBookmarked = (_project, _collection) => {
```
clottman: What do the underscores represent?

sarahzinger: if I called them project/collection we got an linting error because we've already declared project as a variable name outside the scope of this function. in general underscores to me represent that it's private to this func and that I'm not altering them, but I could call them something else if it's easier to understand maybe initialProject/initialCollection? or p/c, open to other names!

clottman: maybe `projectToAdd`?

keithk:
@modernserf Can you take a look at answering this question?
```i did start going down the path of putting this data into the project provider in state/project.js instead, but it was feeling like a lot of refactoring and maybe not worth it? but maybe you feel like it is worth it and maybe I just needed more coffee. I'm open to suggestions here.```

src/components/project/project-item.js
modernserf: I've seen underscores in function args used to represent "this value is not used" -- I would prefer a different name here

modernserf:
My suspicion is that the My Stuff collection should be handled as either part of the current user (like we do with the current user's teams) or as its own resource; in either case I don't think we need to figure that out in this PR.

sarahzinger:
@clottman @modernserf  cool names have been changed

PR #825, opened 2019-08-15
Superb guide: Memoize server rendering, standardize caching

GregWeil:
## Links
https://superb-guide.glitch.me

## Changes:
- Added /server/cache.js which defines a standard timeout cache for server side stuff
- Use that cache to memoize server rendering, based on the url and if the user is signed in
- Wait until the base project details are loaded before requesting users/teams, to save requests on 404 pages and because the users/teams endpoints error for nonexistent projects

## How To Test:
Try loading the homepage/create page/project pages/project 404 while signed in and out

I'm guessing that deploying this won't fix the site, because of the api timeouts. It should be enough to keep the cpu from capping out.


GregWeil:
Not really, but part of that is because we cache in several places in different ways. This pr consolidates server side caching which is for api requests and ssr rendering, but then there's still two levels of api caching on the client side and cloudfront in front of everything

GregWeil:
The two levels of client side caching are the expiration cache in state/api.js, and the ssr friendly one in state/api-cache.js, which returns the cached value right away but then updates it with the latest value when the server comes back

PR #831, opened 2019-08-16
Fallacious slice: Batch together api requests being made on the server

GregWeil:
## Links
https://github.com/FogCreek/Glitch-Community-Work/issues/328

## Changes:
All api requests for a similar data type within a quarter second get grouped together and sent as a single request. This will increase request time a little bit but hopefully solve the api timeouts we've been experiencing. This won't be enough for ssr project pages, as the two requests added per project to get the users/teams can't be batched.

## How To Test:
Try loading some pages. Make sure that the title and description in the document head are accurate.

## Feedback I'm looking for:
This adds up to a 250ms delay to page requests, which maybe is too much? You can run `cat .data/log/requests.log` in the community-staging console to see a probably average set of requests.


GregWeil:
Updated to only batch 50ms

sarahzinger:
I still get hit with an 429 when trying to load all projects on my project page, is that normal/expected? In some cases the user avatars load WAY faster and it's an improvement, in others they seem stuck

GregWeil:
> 
> 
> I still get hit with an 429 when trying to load all projects on my project page, is that normal/expected? In some cases the user avatars load WAY faster and it's an improvement, in others they seem stuck

This won't help with that, because the project member avatars are only loaded by the client.

sarahzinger:
ahh, ok. Should i see any noticeable improvements in particular places on the site?

GregWeil:
Technically this makes the site worse in the best case, since a lone page request will have to wait 50ms before the api request goes out. It should help a lot under 'real' load because the server will only send up to 20 api requests in a second for a given type, rather than making as many api requests as there are page requests.

sarahzinger:
ahhh ok cool. Could a batch conceivably have different users making similar requests? is there any fear that we could like accidentally leak personal data here? or is that not possible because of how these batches are created

GregWeil:
I expect/want requests from different users to be put in the same batch. The server doesn't use auth tokens in its requests, so it can only get data that's already public

PR #832, opened 2019-08-16
jazzy-packet: fix source maps in development mode

taravancil:
## Links
* [jazzy-packet.glitch.me](https://spiny-wire.glitch.me)

lol
Screen Shot 2019-08-16 at 2 22 58 PM

## GIF/Screenshots:
🔢 Look! The line numbers aren't grayed out! You can set breakpoints again!

Screen Shot 2019-08-16 at 2 43 48 PM

## Changes:
Use `eval-source-map` in dev mode instead of `cheap-module-source-map` because `cheap-module-source-map` is broken af when used with babel. I don't have time to fully understand why, but if you're interested to learn more, see [create-react-app/pull/4930](https://github.com/facebook/create-react-app/pull/4930) or [fusion-cli/pull/758](https://github.com/fusionjs/fusion-cli/pull/758) or [webpack/issues/8302](https://github.com/webpack/webpack/issues/8302)

Using `eval-source-map` will make initial builds slower, but not by much (2-3 min total), and should make rebuilds a bit faster.

## How To Test:
* Open the debugger, use `CMD-p`/`Ctrl-p` to open a file like `button/button.js`, you should be able to set breakpoints.

## Feedback I'm looking for:
- Are we cool with slowing down initial builds a bit in devolpment mode?

### DIFF
--- a/src/components/profile-list/index.js
+++ b/src/components/profile-list/index.js
@@ -34,31 +34,34 @@ const TeamItem = ({ team }) => {
 const useResizeObserver = () => {
   const ref = useRef();
   const [width, setWidth] = useState(0);
-  useEffect(() => {
-    const setWidthOfRef = () => {
-      if (ref.current) {
-        const boundingClientRect = ref.current.getBoundingClientRect();
-        if (boundingClientRect) {
-          setWidth(boundingClientRect.width);
+  useEffect(
+    () => {
+      const setWidthOfRef = () => {
+        if (ref.current) {
+          const boundingClientRect = ref.current.getBoundingClientRect();
+          if (boundingClientRect) {
+            setWidth(boundingClientRect.width);
+          }
         }
-      }
-    };
-    const debouncedSetWidth = debounce(setWidthOfRef, 100);
-    setWidthOfRef();
-
-    if (window.ResizeObserver) {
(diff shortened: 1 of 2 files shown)


src/components/profile-list/index.js
```diff
@@ -34,31 +34,34 @@ const TeamItem = ({ team }) => {
 const useResizeObserver = () => {
   const ref = useRef();
   const [width, setWidth] = useState(0);
```
taravancil: This is all prettier

src/components/profile-list/index.js
```diff
@@ -34,31 +34,34 @@ const TeamItem = ({ team }) => {
...
+        observer.observe(ref.current);
 
+        return () => {
+          observer.unobserve(ref.current);
+        };
+      }
+      window.addEventListener('resize', debouncedSetWidth);
       return () => {
-        observer.unobserve(ref.current);
+        window.removeEventListener('resize', debouncedSetWidth);
```
clottman: What's this doing/why did it need to change?

taravancil: whoa, what the heck? all i did was run prettier in that file, no clue how that happened

PR #838, opened 2019-08-20
View all X projects link should be labeled equivalently to visual label

clottman:
## Links
* https://pickled-bougon.glitch.me/
* https://github.com/FogCreek/Glitch-Community-Work/issues/324

## Screenshots 
image
image

## Changes:
* The 'View All X Projects' label for collections should also be the aria-label of the link

## How To Test:
* The View All X Projects buttons should still work
* The accessible name should match what's shown visibly
* The numbers/pluralization should work equivalently to community 

## Feedback I'm looking for:
Anything!


src/components/collection/collection-item.js
```diff
@@ -76,17 +75,16 @@ const CollectionProjects = ({ collection, isAuthorized }) => {
       </div>
     );
   }
-
+  const footerLabel = `View ${projects.length >= 3 ? 'all' : ''} ${projects.length} ${projects.length > 1 ? 'projects' : 'project'}`;
```
clottman: We need a string to pass in as the label, so we can't use the Pluralize component here.

PR #840, opened 2019-08-21
Suspended-share: Sharing a suspended project shouldn't reveal the project's avatar

clottman:
## Links
* https://suspended-share.glitch.me/~nebulous-crab (nebulous-crab is one of my suspended projects)
* https://github.com/FogCreek/Glitch-Community-Work/issues/298

## GIF/Screenshots:
image

That `og:image` tag content is: `https://cdn.gomix.com/2bdfb3f8-05ef-4035-a06e-2043962a3a13%2Fsocial-card%402x.png` 
aka
screenshot

image

## Changes:
* If project is suspended, set the default image as the image

## How To Test:
* Look at the meta tags for a suspended project, they should be the glitch fish, not anything else


clottman:
Hmm.. name you can get from the URL so I don't think we could really obscure it, but we can definitely do description!

PR #842, opened 2019-08-21
Profuse tea: AB testing with a very basic test on the secret page

GregWeil:
## Links
https://profuse-tea.glitch.me/secret
https://github.com/FogCreek/Glitch-Community-Work/issues/303
https://github.com/FogCreek/Glitch-Community-Work/issues/304

## GIF/Screenshots:
image

## Changes:
- Added a basic a/b testing framework on the server. Each test has a unique name and a set of buckets, each with a unique name, a weight, and a value
- The server reads the ab-tests cookie and randomly picks groups for any tests that aren't already assigned
- The tests are exposed in react as testname: groupvalue, so groups can store flags or whatever is needed for the test
- There's a really basic test defined, with 10% good job, 10% nice one, and 80% too bad. It's visible on the secret page, along with a button to wipe the cookie so you can try again and see what group you end up in

## How To Test:
Go to the secret page and try refreshing the page a bunch of times. You should always stay in the same group after a refresh. If you click the button, you should get a new group (in general, it isn't a guarantee)

## Feedback I'm looking for:
Group values are pretty open, they could store alternate text for an element or styling. Maybe it should only be used as flags, and keep the output in the react code?

Maybe remove the example test, or maybe leave it in to see how users get assigned


GregWeil:
Here's some code you can paste in server/ab-tests.js to measure assignment bias
```const groups = tests['Just-A-Test'];
const results = mapValues(groups, () => 0);
for (let i = 0; i < 1000000; ++i) {
  const assignment = assignGroup(groups);
  results[assignment] += 1;
}
const total = Object.values(results).reduce((sum, count) => sum + count, 0);
Object.entries(results).forEach(([group, count]) => console.log(`${group}: ${count * 100 / total}%`));```
It kept within 0.1% of the expected results

PR #843, opened 2019-08-21
circleci: Finishing setup

oliviacpu:


cypress.json
```diff
@@ -1,5 +1,7 @@
 {
-    "pageLoadTimeout": 360000,
-    "viewportWidth": 1280,
-    "viewportHeight": 1024
+  "baseUrl": "http://localhost:8000",
+  "pageLoadTimeout": 360000,
```
clottman: that's a big timeout!!

sh/cypress.sh
```diff
@@ -1,3 +1,3 @@
 #!/bin/bash
 
-npx cypress open --config baseUrl=https://$1.glitch.me
+pnpx cypress open --config baseUrl=https://$1.glitch.me
```
clottman: Do we want to have a script for being able to use cypress against localhost?

oliviacpu:
Oops this was supposed to be a draft!

keithk:
🙏🏼 YES

cypress.json
oliviacpu: We might not need that any more but it was from when I was running against remixes only and there would sometimes be a pause because of the build

oliviacpu:
I had to abandon the orb because it can't be in dependencies when we deploy to Glitch

PR #846, opened 2019-08-22
Nova chance: Add SSR data loading to user/team/collection pages

GregWeil:
## Links
https://github.com/FogCreek/Glitch-Community-Work/issues/264

## Changes:
- Re-enable data loading for project pages
- Add data loading for user/team/collection pages
- Moved user/team loading from pages/team-or-user to State/user and State/team
- Team pins are updated a little to better match the v1 api

## How To Test:
- Try browsing through the site with js disabled
- Try visiting pages signed in, signed out, and stuff that doesn't exist
- Make sure page descriptions and meta tags are correct

## Feedback I'm looking for:
All of the getFromApi functions are very similar between the client and server. Should they get moved into shared/ rather than being duplicated?

This is not safe to deploy until https://github.com/FogCreek/Glitch/issues/1729 is resolved and the server has spent some time rendering project pages


GregWeil:
This pr is kind of old and the parts that aren't covered by https://github.com/FogCreek/Glitch-Community/pull/874 can be recreated pretty easily

PR #854, opened 2019-08-27
hickory-zircon: Mark shared component

taravancil:
## Links
* [hickory-zircon.glitch.me](https://hickory-zircon.glitch.me)
* [#367](https://github.com/FogCreek/Glitch-Community-Work/issues/367)

## GIF/Screenshots:
Screen Shot 2019-08-27 at 2 21 08 PM

## Changes:
Use the shared components `Mark` component. I left the `Unmarked` component alone, it's related but isn't in the library yet.

## How To Test:
Check out the homepage and /create. The marked text should look good.


taravancil:
It's new, but I think it's fine. We'll likely encounter this with a few components, so maybe we can evaluate them altogether if/once it becomes a larger problem.

PR #858, opened 2019-08-27
Tall cheese: rebuild current user

modernserf:
## Links
* Remix link: https://tall-cheese.glitch.me/
* Issue-tracker link: https://app.zenhub.com/workspaces/community-5d07d1359dee6059a9688176/issues/fogcreek/glitch-community-work/352

## Changes:
* move superuser helpers into their own hook
* rebuild current user in redux

## How To Test:
* open two tabs to https://tall-cheese.glitch.me/
* sign into one of the tabs
* you should be signed into both tabs
* refresh one of the tabs
* it should first show an anonymous user, then your cached user state (showing a loader in place of your recent projects), then your fresh user state
* sign out of one of the tabs
* you should be signed out of both tabs

## Where am I going with this:
We've gotten pretty far just using context instead of Redux; what's the point of adding another library?
### Consistency
Our state is currently managed by a collection of interdependent context providers, each of which has its own state and interface. Most of our significant interactions require coordinating many of these at once: adding a user to a team project interacts with the api, the current user, notifications, error handlers, and the projects cache. Right now, the responsibility for wrangling all of this falls on the consumer, and this means that its easy to have two places that perform the same basic action but have inconsistent secondary effects (e.g. the notification is inconsistently displayed, the cache is not updated).

Redux encourages consistency both in how these systems are designed and in how consumers interact with them: the caller is responsible only for dispatching a message, and each system responds to that message by either transforming the state (as a reducer does) or by running a side effect (as a middleware does). This ensures that actions will have consistent behaviors, regardless of where they're dispatched from, and changing behavior (e.g. using a different API endpoint, changing the notification style) only needs to be done in one place.

### Performance
There are some significant performance "gotchas" when using "vanilla" React context -- context forces its consumer to re-render whenever its value changes, so you occasionally need to [split a context into multiple contexts for performance reasons](https://github.com/FogCreek/Glitch-Community/pull/683#discussion_r298367243), and you need to `useMemo` any functions you include to trigger actions. In practice, this means that we tend not to fix performance issues until they become painful.

Redux mitigates these issues by passing the _store_ as its context value, and setting up subscriptions on that; action creators are static and don't need to be memoized. There are still cases where one needs to optimize for performance in Redux, but in general it has better default behavior.


src/state/current-user.js
```diff
@@ -164,119 +154,179 @@ const logSharedUserError = (sharedUser, newSharedUser) => {
...
+  return {
+    toggleSuperUser: async () => {
+      if (!cachedUser) return;
+      const api = getAPIForToken(cachedUser.persistentToken);
+      await api.post(`https://support-toggle.glitch.me/support/${superUserFeature ? 'disable' : 'enable'}`);
+      window.scrollTo(0, 0);
+      window.location.reload();
+    },
+    canBecomeSuperUser:
+      cachedUser && cachedUser.projects && cachedUser.projects.filter((p) => p.id === 'b9f7fbdd-ac07-45f9-84ea-d484533635ff').length > 0,
```
GregWeil: This isn't a new change, but it could use `projects.some(...)` instead of `projects.filter(...).length > 0`

src/state/store.js
```diff
@@ -0,0 +1,15 @@
...
+import { Provider } from 'react-redux';
+import createHandlerMiddleware from './handler-middleware';
+import * as currentUser from './current-user';
+
+const store = configureStore({
+  reducer: {
+    currentUser: currentUser.reducer,
+  },
+  middleware: [...getDefaultMiddleware(), createHandlerMiddleware(currentUser.handlers)],
+  devTools: process.env.NODE_ENV !== 'production',
```
GregWeil: Does webpack inline process.env.NODE_ENV ? We write it into window.ENVIRONMENT but if we don't need it there I'd love to remove it

src/state/handler-middleware.js
```diff
@@ -0,0 +1,25 @@
...
+
+    const handlersForAction = combinedHandlers[actionProcessedByMiddleware.type];
+    if (!handlersForAction) return actionProcessedByMiddleware;
+
+    handlersForAction.forEach((handler) => handler(actionProcessedByMiddleware, store));
+    return actionProcessedByMiddleware;
+  };
+};
+
+export default createHandlerMiddleware;
```
GregWeil: Could you add a comment to this file explaining what it does? (It seems like it takes several sets of handlers and creates a thing that applies all of them when called by redux)

src/state/current-user.js
```diff
@@ -164,119 +154,179 @@ const logSharedUserError = (sharedUser, newSharedUser) => {
...
+    clear: () => dispatch(actions.loggedOut()),
+  };
+};
 
+export const CurrentUserProvider = ({ children }) => {
+  const dispatch = useDispatch();
   useEffect(() => {
-    if (ready) load();
-    // for easier debugging
-    window.currentUser = cachedUser;
```
GregWeil: This got lost in the translation, was that intentional?

src/state/current-user.js
```diff
@@ -1,32 +1,42 @@
...
-  coverColor: null,
-  emails: [],
-  features: [],
-  projects: [],
-  teams: [],
-  collections: [],
-};
+// takes a generator that yields promises,
+// returns an async function that restarts from the beginning every time it is called.
+function runLatest(fn) {
```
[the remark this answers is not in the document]
GregWeil: It doesn't look like there's anything there, is that the right link?

src/state/current-user.js
modernserf: you'll get a much more detailed view into this with the redux dev tools, though i can put the redux store on `window` if you'd like

src/state/current-user.js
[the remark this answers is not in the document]
modernserf: ugh, it seems to have eaten my changes. serves me right

src/state/store.js
modernserf: I think it does, but I'll use `window.ENVIRONMENT` here for now; we can reevalutate that later

modernserf:
Updated with feedback & much more documentation

src/state/store.js
```diff
@@ -0,0 +1,15 @@
...
+import { Provider } from 'react-redux';
+import createHandlerMiddleware from './handler-middleware';
+import * as currentUser from './current-user';
+
+const store = configureStore({
+  reducer: {
+    currentUser: currentUser.reducer,
+  },
+  middleware: [...getDefaultMiddleware(), createHandlerMiddleware(currentUser.handlers)],
+  devTools: window.ENVIRONMENT === 'dev',
```
GregWeil: This blocks server side rendering because of the `window` reference. It can ether use isBrowser Utils/constants to protect the call or move ENVIRONMENT into constants. (or just switch back to process.env if that ssr's ok).

Separately, does creating the store here mean that every ssr render will reuse the same store? I think that's ok since ssr won't actually call any actions. Are there any risks there?

modernserf: i honestly have no idea; the prudent thing might be to wrap the whole business in `isBrowser`

modernserf: after thinking about this over the weekend, I think the right approach for now is for SSR to include the redux store but disable devtools and not actually dispatch any actions; that way everything will work as if it has the initial state

src/state/store.js
```diff
@@ -0,0 +1,16 @@
...
+import { isBrowser } from 'Utils/constants';
+import createHandlerMiddleware from './handler-middleware';
+import * as currentUser from './current-user';
+
+const store = configureStore({
+  reducer: {
+    currentUser: currentUser.reducer,
+  },
+  middleware: [...getDefaultMiddleware(), createHandlerMiddleware(currentUser.handlers)],
+  devTools: isBrowser() && (window.ENVIRONMENT === 'dev'),
```
GregWeil: `isBrowser` is a constant

PR #866, opened 2019-08-28
Curvy-bandicoot: maintain same page if a project or collection gets deleted

clottman:
## Links
* https://curvy-bandicoot.glitch.me/
* https://github.com/FogCreek/Glitch-Community-Work/issues/354

## GIF/Screenshots:
https://media.giphy.com/media/W1llYOepQbqaM0bHbf/giphy.gif

## Changes:
* if the number of pages changes, go to the previous page if the last page no longer exists. otherwise stay on the current page.
* removed a console log I think I accidentally added once 

## How To Test:
* delete a project from various pages of your recent projects, or a collection from your collections section on your profile

## Feedback I'm looking for:
- Anything I might be doing wrong about these reducers


clottman:
@GregWeil what would the interaction have been if number of pages drops by more than one? The only way I can think of that happening is if the entire project list refreshed based on another tab or another user's changes, in which case I think it would make more sense to reset the page back to 1. We don't currently ever update the project list as far as I know though, besides on page refresh.

GregWeil:
It's pretty unlikely, but I think it can be done by removing yourself from a team (and several projects) in another tab, then letting https://github.com/FogCreek/Glitch-Community/blob/master/src/state/user.js#L39 update the page.

Either way, having the component get into a bad state if the number of projects changes by more than one seems dangereous.

src/components/pagination-controller/index.js
```diff
@@ -31,12 +31,20 @@ const paginationReducer = (oldState, action) => {
...
-      return {
-        ...oldState,
-        page: 1,
-        announce: `Showing page 1 of ${action.totalPages}`,
-      };
+    case 'numberOfPagesChanged':
+      if (oldState.page > action.totalPages) {
+        let pageToReturnTo = 1;
+        if (oldState.page > 1) {
+          pageToReturnTo = oldState.page - 1;
```
GregWeil: This will prevent going to page -1, but still won't ensure it's not too high

src/components/pagination-controller/index.js
```diff
@@ -31,12 +31,23 @@ const paginationReducer = (oldState, action) => {
...
+        }
+        if (pageToReturnTo > action.totalPages) {
+          pageToReturnTo = action.totalPages - 1;
+        }
+        return {
+          ...oldState,
+          page: pageToReturnTo,
+          totalPages: action.totalPages,
+          announce: `Showing page ${action.totalPages - 1} of ${action.totalPages}`,
```
GregWeil: This doesn't use the same number as the actual page, so it'll announce the wrong thing when there's only one page

PR #871, opened 2019-08-29
continuous-echo: Add join/leave project button to ~ pages

taravancil:
## Links
* [continuous-echo.glitch.me](https://continuous-echo.glitch.me)
* [#356](https://github.com/FogCreek/Glitch-Community-Work/issues/356)

## GIF/Screenshots:
**Joining and leaving a team project**
join-leave-team-member

**Leaving a project you were invited to**

## Changes:
This adds a Join/Leave Project button to project pages.

- Moved the leave project popover into a reusable component
- Added `userIsProjectTeamMember` to `Models/project.js`

## How To Test:
- Go to the ~ page for one of your team's projects. You should be able to join and leave directly, with no popovers or confirmation
- Join a project that's *not* a team project (ping me if you need to be invited to a project)
  - Visit the ~ page
   - Click the "Leave Project" button, you should see a popover that asks you to confirm that you want to leave
  - After leaving, you should *not* see the "Join Project" button
- Visit the ~ page of a project you are the only admin for. You should not see the "Leave Project" button
- Open the project options pop (accessible from collections) for some projects
  - Confirm that you can still join and leave those projects as expected
  - Confirm that join/leave buttons are not shown unless you actually can join or leave

- [x] Project membership data is stale, which means the effect that sets `isMembership` doesn't run when you join/leave a project. Need to figure this out and fix it!
- [ ] Design work
- [x] QA


taravancil:
@ttseng I haven't consulted you yet on this button at all - what do you think of how it turned out?

Happy to delay deploying this if you'd like to think it through a bit first!

PR #873, opened 2019-08-30
Acidic-tomate: standardize error handling on image upload failures

clottman:
## Links
* acidic-tomate.glitch.me
* https://github.com/FogCreek/Glitch-Community-Work/issues/344

## GIF/Screenshots:
if the upload itself failed: 
image
If the upload succeeded but we can't process the image: 
image

## Changes:
* Only one error notification ever shows up on file upload failures. Before, you might have got an 'upload failed' notification followed immediately by a 'something went wrong, try refreshing?' notification because we kept acting as if the upload succeeded when it failed
* We don't let errors bubble up to sentry if they're just file upload errors - more than likely it's the user's problem, not ours
* removes a console.log that got left in when I was working on the team analytics section

## How To Test:
* Try uploading [the image in this zip file](https://fogcreek.slack.com/archives/CJS8MABLZ/p1567098131017400?thread_ts=1567097911.016700&cid=CJS8MABLZ) as your cover photo for a team or user, or your avatar. 
  * this still fails silently on project avatar uploads, but I don't know enough about image processing/ how our uploads work to figure out why

## Feedback I'm looking for:
anything really

### DIFF
--- a/src/state/error-handlers.js
+++ b/src/state/error-handlers.js
@@ -21,12 +21,20 @@ function handleCustomError(notify, error) {
   return Promise.reject(error);
 }
 
+function handleImageUploadError(notify, error) {
+  console.error(error);
+  notify('Unable to process image. Please try another image.');
+  // swallow the error
+  return false;
+}
+
 const useErrorHandlers = () => {
   const { createErrorNotification } = useNotifications();
   return {
     handleError: (error) => handleError(createErrorNotification, error),
     handleErrorForInput: (error) => handleErrorForInput(createErrorNotification, error),
     handleCustomError: (error) => handleCustomError(createErrorNotification, error),
+    handleImageUploadError: (error) => handleImageUploadError(createErrorNotification, error),
   };
 };
 
--- a/src/state/team.js
+++ b/src/state/team.js
@@ -17,7 +17,7 @@ export function useTeamEditor(initialTeam) {
   const { currentUser, update: updateCurrentUser } = useCurrentUser();
   const { uploadAssetSizes } = useUploader();
   const { createNotification } = useNotifications();
-  const { handleError, handleErrorForInput } = useErrorHandlers();
+  const { handleError, handleErrorForInput, handleImageUploadError } = useErrorHandlers();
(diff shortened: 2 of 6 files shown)


src/state/error-handlers.js
```diff
@@ -21,12 +21,20 @@ function handleCustomError(notify, error) {
   return Promise.reject(error);
 }
 
+function handleImageUploadError(notify, error) {
+  console.error(error);
```
sarahzinger: what do you think about using captureException instead of console.erroring?

src/state/team.js
```diff
@@ -129,7 +129,10 @@ export function useTeamEditor(initialTeam) {
       assets.requestFile(
         withErrorHandler(async (blob) => {
           const { data: policy } = await getAvatarImagePolicy({ team });
-          await uploadAssetSizes(blob, policy, assets.AVATAR_SIZES);
+          const success = await uploadAssetSizes(blob, policy, assets.AVATAR_SIZES);
+          if (!success) {
+            return;
```
sarahzinger: if there's no sucesss shouldn't we throw? (apologies if this is me misunderstanding how this works!)

src/state/error-handlers.js
sarahzinger: ah i see your comment "We don't let errors bubble up to sentry if they're just file upload errors - more than likely it's the user's problem, not ours" if we know there are specific user errors here can we give the user more helpful feedback? are they uploading files that aren't images?

src/state/team.js
clottman: There's a try/catch handler in `uploadWrapper` from `uploader.js`, so we've already thrown and notified the user about an error.

src/state/error-handlers.js
clottman: We don't know what the error was more specifically as far as I can tell - maybe the upload failed because the asset file was corrupted, it was the wrong type, or too big. 

The reason for `console.error` is that a tech-savvy user might look there for more info on what went wrong; I'm open to taking that out too if we don't want it there.

sarahzinger: yah your reasoning makes sense to me!

PR #874, opened 2019-08-30
Superb guide: Clean up the standard data loading code and move it to shared

GregWeil:
## Links

## Changes:
- Made standard getThing functions in the shared folder. These are used to load a full entity to render a page with. Right now the server doesn't use them at all, but with ssr in place it could
- Cleaned up user/team property naming, particularly pins/teamPins -> pinnedProjects to match the name in the api
- Use orderBy to sort team users/projects because I always have to look up how custom sort ordering works
- Removed a window reference in project items

## How To Test:
Try navigating around the site and visiting different collection/project/team/user pages, and pages for names that don't exist

## Feedback I'm looking for:
- I kind of think models/ should go in shared/ too, but editing anything shared with the server causes a full webpack rebuild which takes a very long time
- These functions don't use the new batched calls on the server. Should batching happen on the client side too, or should it get more fancy to support that?


GregWeil:
@modernserf Could you take a look at this, since you've been investigating redux loading? It probably makes the move to shared less helpful, but the rest of it is still worthwhile and unblocks ssr for the rest of the site.

PR #876, opened 2019-09-04
WIP aka JUST A DRAFT: Working-splash: resume coding on the current project page's project

clottman:
## Links
* working-splash.glitch.me
* https://github.com/FogCreek/Glitch-Community-Work/issues/378

## Changes:
* If you're on a Glitch project page, the 'resume coding' button in the header should go to that project
* If you're on any other page, the button goes to your last modified project

## How To Test:
* Click the 'resume coding' button from various places on the site (must be logged in)


taravancil:
Code looks good! 👍🏻

I have a question about the UX of this though - I'm not sure it makes sense that the "Resume Coding" button should behave like this when you're on a project page that *you're not a member of*. I use the "Resume Coding" button pretty regularly and I think it will feel weird to sometimes be taken to a project I can't edit.

@ttseng what do you think?

clottman:
That's a good question @taravancil - in that case I wonder if it makes sense at all for resume coding to do different things based on if you're on a project page that you can edit, compared to any place else on the site. That would mean a button that looks the same on two pages that largely look the same (your own project vs another person's project) have different behaviors. 🤔

taravancil:
> a button that looks the same on two pages that largely look the same have different behaviors

Yeah, that's tough. I'm not sure what the best option is. The original ticket has the "Ready For: Design" label, so maybe we should hold off on this PR and mark it as WIP until Tiff or Yasi has a chance to think it through?

ttseng:
This is an interesting approach.  

I think the underlying goal is to make it clearer to users how to jump back into editing a project they own (rather than open the editor for _any_ project, regardless of whether you own it or not).  

But I think the `resume coding` button is not the best way to do this because
1) it's not standard for navbar buttons to be context-dependent (navbar content is usually meant for quick, consistent actions you want to access anywhere)
2) the wording `resume` is confusing for projects you don't own since`resume` implies you've already touched the code before
3) the placement of the button in the nav is far from the project info, so it's not obvious that it's linked to the project you're viewing.

What if, instead, we gave CTA styling to the Edit Project button in the Profile Wrap, so it stands out to project authors but doesn't get applied unless you can actually edit the project directly (i.e., the CTA styling is *not* applied to the `View Source` button for projects that you don't own)?

image

clottman:
Closed as 'wrong solution'

PR #877, opened 2019-09-04
Checker eclipse: handle collection->project and project->user/team in redux

modernserf:
## Links
* Remix link: https://checker-eclipse.glitch.me
* Issue-tracker link: https://app.zenhub.com/workspaces/community-5d07d1359dee6059a9688176/issues/fogcreek/glitch-community-work/405

## Changes:
* create a resource manager redux store & hooks
* replace collection and project contexts with resource manager
* get collection curator with resource manager
* update project & collection membership optimistically, instead of re-fetching on every change

## How To Test:
* log into https://checker-eclipse.glitch.me
* test out the following actions:
  - join & leave a team project
  - add & remove a project from a team
  - add & remove a project from a collection
  - toggle bookmarked status on a project

## Feedback I'm looking for:
* How can the resource manager be better documented?
* What approach should we take for handling errors?


oliviacpu:
my-stuff
Adding a project to a collection will update that collection on the page. This doesn't seem to work for "My Stuff" (or bookmarks?) though.

modernserf:
oops, i had started going down a different route with My Stuff and forgot to revert it; should be working now

oliviacpu:
As far as documenting this, I am thinking about what questions I would have if I were to try to "get started" with this in my own remix. 

I can follow what you're doing to declare what resources you need (`useResources(type, id, childType)`) and how to handle when you need multiple types of resources (`allReady`). One thing I'm wondering is how often "childType" is relevant and what is considered a child of what (getting members of a team, projects of a team, projects of a collection).

We don't need to redocument redux, but a description of how that fits into what we're doing here (you've checked out these concepts, now here's how we're applying them) could be nice, or patterns/ways of thinking that you've settled on for other people to reproduce if they need to join you.

I am always happy to be a "rubber duck" that can type to talk through and document things with you.

modernserf:
@oliviacpu I updated the documentation (particularly around "children"/"references")

keithk:
👋  This has been open for awhile and seems out of date, do you think we should table it or get it up to date with master?

PR #878, opened 2019-09-04
Fallacious slice: Remove the feature switch in front of email invites

GregWeil:
## Links
https://fallacious-slice.glitch.me
https://github.com/FogCreek/Glitch-Community-Work/issues/381

## GIF/Screenshots:
image

## Changes:
- Removed the email invites dev toggle
- Fixed a proptypes error with the invite by email button
- Some changes to useCheckedDomains to prevent duplicate requests
- Added a missing param to the memo so it consistently shows the whitelist option

## How To Test:
Try inviting yourself to a team via email

### DIFF
--- a/src/components/team-users/add-team-user.js
+++ b/src/components/team-users/add-team-user.js
@@ -38,7 +38,7 @@ const InviteByEmail = ({ result: email, active, onClick }) => {
   const { current: color } = useRef(randomColor({ luminosity: 'light' }));
   return (
     <ResultItem onClick={onClick} active={active}>
-      <UserAvatar user={{ color }} hideTooltip />
+      <UserAvatar user={{ id: 0, color }} hideTooltip />
       <ResultInfo>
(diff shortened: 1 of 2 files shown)


whimsicallyson:
when I invite an existing user to my team, the avatar immediately shows up -- but with an email invite, I need to refresh the page to see them. Can we make this interaction match?

GregWeil:
> 
> 
> when I invite an existing user to my team, the avatar immediately shows up -- but with an email invite, I need to refresh the page to see them. Can we make this interaction match?

It can't be done that easily, because we don't actually have the user object. I can add a fake user avatar, but it won't let you resend or revoke the invite because there's no user id. (Another option is to reload the invited list, but that would probably shuffle the users)

GregWeil:
Reworked invitees, it's way fancier now and shows people invited by email right away rather than waiting for a refresh.

PR #851, opened 2019-08-26
Fixed event: starting shared-components

whimsicallyson:
## Links
* https://fixed-event.glitch.me/
* https://app.zenhub.com/workspaces/community-5d07d1359dee6059a9688176/issues/fogcreek/glitch-community-work/363
https://app.zenhub.com/workspaces/community-5d07d1359dee6059a9688176/issues/fogcreek/glitch-community-work/365
partially https://app.zenhub.com/workspaces/community-5d07d1359dee6059a9688176/issues/fogcreek/glitch-community-work/360

## Changes:
* We shouldn't have to use the Emoji or Button code from ~community anymore (in fact, they're already deleted in this remix). I'm still working on AnimationContainer in one spot though, so its not removed.

## How To Test:
* Click around, look for anything broken or inconsistent. There will be some slight changes in font size, spacing, colors, because we're condensing down to a smaller number of total options, but nothing should look actively *wrong*. If it does, point it out!

## Feedback I'm looking for:
Is anything broken? Does it all seem ok still?

## Things left to do before deploying:
- Disabled button styling should work properly once https://github.com/FogCreek/shared-components/pull/38 is merged in.

ps @modernserf you may be interested in this, as a reviewer or just observer.

PR #886, opened 2019-09-05
Telling rose: Button and Icon components (take 2)

whimsicallyson:
## Links
* telling-rose.glitch.me
* https://app.zenhub.com/workspaces/community-5d07d1359dee6059a9688176/issues/fogcreek/glitch-community-work/363
https://app.zenhub.com/workspaces/community-5d07d1359dee6059a9688176/issues/fogcreek/glitch-community-work/365

## Changes:
* Because #851 got all manner of borked up, I created a new remix and manually transferred over all the relevant changes

## How To Test:
* Click around, make sure things look correct and work?

## Feedback I'm looking for:
Hopefully this is (finally) ready to get deployed

not screw up merging!


src/components/project/project-actions.js
```diff
@@ -15,24 +17,24 @@ ShowButton.propTypes = {
...
 export const EditButton = ({ name, isMember, size }) => (
-  <Button href={getEditorUrl(name)} size={size}>
+  <Button as="a" href={getEditorUrl(name)} size={size}>
     {isMember ? 'Edit Project' : 'View Source'}
   </Button>
 );
 EditButton.propTypes = {
   name: PropTypes.string.isRequired,
   isMember: PropTypes.bool,
-  size: PropTypes.oneOf(SIZES),
```
whimsicallyson: From the `fixed-event` massive PR, @clottman asks:
Why do we need this change? Can we export the acceptable button sizes from the shared components library so we can still have prop type checking as strict as we previously did? @modernserf 

AFAICT, there isn't a way to export the sizes from `shared-components`. Is this something we want to add in, or is changing the prop type check to 'string' acceptable?

modernserf: the <Button> component itself checks the sizes; we don't need to also check them for EditButton

whimsicallyson: removed checks!

PR #888, opened 2019-09-06
Historical salesman: Badge shared component

whimsicallyson:
## Links
* historical-salesman.glitch.me
* https://app.zenhub.com/workspaces/community-5d07d1359dee6059a9688176/issues/fogcreek/glitch-community-work/362

## Changes:
* only a few badges on the community site! 
  - count in the "add to collection" pop if that project's already in 3+ collections
  - counts of search results
  - when a project in a collection list is private, which I turned into an Icon instead

## How To Test:
* Check those 3 ^^ and make sure they look good!


taravancil:
The vertical alignment of the private icon looks a bit off. Can we give a 3.5px or 4px `margin-top` instead?

Screen Shot 2019-09-09 at 1 45 49 PM

src/components/project/project-item.styl
```diff
@@ -123,4 +123,9 @@ a.linkBody:hover
...
   width: 90%
   &:empty
-    margin-bottom: 0
\ No newline at end of file
+    margin-bottom: 0
+    
+.privateIcon
+  margin-left: 5px
+  margin-top: 3px
+  color: private-foreground !important
```
taravancil: > icons are implemented as SVGs and match the color of their surrounding text

Instead of using `!important`, could we wrap the icon in a `<span className={styles.privateIcon}`> so that it will inherit its color from the `<span>`?

src/components/project/project-item-small.js
```diff
@@ -26,9 +26,9 @@ const ProjectItemSmall = ({ project }) => (
           <span className={styles.projectName}>{project.domain}</span>{' '}
         </Text>
         {project.private && (
-          <Badge type="private" aria-label="private">
+          <Icon icon="private" className={styles.privateIcon} aria-label="private">
```
taravancil: Can we make this a self-closing tag (`<Icon />`) and get rid of line 30?

Also according to the shared components docs, I think we'll want to use `alt` instead of `aria-label`:

> alt
    The accessible label for the icon. If an icon is paired with a text label (e.g. in a button) you should leave this blank.

src/components/project/project-item.styl
whimsicallyson: good call!

PR #889, opened 2019-09-06
Profuse tea: Rework webpack and ssr transpile for faster startup

GregWeil:
## Links
https://profuse-tea.glitch.me

## Changes:
- When `DEPLOY_ENV` is set to `production` the server expects the build to be managed elsewhere
  - Webpack is run via `prestart` in watch mode (rather than being spawned by the server code, so editing the server doesn't cause the client to rebuild from scratch!)
  - A babel transpile is started in watch mode for ssr code (rather than using babel register)
  - This ensures that the server code waits for nothing and can immediately server requests using the last build
- When `DEPLOY_ENV` is set to `ci` the server acts the same as above but the two watchers do not get started, so it can serve a static build instead
- When `DEPLOY_ENV` is set to anything else, the site stalls requests in order to use the latest code
  - Webpack is run in the server process using webpack-dev-middleware
  - Babel is run in the server process using babel register (like it is on prod now)
- The client code is built into build/client instead of build/, so that webpack doesn't clean up files from other builds
- I moved shared/aliases.js to the root folder because it's used by webpack and eslint, so the server needs to restart when it is changed
- I told webpack not to output the name/path of every file it parses because it floods the logs

## How To Test:
-Try running the stop command against the project and see how quickly you can reconnect. (I'm averaging about 30s but sometimes it hits me with a pnpm reinstall and takes a minute or so)
- Join the project and make edits in different places and make sure the watchers all still do what they should

## Feedback I'm looking for:
- `DEPLOY_ENV` controls how the site manages builds, should it get renamed to something like `BUILD_TYPE=static|watcher|dev`?
- I don't have any handling for running on prod before the initial ssr build finishes. it'll just fail to render. What should it do, if anything else?


server/render.js
```diff
@@ -2,60 +2,67 @@ const path = require('path');
...
 
 const React = require('react');
 const ReactDOMServer = require('react-dom/server');
 const { Helmet } = require('react-helmet');
-setImmediate(() => requireClient()); // transpile right away rather than waiting for a request
+setImmediate(() => {
+  try {
+    requireClient();
+  } catch (error) {
+    // try importing right away so we don't have to wait
```
clottman: Should we log here so we can tell if this is ever happening somewhere it shouldn't?

GregWeil: Probably yes. I'll captureException too in case it happens to a prod server somehow

package.json
```diff
@@ -17,18 +17,24 @@
     "./src/polyfills.js"
   ],
   "scripts": {
-    "prestart": "nodemon --watch server --watch shared --exec 'npm run lint:server' &",
-    "start": "forever --minUptime 5000 --spinSleepTime 1000 -c 'nodemon --watch server --watch shared --watch views --exitcrash' ./server/server.js",
+    "prestart": "npm run watch:lint:server & [ \"$DEPLOY_ENV\" == \"production\" ] && (npm run watch:babel & npm run watch:webpack &)",
```
GregWeil: ci doesn't seem to like this (neither do I)

Can it be done in a more legible way? (goal is run watch:lint:server in background. if production, also run watch:babel and watch:webpack in the background)

clottman: I'd recommend using [if-env](https://www.npmjs.com/package/if-env) or [per-env](https://github.com/ericclemmons/per-env) here to get more readable syntax for what you're trying to do. (And have it work in all environments)

GregWeil: That looks nice, I'll add per-env

GregWeil: Actually using if-env, per-env only works on NODE_ENV but I need DEPLOY_ENV

package.json
```diff
@@ -17,19 +17,25 @@
     "./src/polyfills.js"
   ],
   "scripts": {
-    "prestart": "nodemon --watch server --watch shared --exec 'npm run lint:server' &",
-    "start": "forever --minUptime 5000 --spinSleepTime 1000 -c 'nodemon --watch server --watch shared --watch views --exitcrash' ./server/server.js",
+    "prestart": "npm run watch:lint:server & \n if if-env DEPLOY_ENV=production; then \n npm run watch:babel & npm run watch:webpack & \n fi",
```
GregWeil: I'm not a fan of the newlines here, but this is what I managed to get working on both glitch and ci

PR #892, opened 2019-09-09
Sumptuous rainbow: styled components SSR, render '→' as icon

modernserf:
## Links
* Remix link: http://sumptuous-rainbow.glitch.me
* Issue-tracker link: https://app.zenhub.com/workspaces/community-5d07d1359dee6059a9688176/issues/fogcreek/glitch-community-work/111

## Changes:
* add babel plugin, SSR for styled components 
* replace arrow unicode with icon for consistent alignment

## How To Test:
* visit http://sumptuous-rainbow.glitch.me
* there should be no 'flash of unstyled content' while shared components styles load
* arrows in buttons should be consistently rendered & aligned across browsers & platforms


src/components/arrow/index.js
```diff
@@ -1,3 +1,4 @@
 import React from 'react';
+import { Icon } from '@fogcreek/shared-components';
 
-export default () => <span aria-hidden="true">→</span>;
+export default () => <Icon icon="arrowRight" />;
```
clottman: This should still be aria-hidden - is it?

whimsicallyson: also, now that its just using an Icon component, would it make more sense to import directly when needed (as opposed to pulling in this file)?

server/render.js
```diff
@@ -54,29 +57,35 @@ const requireClient = () => {
...
-    HOME_CONTENT,
-    SSR_SIGNED_IN,
-    EXTERNAL_ROUTES,
-  });
-
-  const html = ReactDOMServer.renderToString(page);
-  const helmet = Helmet.renderStatic();
-  const context = { AB_TESTS, API_CACHE, EXTERNAL_ROUTES, HOME_CONTENT, SSR_SIGNED_IN, ZINE_POSTS };
-  return { html, helmet, context };
+  try {
```
clottman: We're not `catch`ing anything - do we need the `try/finally`? Or is there an error we should be catching & logging?

modernserf: we're not catching anything, but the `.seal()` needs to happen even if this function crashes; otherwise this will leak memory (per the styled-components docs)

modernserf:
re SSR -- styled components insert their stylesheets when the component mounts. this is invisible if the component renders client-side, but causes a flash of unstyled content if the component is server-side rendered (because the HTML is already present when the react component mounts.)

src/components/arrow/index.js
modernserf: yes, it is aria-hidden

whimsicallyson: updated all arrows in #897

PR #896, opened 2019-09-10
Decisive kilometer: fix button text on join/leave 

sarahzinger:
## Links
* Remix link: https://decisive-kilometer.glitch.me/@test-sarah/joining-and-leaving
* Issue-tracker link: https://app.zenhub.com/workspaces/community-5d07d1359dee6059a9688176/issues/fogcreek/glitch-community-work/116

## Changes:
* subscribe to useProjectMembers hook so that when the project members updates, our button text updates. I had originally thought it was strange we don't have success messages which is why I opened the ticket, but it looks like that's how the functionality works on the site in general. Figured I'd just fix the button text, which also helps inform that you've successfully joined or left a project.

## How To Test:
* join a project (either a team project or just as a reg user)
* add it to a collection
* try to leave the project from that collection
* if it's a team project you should be able to join it again as the button text has updated. 

## Feedback I'm looking for:
anything

### DIFF
--- a/src/state/project-options.js
+++ b/src/state/project-options.js
@@ -38,8 +38,13 @@ const useDefaultProjectOptions = () => {
       reloadProjectMembers([project.id]);
     }, handleError),
     leaveProject: withErrorHandler(async (project) => {
-      await removeUserFromProject({ project, user: currentUser });
-      reloadProjectMembers([project.id]);
+      console.log("default leaveProject")
+      try {
+        await removeUserFromProject({ project, user: currentUser });
+        reloadProjectMembers([project.id]);
+      } catch (error) {
+        console.log("error", error)
+      }
     }, handleError),
     // toggleBookmark is defined here and on state/collection and are very similar. Their only differences are how they modify state.
     // we'll probably want to revisit condensing these when we have a centralized state object to work off of.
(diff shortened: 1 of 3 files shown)


src/state/project-options.js
```diff
@@ -78,7 +78,8 @@ export const useProjectOptions = (project, { user, team, collection, ...options
   }, [user, team, project]);
 
   const isLoggedIn = !!currentUser.login;
-  const isProjectMember = userIsProjectMember({ project, user: currentUser });
+  const { value: members } = useProjectMembers(project.id);
+  const isProjectMember = !!(members && members.users && members.users.some(({ id }) => id === currentUser.id));
```
taravancil: Oh I see, we need to do this because `userIsProjectMember` in `Models/project` is *not* subscribed to `useProjectMembers`.

Hmm, it might be nice to re-use what's already in [Models/project](https://github.com/FogCreek/Glitch-Community/blob/master/src/models/project.js#L46) for determining project membership. Would it be weird to modify `project.permissions` and then pass that to `userIsProjectMember`?

sarahzinger: @taravancil updated `userIsProjectMember` what do you think?

PR #903, opened 2019-09-11
Search results styling

whimsicallyson:
## Links
* https://historical-salesman.glitch.me
* https://app.zenhub.com/workspaces/community-5d07d1359dee6059a9688176/issues/fogcreek/glitch-community-work/422

## GIF/Screenshots:
Screen Shot 2019-09-11 at 10 57 17 AM

## Changes:
* Updated styling on search results to better match previous (and not be so squished)

## How To Test:
* Search for a thing, specifically one with lots of results

### DIFF
--- a/src/components/search-results/search-results.styl
+++ b/src/components/search-results/search-results.styl
@@ -26,3 +26,6 @@
   margin-bottom: 2rem
   .resultsContainer
     align-items: end
+    
+span[data-module="Badge"]
+  margin-left: 5px
--- a/src/components/search-results/index.js
+++ b/src/components/search-results/index.js
@@ -26,7 +26,7 @@ const FilterContainer = ({ filters, activeFilter, setFilter }) => {
     ),
   }));
 
-  return <SegmentedButton value={activeFilter} options={buttons} onChange={setFilter} />;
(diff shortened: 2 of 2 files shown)


src/components/search-results/search-results.styl
```diff
@@ -25,4 +25,7 @@
   padding-bottom: 2rem
   margin-bottom: 2rem
   .resultsContainer
-    align-items: end
\ No newline at end of file
+    align-items: end
+    
+span[data-module="Badge"]
```
sarahzinger: this is interesting syntax that I'm not that familiar with, you mind explaining it a bit?

whimsicallyson: because the Badge components (and, I believe, all of the shared-components we'll be using) have attribute tags with their module names in them, this is an easy way to target without adding new class names everywhere. I needed to start with `span` to get the specificity to override the default module styling, though.

Examples here: https://css-tricks.com/almanac/selectors/a/attribute/

sarahzinger: ahh ok that makes sense. Is this the general best practice that we're going with moving forward?

whimsicallyson: I think it probably should be, when relevant, as it keeps us from having to add too many classnames and/or worry about nesting selectors in the CSS to make sure we're targeting the correct thing.

PR #910, opened 2019-09-13
User Options Pop Button Wrap & Top Search Result Vertical Alignment

ttseng:
## Links
* Remix link: https://vaulted-mirror.glitch.me/search?q=space
* Issue-tracker link: 
https://github.com/FogCreek/Glitch-Community-Work/issues/351
https://github.com/FogCreek/Glitch-Community-Work/issues/399

## GIF/Screenshots:
Original on left, changes on right
image

## Changes:
* Top search results that include projects will be vertically aligned to the top of the project-item-container
* Also changed the buttonWrap of the user-options-pop to account for long team names.

## How To Test:
* Search for terms that give multiple top results ("space", "react") - the items should be vertically aligned
* Search for terms that give top results that don't include a project - there shouldn't be any extra margins applied (e.g. "scientiffic")
* Check that long names wrap well in the user options pop

## Feedback I'm looking for:
Is there a better way to determine if the top results include a project? 
```
  const topResultsIncludesProject = showTopResults && values(mapValues(searchResults.topResults, (result) => result.type)).includes('project');
```

### DIFF
--- a/src/components/images/avatar.styl
+++ b/src/components/images/avatar.styl
@@ -10,7 +10,7 @@ img.avatar
   height: 32px
 img.tiny
   vertical-align: bottom
-  width: 16px
-  hx
-  min-width: 15px
-  margin-top: -2pxisplay: inline-block
+  width: 15px
+  height: 15px  
+  margin-top: -2px
+  display: inline-block
--- a/src/components/images/avatar.styl
+++ b/src/components/images/avatar.styl
@@ -11,5 +11,6 @@ img.avatar
 img.tiny
   vertical-align: bottom
   width: 16px
-  height: 16px
-  display: inline-block
+  hx
+  min-width: 15px
+  margin-top: -2pxisplay: inline-block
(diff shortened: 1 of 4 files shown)


whimsicallyson:
I notice that extra space is handled differently in different parts of the search results, do we want to standardize that? (might be out of the scope of this PR?)
Screen Shot 2019-09-13 at 3 21 09 PM
vs
Screen Shot 2019-09-13 at 3 21 17 PM

Also I really want the team avatar to be centered for multi-line buttons, but I'm having trouble making that happen properly
Screen Shot 2019-09-13 at 3 33 04 PM
.

ttseng:
@whimsicallyson thanks for the review!

because the row height is driven by the tallest card in the row, the spacing might look a little different between types, but the row gap is actually the same.

For the avatar vertical alignment, I'm going to keep it top aligned for consistency (for example, project and collection avatars are always top aligned even if their descriptions expand multiple lines, such as in the add-project-to-collection-pop or add-team-project-pop)

PR #911, opened 2019-09-13
Fallacious slice: Remove the global stylesheet

GregWeil:
## Links

## GIF/Screenshots:
It should look exactly the same as glitch.com, but maybe build a little faster

## Changes:
- Removed everything in the styles/ folder, and one of the webpack entry points with it
- Used shared components in place of a few things that still relied on global styles
- Moved the c3 styles into the team analytics code so it's at least located in the right place
- Moved the remaining global styles into the css modules globals

## How To Test:
- Try browsing around the site
- Check out team analytics


whimsicallyson:
Code looks good, while clicking around I did notice that the add-to-collection pop has some real small text
Screen Shot 2019-09-16 at 10 28 12 AM

PR #917, opened 2019-09-17
Add Optimizely rollouts in order to have the ability to turn off team analytics

Johnicholas:
Computing the teams analytics (the periodic writing, not the reading) is a heavyweight computation, that has been taking down machines in the ElasticSearch cluster periodically. 

In order to create a pause in firefighting to give space for designing something better, I'd like to have the option (not necessarily exercised) to turn off the Teams Analytics feature, possibly with a nice "Analytics are not available right now" message

## Links
* glitch.com/~oval-asp is a ~community remix that has this change
* glitch.com/~guttural-swordfish is an illustration of using Optimizely (with trello/Scientist, which is not necessary), in isolation.

## GIF/Screenshots:
Screen Shot 2019-09-18 at 3 40 16 PM

Screen Shot 2019-09-17 at 5 46 24 PM

## Changes:
* This adds a library, optimizely-sdk. There are a lot of similar libraries (split.io, rollout.io, launchdarkly.com), but the crucial point is that after a dev (such as me) put a feature behind a toggle and deploy it, a product owner (such as Rachel or Yasi) can rollout the feature:
  1. in development only
  2. in production
  3. to only some of the users
  4. to all of the users
and reverse any or all of those rollouts if and when they want to, based on direct feedback from customers.
* This adds a javascript file, feature-toggles.js. The intent is that it offers a single method, isFeatureEnabled(whichFeature, forWhichEntity), but in practice it publishes the whole Optimizely Client instance.
* This uses the feature-toggles.js in team-analytics, to allow us (Glitch) to turn off traffic from the teams page to the analytics api endpoint (and instead put placeholder data).

## How To Test:
* log in, go to a teams page, scroll to the bottom and wait a short or long time for the analytics to show up
* toggle the development feature "off" or "on" on the optimizely dashboard
* go back to the teams page, and wait a long or short time for the analytics to show up, and this time they are real if they were fake before, or fake if they were real before

## Feedback I'm looking for:
* What should we show if and when we turn off analytics?
* Can we commit to using something (anything, I am not particular regarding Optimizely, it's just what I happened to get working) that will support feature rollouts? We could develop our own, but I think it's pretty crucial that we can roll out (or roll back) features remotely, without needing to do a software deploy.
* Where should the feature-toggles file actually go, and how should changes to a component like this be tested?

- [ ] ...I don't know?


GregWeil:
I think showing the sample data and an 'analytics are not available right now' message in the same style as the 'add a project to the team to see analytics' message is alright.

keithk:
We just rolled our own a/b testing framework, which is also something that optimizely does, so if we decide to go down this path perhaps we'd switch to that as well.

Opening this up to the team for broader discussion about this being the way that we'd want to handle feature flags going forward

taravancil:
Strong second ➕ 
> I think it's pretty crucial that we can roll out (or roll back) features remotely, without needing to do a software deploy

GregWeil:
I think the best place to keep this would be in src/state, exporting a useFeatureEnabled hook that takes a name and entity id. That said, the most common use would probably be based on the current user id, which is not available during server side rendering without either hurting caching or some amount of pop in/out on page load.

src/components/team-analytics/index.js
```diff
@@ -43,7 +45,9 @@ function getSampleAnalytics() {
 }
 
 const useAnalyticsData = createAPIHook(async (api, { id, projects, fromDate, currentProjectDomain }) => {
-  if (!projects.length) return getSampleAnalytics();
+  if (!useRolloutToggle('analytics', String(id)) || projects.length === 0) {
```
GregWeil: Technically if this is a hook (name starts with use) it can't be used in async functions, but useRolloutToggle isn't doing anything that necessitates it being a hook. Could you either rename it back to not starting with use, or move its usage out of here?

Johnicholas: OK, I think I have done it. I didn't know that "hook" meant "function with name that starts with use" in React!

keithk:
Today in a meeting we decided:
- Go ahead with using Optimizely Rollout to handle this and future feature flags
- At some point have @GregWeil move our current A/B testing to use optimizely as well

Separately, Yasi, Tiff and @Johnicholas are going to work together to come up with a product plan for team analytics in general, and making them more useful (and designed)

keithk:
@Johnicholas Is there anything else you need from community to support you here?

Johnicholas:
I tried to get this merged last week before code freeze on friday, but I didn't quite make it. I went on 
vacation, and I will be back (working from Cork) tuesday. If you would like it merged, I think you could do it without me, or you could wait.

PR #922, opened 2019-09-20
Nova chance: Server side render collection/team/user pages

GregWeil:
## Links
https://github.com/FogCreek/Glitch-Community-Work/issues/264

## Changes:
- The server loads the full data for user, team, and collection pages so it can render them
- Removed all uses of `team.adminIds` and refactored them to use `team.teamPermissions`
- Entity editors watch for initial data changes and update state to match them

## How To Test:
- Try refreshing the site on different pages (pages that exist, pages that don't, while signed in, etc)
- Try joining/leaving teams and assigning/removing admin status


src/state/team.js
```diff
@@ -36,6 +36,10 @@ export function useTeamEditor(initialTeam) {
   } = useAPIHandlers();
   const [team, setTeam] = useState(initialTeam);
 
+  useEffect(() => {
```
clottman: Why does this need to be in an effect?

GregWeil: I only want to setState when the initialTeam changes, not every render which is more likely caused by changes to the team in state.

Wouldn't calling setState in the render function/hook cause an infinite loop?

clottman: From [the hooks api reference](https://reactjs.org/docs/hooks-reference.html#usestate)

> If you update a State Hook to the same value as the current state, React will bail out without rendering the children or firing effects. (React uses the Object.is comparison algorithm.)
> 
> Note that React may still need to render that specific component again before bailing out. That shouldn’t be a concern because React won’t unnecessarily go “deeper” into the tree. If you’re doing expensive calculations while rendering, you can optimize them with useMemo.

PR #926, opened 2019-09-23
Functional politician: Replacing Default Collection Avatars

ttseng:
## Links
* Remix link: https://functional-politician.glitch.me/@scientiffic/lists
* Issue-tracker link: https://app.zenhub.com/workspaces/community-5d07d1359dee6059a9688176/issues/fogcreek/glitch-community-work/426
* Specs/doc links: https://docs.google.com/document/d/12elIoMtCI-si1rbzYs6khnNcxZvNWG79HlmR8Vol1A0/edit#

## GIF/Screenshots:
avatar-generation
Generated avatars appear on collection pages and are created based on the collection color, its complementary color, and the avatars of the first 1-3 projects in the collection.

If the complementary color has low contrast with the collection color, I pick either white or black as the contrast color:
contrast

Avatars are removed elsewhere on collection cards:
image
More by

image
Profile Page

image
Search Results

For Search Autocomplete, a default avatar is used composed of the collection coverColor and the stacked border
image

I also improved the info-hierarchy of collection cards on the homepage so that the name stick out more than the # of projects in the collection:
image

## How To Test:
Viewing collections - check that collection cards match the spec on these pages:
- user profile pages (https://functional-politician.glitch.me/@scientiffic/)
- project pages included in collections sections (https://functional-politician.glitch.me/~microbit-ble)
- collection pages (https://functional-politician.glitch.me/@scientiffic/microbit)
- search results (https://functional-politician.glitch.me/search?q=microbit&activeFilter=collection)

Editing Collections
- Sign into Functional Politician
- Go to one of your collection pages and try editing the color and see how the avatar changes
- Try adding / removing projects from a collection and check how the collection avatar changes

## Feedback I'm looking for:
- Did I break something?
- Are there ways I should be handling the collection SVG to ensure it's accessible?
- I haven't deleted adding the default avatar to new collection items because I'm scared it will break something.


src/components/collection/collection-avatar.js
```diff
@@ -0,0 +1,86 @@
...
+};
+
+const getComplementaryColor = (inputColor) => {
+  const color = new CompColors(inputColor);
+
+  // CompColors format: {r: 255, g: 255, b: 255}
+  const originalColor = color.primary()[0];
+  const complementaryColor = color.complementary()[1];
+
+  const originalColorHex = rgbToHex(originalColor.r, originalColor.g, originalColor.b);
```
clottman: Do we need these in hex? Any reason we can't use the `rgb` method of our contrast checker directly and save some computation?

src/components/collection/collection-avatar.js
```diff
@@ -0,0 +1,86 @@
...
+      styles.avatarContainer,
+      collection.projects.length === 0 && styles.empty,
+      collection.projects.length >= 3 && styles.stacked,
+      collection.projects.length > 0 && collection.projects.length < 3 && styles.centered,
+    )}
+    style={{ backgroundColor: getComplementaryColor(collection.coverColor) }}
+  >
+    {getPattern(collection.id, collection.coverColor)}
+
+    {collection.projects
```
clottman: Will `collection.projects` always be sorted in reverse chronological order? (Are we guaranteeing that somehow or relying on the order returned by the database to always be the same?)

src/components/collection/collection-item.styl
```diff
@@ -95,20 +133,53 @@
...
 .smallCollectionLink
+  width: 100%
+  height: 100%
+  position: relative
   color: inherit
   text-decoration: none
   display: flex
   flex-direction: column
   flex: 1 1 100%
+  border-color: var(--border-color)
```
clottman: Probably worth adding a comment here explaining how we expect these to be set

src/components/collection/collection-item.styl
```diff
@@ -95,20 +133,53 @@
...
     
 .bubbleContainer
   flex: 1 0 auto
   position: relative
   border-radius: 5px 5px 0 0
   word-break: break-word
   padding: 1rem 1rem  
+  max-height: 140px
+  z-index: 1
+  border-color: var(--border-color)
```
clottman: Same here re: adding a comment

src/components/collection/collection-patterns.js
```diff
@@ -0,0 +1,68 @@
+import React from 'react';
+
+export const Waves = ({ color }) => (
+  <svg viewBox="0 0 87 258" version="1.1" aria-label="">
+    <g stroke="none" strokeWidth="1" fill="none" fillRule="evenodd" strokeLinejoin="round">
+      <g transform="translate(-23.000000, -27.000000)" stroke={color} strokeWidth="13">
+        <g transform="translate(30.000000, 27.000000)">
+          <path
+            d="M38.8272916,257.292177 L38.8272916,254.693978 C40.4435881,247.40588 41.2517363,240.110398 41.2517363,232.807532 C41.2517363,221.853234 31.2018277,201.797488 31.2018277,189.252818 C31.2018277,176.708147 41.2517363,163.946407 41.2517363,149.293053 C41.2517363,134.6397 31.2018277,117.591928 31.2018277,103.359425 C31.2018277,89.9987434 41.2517363,78.5214872 41.2517363,66.1617696 C41.2517363,53.802052 31.5331422,28.7962359 31.2018277,20.7121198 C30.9809514,15.322709 31.7819491,9.93065296 33.6048209,4.53595169 L33.6048209,0.557825971"
+            id="Line-39"
```
clottman: Just in case we do decide to reuse these at some point, we should remove the `id` elements in case these are ever used multiple times on the same page. (Duplicate IDs can be a problem for assistive tech)

src/components/collection/container.js
```diff
@@ -54,10 +55,10 @@ const CollectionContainer = ({ collection, showFeaturedProject, isAuthorized, pr
   let avatar;
   if (myStuffIsEnabled && collection.isMyStuff) {
     avatar = <BookmarkAvatar width="50%" />;
-  } else if (collection.avatarUrl) {
-    avatar = <Image src={collection.avatarUrl} alt="" />;
-  } else {
+  } else if (collection.projects.length > 0) {
     avatar = <CollectionAvatar collection={collection} />;
+  } else {
+    avatar = '';
```
clottman: You could get rid of this `else` by initializing on line 55 
`let avatar = ''` (or `= null`)

src/components/collection/collection-avatar.js
ttseng: great point!

src/components/collection/collection-avatar.js
ttseng: yeah, it's always sorted by `projectOrder` (the sorted order of projects in the collection)

ttseng:
good catch with needing to filter out the currently displayed collection from the More Collections section – it should be fixed now.

I also did a slightly update to the collection avatar in search autocomplete results:  
avatar-search-count

(those changes are in `collection-avatar.js`)

src/presenters/pages/home-v2/styles.styl
```diff
@@ -265,6 +265,18 @@
...
+    .curatedCollectionButtonWrap
+        box-shadow: 2px 2px 0 #222
+        border-radius: 5px
+        width: fit-content
+        height: fit-content
+  .collectionLink
+    font-size: 14px
+    font-weight: 600
+    color: primary
+    &:hover
```
clottman: Do we also want `:focus` here and on the above `:hover` state?

src/utils/color.js
```diff
@@ -26,6 +26,17 @@ export const hexToRgbA = (hex) => {
   return false;
 };
 
+// from https://stackoverflow.com/a/5624139/1720985
+const componentToHex = (c) => {
```
clottman: We can still get rid of these and use `rgb` from wcag-contrast, maybe by extending getContrastWithLightText to work with either a hex or rgb value?

ttseng: I think having functions to convert hex->rgb and rgb->hex will be useful in other cases, other than determining contrast.

PR #929, opened 2019-09-24
Romantic broom: fixed ratio containers for large videos on create page

modernserf:
## Links
* Remix link: https://romantic-broom.glitch.me/
* Issue-tracker link: https://app.zenhub.com/workspaces/community-5d07d1359dee6059a9688176/issues/fogcreek/glitch-community-work/158

## Changes:
* wrapped large videos in a container that reserves space while video is loading, to prevent layout jumps

## How To Test:
* visit https://romantic-broom.glitch.me/create with network throttled to 'low-end mobile'
* there should be empty containers where the large videos go
* when the videos load, the layout should not jump around

## Feedback I'm looking for:
* should we do anything for smaller screens? it seems like those have fixed heights instead of aspect ratios

### DIFF
--- a/src/presenters/pages/create.js
+++ b/src/presenters/pages/create.js
@@ -1,6 +1,7 @@
 import React, { useEffect, useState } from 'react';
 import { Tab, Tabs, TabList, TabPanel } from 'react-tabs';
 import classNames from 'classnames/bind';
+import styled from 'styled-components';
 import { Button, Icon, Loader, Mark } from '@fogcreek/shared-components';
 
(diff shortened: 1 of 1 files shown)


ttseng:
I think there was an unintended consequence where the spacing underneath the video has been reduced; the video now overlaps with the heading underneath it, which makes it look like the heading is a caption (though it's not related to the video):

image
(romantic-broom on left / prod on right)

taravancil:
Besides the issue Tiff highlighted, this looks good to me!

PR #473, opened 2019-04-25
Superb guide: Get rid of old input presenters 

GregWeil:
## Links
https://superb-guide.glitch.me/storybook/index.html?path=/story/input-fields--errors
https://glitch.fogbugz.com/f/cases/3328395/Add-InputErrorIcon-to-storybook
https://glitch.fogbugz.com/f/cases/3328394/Add-InputErrorMessage-to-storybook
https://glitch.fogbugz.com/f/cases/3328356/Build-color-picker-component

## GIF/Screenshots:
image

## Changes:
- Created a CollectionNameInput component to replace EditCollectionNameAndUrl
- Removed presenters/field-helpers.js now that the above is gone
- Added stories for InputErrorMessage and InputErrorIcon
- Refactored the collection color picker a little bit

## How To Test:
Try out the different input stories, try changing the color of a collection

PR #930, opened 2019-09-24
spectacular-coyote: fix onboarding banner logic on profile pages

taravancil:
## Links
* [spectacular-coyote.glitch.me](https://spectacular-coyote.glitch.me/~cowards)
* [#473](https://github.com/FogCreek/Glitch-Community-Work/issues/473)

## GIF/Screenshots:
A user with one pinned project does not see the onboarding banner on their profile

Screen Shot 2019-09-24 at 4 39 59 PM

## Changes:
Check if the *current user* has any projects, not if the profile we're visiting has any *recent projects*

## How To Test:
- Log in with an account that has one project
- Pin that project, refresh. You should not see the onboarding banner on your profile
- Delete that project, refresh. You *should* see the onboarding banner on your profile

## Feedback I'm looking for:
Work as expected?

### DIFF
--- a/src/presenters/pages/user.js
+++ b/src/presenters/pages/user.js
@@ -92,7 +92,6 @@ const UserPage = ({ user: initialUser }) => {
   const { currentUser: maybeCurrentUser } = useCurrentUser();
   const isSupport = maybeCurrentUser && maybeCurrentUser.isSupport;
   const isAuthorized = maybeCurrentUser && maybeCurrentUser.id === user.id;
-  console.log(maybeCurrentUser);
 
   const pinnedSet = new Set(user.pinnedProjects.map(({ id }) => id));
   // filter featuredProject out of both pinned & recent projects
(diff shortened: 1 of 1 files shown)


src/presenters/pages/user.js
```diff
@@ -128,7 +128,7 @@ const UserPage = ({ user: initialUser }) => {
           />
         </UserProfileContainer>
 
-        {isAuthorized && !recentProjects.length && <OnboardingBanner />}
+        {isAuthorized && !maybeCurrentUser.projects.length && <OnboardingBanner />}
```
GregWeil: If this uses `user.projects` it will update right away rather than requiring a refresh

ttseng:
cool, I no longer see the onboarding banner!

src/presenters/pages/user.js
taravancil: Cool, thanks!

PR #931, opened 2019-09-24
Imminent twill: re-componentize add collection project pop

modernserf:
## Links
* Remix link: https://imminent-twill.glitch.me/@modernserf/linen-collection
* Issue-tracker link: https://app.zenhub.com/workspaces/community-5d07d1359dee6059a9688176/issues/fogcreek/glitch-community-work/151

## GIF/Screenshots:
Screen Shot 2019-09-24 at 5 36 31 PM

## Changes:
* rebuild PopoverSearch and ProjectResultItem with shared components
* remove 'View' link and user/team profile links from result item -- this makes the keyboard navigation vastly simpler. If people complain we can add them back

## How To Test:
* log into https://imminent-twill.glitch.me
* go to one of your collections
* add a project with the "Add Project" button
* try navigating with the keyboard


ttseng:
A few UX things I'm noticing:
- I'm not seeing any hover styling appear when I hover over result items.
Sep-25-2019 13-15-32
- I can tab to the first result, but when I hit tab again, it goes to the project items within the collection:
tab
- The arrow keys aren't working for me for going through items in the list
- This is a bug that I'm noticing on production (so can probably be addressed in a separate PR but noting it here) - if you enter in a URL, it should only return the exact result that matches that url, rather than a list of results based on name:
url-search

- I feel like the view link here is the most important of all the places we show it because it's a Glitch-wide search (rather than searching your own projects or collections, where you probably know what it is you're looking at).  I know that I personally use the View link frequently.

modernserf:
@clottman what 'add to new collection' button? this is just the popover on the collections page itself, not the popover on each project

clottman:
@modernserf you might make that clear in the description; the issue you linked refers to all the filter lists.

clottman:
I think for consistency's sake we should probably deploy keyboard behavioral updates to all the variations of adding a project to a collection at once.

PR #932, opened 2019-09-25
Superb guide: SSR all pages, move seo into react code

GregWeil:
## Links
https://github.com/FogCreek/Glitch-Community-Work/issues/214

## Changes:
- Helmet data is used when rendering on the server, so seo fields are set by the page
- All pages render ssr now, which changes /questions, /search, and category pages (they don't data load so they just appear as loaders)
- Created some functions for standalone markdown rendering in utils/,  used for seo descriptions
- Pages had a canonical url of glitch.com/ if it wasn't otherwise specified. I changed it to only include the url tag if it is defined, so that we don't provide bad info

## How To Test:
- Try doing a full load on each page and make sure the seo content is the same as on glitch.com
- Make sure seo images are png and not the svgs used on the page

## Feedback I'm looking for:
- Should anything special be done for search/categories, or is it ok to leave them as loaders?

Merge nova-chance, this pr is based on that branch so it includes all of the changes from it


sarahzinger:
hey @GregWeil I'm somewhat new to this, how would I go about doing this?
> and make sure the seo content is the same as on glitch.com

GregWeil:
I've just been peeking at the `<meta>` tags in the `<head>`, but https://www.heymeta.com/url/superb-guide.glitch.me/~superb-guide seems like a good option too

src/components/glitch-helmet/index.js
```diff
@@ -0,0 +1,43 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Helmet } from 'react-helmet-async';
+import { APP_URL } from 'Utils/constants';
+
+const GlitchHelmet = ({ title, description, image, socialTitle, canonicalUrl }) => {
```
sarahzinger: is helmet a well known term you think? i wonder if it's worth putting a comment explaining what this is and what it's used for, of if everyone already gets it?

src/presenters/pages/secret.js
```diff
@@ -66,10 +66,12 @@ const Secret = () => {
     setEnabledToggles(newToggles);
   };
 
+  const tagline = "It's a secret to everybody.";
+
   return (
     <main className={styles.secretPage}>
-      <Helmet title="Glitch - It's a secret to everybody." />
-      <VisuallyHidden as={Heading} tagName="h1">It's a secret to everybody</VisuallyHidden>
+      <GlitchHelmet title={`Glitch - ${tagline}`} description={tagline} />
```
sarahzinger: do we want to have seo on our secret page? seems unnecessary right?

GregWeil:
The community server has a 30 minute cache on the api responses used for seo (which applies to ssr too, but the client then pulls the latest data and updates the page)

src/presenters/pages/secret.js
taravancil: Unecessary, but I think cool to have in case people share the page on Twitter like "hey I found this easter egg". It's always nice to have the metadata so Twitter or whatever can render a full card instead of just a link

PR #947, opened 2019-10-01
Nova chance: Preload optimizely config, track optimizely in react

GregWeil:
## Links
https://github.com/FogCreek/Glitch-Community-Work/issues/481

## Changes:
- The server maintains an optimizely instance it uses for server rendering
- The client inits optimizely using the latest data loaded by the server instance
- Renamed state/rollout-toggles to state/optimizely, now exports a useFeatureEnabled hook
- Dried out utils/constants just a little bit

## How To Test:
* Make sure there aren't any errors on team pages, if you're a member or not, and on initial load


GregWeil:
Updated to add listeners for optimizely config changes

shared/constants.js
```diff
@@ -8,6 +8,7 @@ const envs = {
     CDN_URL: 'https://cdn.glitch.com',
     GITHUB_CLIENT_ID: 'b4cb743ed07e20abf0b2',
     FACEBOOK_CLIENT_ID: '660180164153542',
+    OPTIMIZELY_KEY: 'SUV4dn9FiCp1osqWQUZr82',
```
Johnicholas: Could you use this instead? GZMqH7Aou8QmcPV2hm16wj

The key that starts "SUV" is under my control, and I'd rather not be the person who has the power to change the behavior.

server/render.js
```diff
@@ -3,6 +3,7 @@ const { performance } = require('perf_hooks');
 const dayjs = require('dayjs');
 const { captureException } = require('@sentry/node');
 const createCache = require('./cache');
+const { getOptimizelyClient, getOptimizelyData } = require('./optimizely');
```
Johnicholas: I'd really like it if we could confine the vendor name "optimizely" behind some abstraction that offers us the ability to switch vendors, or move from a vendor to a home-built implementation.

For example, could we use `await rolloutToggles.ready()` and `rolloutToggles.isFeatureEnabled(toggleName, entityName)`?

GregWeil: I changed the naming exposed in the react code, but I left the server/client setup the same. That stuff is pretty tied to how optimizely wants to be initialized.

PR #949, opened 2019-10-01
Sour maize: update recent projects when leaving a project

modernserf:
## Links
* Remix link: http://sour-maize.glitch.me
* Issue-tracker link: https://app.zenhub.com/workspaces/community-5d07d1359dee6059a9688176/issues/fogcreek/glitch-community-work/486

## Changes:
* update current user's "projects" array when leaving a project

## How To Test:
* log into http://sour-maize.glitch.me
* click on the project options menu for one of your recent projects
* leave the project
* the project should be removed from your recent projects

## Future work (not in this PR)
Update the current user's cached projects/collections/etc when any other relevant action happens

### DIFF
--- a/src/state/current-user.js
+++ b/src/state/current-user.js
@@ -188,6 +188,10 @@ export const { reducer, actions } = createSlice({
     }),
     // TODO: more granular actions for managing user's teams, collections etc
     updated: (state, { payload }) => ({ ...state, ...payload }),
+    // TODO: use the same action that's defined in the resource manager PR
+    leftProject: (state, { payload }) => {
+      state.projects = state.projects.filter((p) => p.id !== payload.id);
+    },
   },
 });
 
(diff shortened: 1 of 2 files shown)


src/state/current-user.js
```diff
@@ -188,6 +188,10 @@ export const { reducer, actions } = createSlice({
     }),
     // TODO: more granular actions for managing user's teams, collections etc
     updated: (state, { payload }) => ({ ...state, ...payload }),
+    // TODO: use the same action that's defined in the resource manager PR
+    leftProject: (state, { payload }) => {
+      state.projects = state.projects.filter((p) => p.id !== payload.id);
```
GregWeil: Shouldn't this return a new state object, not modify the old one?

modernserf: its using immer

GregWeil: Neat, I'll have to read up on that a bit more

PR #951, opened 2019-10-02
Testy-time: fixes some a11y issues on the Glitch homepage and add tests for it

clottman:
## Links
* Remix link: n/a

## GIF/Screenshots:
image
(should be same as current) 

## Changes:
* Runs axe (an accessibility testing library) on the glitch homepage when an anonymous user visits it
* Fixes the errors from axe on the homepage

## How To Test:
* Let Cypress do its thing


clottman:
@oliviacpu this has some cypress stuff testing the homepage as well as a few fixes for a11y stuff - think we should merge it for now? Or should I cherry-pick the a11y fixes that I want to keep and close this?

clottman:
Not gonna merge because we're moving away from Cypress. Created another ticket to track that the homepage a11y fixes should happen. https://app.clubhouse.io/glitch/story/3154/minor-a11y-issues-on-homepage

PR #955, opened 2019-10-03
circular-passionfruit: Remove "Resume Coding" button and add "Edit Project" to recent projects project items

taravancil:
## Links
* [circular-passionfruit.glitch.me](https://circular-passionfruit.glitch.me/)
* [#453](https://github.com/FogCreek/glitch-community-work/issues/453)
* [User dashboard spec](https://docs.google.com/document/d/10KS93TQ6vaQcVSL0ujUj4__GGl7M5MlS1vimzNqiizU/edit#heading=h.3im6lap0txtw)

## GIF/Screenshots:
Screen Shot 2019-10-03 at 11 23 55 AM

## Changes:
* Remove the "Resume Coding" button
* Add the `showEditButton` prop to `<ProjectsList>` and `<ProjectItem>`, which if true, will render a footer with an "Edit Project Button"

## How To Test:
* Log in, you should see the "Edit Project" button on your recent projects dashboard on the homepage
* The "Resume Coding" should be gone
* Check out other `<ProjectsList>`s, you should not see the "Edit Project" button on those `<ProjectItem>`s
  - Team page recent projects
  - User page recent projects

### DIFF
--- a/src/components/project/project-item.styl
+++ b/src/components/project/project-item.styl
@@ -132,6 +132,6 @@ a.linkBody:hover
   color: private-foreground
   
 .footer
-  background: secondary
+  background: secondary-background
   padding: .5rem 1rem
   border-radius: 0 0 5px 5px
--- a/src/components/project/project-item.styl
+++ b/src/components/project/project-item.styl
@@ -85,6 +85,8 @@
   &.private
     background-color: private
     border-color: private
+  &.hasFooter
+    border-radius: 5px 5px 0 0
 a.linkBody:hover
   text-decoration: none
   .itemButtonWrap
@@ -128,3 +130,7 @@ a.linkBody:hover
 .privateIcon
   margin: 4px 5px 0
   color: private-foreground
+  
+.footer
+  background: secondary
+  padding: 1rem
--- a/src/components/project/project-item.styl
+++ b/src/components/project/project-item.styl
@@ -133,4 +133,5 @@ a.linkBody:hover
   
 .footer
   background: secondary
-  padding: 1rem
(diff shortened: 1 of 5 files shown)


taravancil:
@ttseng to confirm: the "Edit Project" button should only be on project items in the user dashboard, right?

taravancil:
> I would really want this on any project I have the ability to edit, though...

Maybe! Project lists often display dozens of projects at a time - it's not clear to me if it would be odd to have the button displayed dozens of times on a page. I'll leave the decision to Tiff!

src/components/project/project-item.styl
```diff
@@ -127,4 +129,9 @@ a.linkBody:hover
     
 .privateIcon
   margin: 4px 5px 0
-  color: private-foreground
\ No newline at end of file
+  color: private-foreground
+  
+.footer
+  background: secondary
```
ttseng: do you mind changing this to `secondary-background` instead? That way the project info is the primary thing that your eye is drawn to, and the edit project appears secondary: 
image

ttseng:
let's see what the reaction is to just user dashboard changes, and then determine whether it's needed everywhere.  

having them default link to project pages everywhere else just may be the only way where project authors discover we have project pages - and that their READ.mes appear there 😂

taravancil:
Updated the footer background to `secondary-background`

Screen Shot 2019-10-03 at 3 07 01 PM

taravancil:
> is the idea to add this in advance of the rest of the dashboard work?

yep, if that's ok with you!

PR #959, opened 2019-10-04
Fix bugs for My Stuff and Private Collections, mostly a11y issues

sarahzinger:
## Links
* Remix link: https://ember-spell.glitch.me/
* Issue-tracker link:
    - https://app.zenhub.com/workspaces/community-5d07d1359dee6059a9688176/issues/fogcreek/glitch-community-work/499
    - https://app.zenhub.com/workspaces/community-5d07d1359dee6059a9688176/issues/fogcreek/glitch-community-work/500
    - https://app.zenhub.com/workspaces/community-5d07d1359dee6059a9688176/issues/fogcreek/glitch-community-work/498

## Changes:
* We decided to remove the tooltip from "My Stuff", as it was acting weird on mobile, and we're hoping the button is intuitive enough that we don't need one
* Added a label to the Collection Link items in a collections list on the user page so it would announce via screenreader when a collection is private
* fixed an issue where private collections were not showing with a yellow background in the add project to collection dropdown
* fixed an issue with our privacy toggles, where they were not outlining when they were focused
* added some missing alt tags to our privacy icons
* added an aria label to the privacy toggle to match the behavior we have on the my stuff-bookmark buttons (the label stays the same, and voice over will clarify if it's checked or unchecked, rather than alternating the help text if it's checked or unchecked)

## How To Test:
* enable My Stuff
* ensure you can use the My Stuff Buttons and the Privacy Toggle buttons well, including while using voice over

## Feedback I'm looking for:
anything else we might have missed? I've done a bunch of manual testing and very quickly ran through the Accessibility Insights for Web extension, but I'm sure I've missed things.

### DIFF
--- a/src/components/private-badge/styles.styl
+++ b/src/components/private-badge/styles.styl
@@ -32,3 +32,7 @@
   color: private-foreground
   svg
     opacity: 0.7
+
+.focused
+  outline-color: -webkit-focus-ring-color;
+  outline-style: auto;
--- a/src/components/private-badge/styles.styl
+++ b/src/components/private-badge/styles.styl
@@ -34,5 +34,5 @@
     opacity: 0.7
 
 .focused
-  outline-color: -webkit-focus-ring-color;
-  outline-style: auto;
+  outline: 1px dotted #212121
+  outline: 5px auto -webkit-focus-ring-color
--- a/src/components/buttons/bookmark-button.js
+++ b/src/components/buttons/bookmark-button.js
@@ -3,7 +3,6 @@ import PropTypes from 'prop-types';
 import Image from 'Components/images/image';
 import classNames from 'classnames/bind';
 import { CDN_URL } from 'Utils/constants';
-import TooltipContainer from 'Components/tooltips/tooltip-container';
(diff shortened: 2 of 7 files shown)


src/components/private-badge/styles.styl
```diff
@@ -31,4 +31,8 @@
   background-color: secondary-background
   color: private-foreground
   svg
-    opacity: 0.7
\ No newline at end of file
+    opacity: 0.7
+
+.focused
+  outline-color: -webkit-focus-ring-color;
```
GregWeil: This value doesn't work on firefox, could you switch it with the literal color value

sarahzinger: @GregWeil want to take a look again? I think it should now match whatever the default focus state is for each browser

sarahzinger:
hmm just found some edge bugs i think i'll fix that in this pr, moving this to Ready for dev instead

sarahzinger:
ok fixed the bugs in edge and have node_env set to production so you should be able to see it on edge hopefully

PR #964, opened 2019-10-08
Attempt to test components using Jest

clottman:
- Run this using `npm run jesttest`. Make sure you've run `npm install` beforehand. 

- See it fail because regenerator runtime is not defined, even though it is

- Go back to mocha? Maybe you can see what's wrong and want to pick it up from here? 

This PR was based off the mocha attempt and so has some mocha-specific stuff in here we'd need to clean up if we got it working


clottman:
Closed; sticking with Mocha

PR #969, opened 2019-10-09
Component tests: Test the Profile List component

clottman:
## Summary

tests/ProjectsList/index.js has some tests! 

The tests: 
 - check if there are any accessibility errors when there are no projects given to ProjectsList
 - renders ProjectsList with some projects, and checks:
   - are there any accessibility errors?
   - does it show a project that was passed in? 
   - if you type in the filter, do only projects matching the filter remain? 

See TESTING.md changes for info on running these on your local machine.


test/FilterController/index.test.js
```diff
@@ -0,0 +1,89 @@
...
+  });
+
+  it('should show projects', function() {
+    const projects = [makeTestProject({ id: '1' }), makeTestProject({ id: '2' }), makeTestProject({ id: '3' })];
+    const reduxStore = mockStore({ currentUser: makeTestUser({ projects, teams: [] }) });
+    const errorStub = sinon.stub(ErrorHandlers, 'default');
+    errorStub.returns({ handleError: () => {}, handleCustomError: () => {} });
+
+    const visibilityContainerStub = sinon.stub(VisibilityContainer, 'default');
+    visibilityContainerStub.returns(FakeVisibilityContainer);
```
Osmose: The warning is caused by this. The stub is returning a reference to the class when called, so if we wrote `<VisbilityContainer />` it would evaluate not to a react component, but to the class itself. And, since classes are functions, it triggers the function value warning.

Instead, we should be using `callsFake`:

```suggestion
    visibilityContainerStub.callsFake((props) => new FakeVisibilityContainer(props));
```

PR #972, opened 2019-10-10
Occipital point - /login page forget password flow

sarahzinger:
## Links
* Remix link: https://occipital-point.glitch.me/

## Changes:
* Adds a sign in with password and forgot password workflow for /login
* a full list of changes and screenshots can be found here: https://github.com/FogCreek/Glitch-Community/pull/934
* basically we tried deploying this before but had to revert it because after we implemented shared components, this code did not work anymore :) 

## How To Test:
* enable user passwords
* make sure you have a password under account settings
* trying signing in/resetting your password on /login

## Feedback I'm looking for:
when i used shared components I started to get an error about us using this useEmail hook in our onChange event, so i reworked how we store/update email state, but i feel like it looks ugly and is awkward and there's a better way to do it. Would love ideas on this!

nope


src/components/sign-in/forgot-password.js
```diff
@@ -0,0 +1,109 @@
...
+    errorMessage: null,
+    isFocused: true,
+  });
+  const { email, status, errorMessage, isFocused } = state;
+  const [emailInHook, setEmail, validationError] = useEmail();
+  useEffect(() => {
+    if (email !== emailInHook) {
+      setEmail(email);
+    }
+  }, [email]);
```
sarahzinger: originally we kept email out of the state object and just used this `useEmail` hook directly but we're getting console.log errors about it `Warning: State updates from the useState() and useReducer() Hooks don't support the second callback argument. To execute a side effect after rendering, declare it in the component body with useEffect().` i see them elsewhere so we probably want to think about the best way to handle this, but that can probably happen in another pr maybe

PR #976, opened 2019-10-11
Ancient salmonberry: Convert Categories to Collections

sarahzinger:
UPDATE! this pr has changed a bunch since I wrote this description. This should now be a 1:1 replacement of categories to collections, and should hopefully not be noticeable to users

## Links
* Remix link: https://ancient-salmonberry.glitch.me/
* Issue-tracker link:
    - https://app.zenhub.com/workspaces/community-5d07d1359dee6059a9688176/issues/fogcreek/glitch-community-work/528
    - https://app.zenhub.com/workspaces/community-5d07d1359dee6059a9688176/issues/fogcreek/glitch-community-work/525
* Specs/doc links: https://docs.google.com/document/d/127bu23-WBPQmkFWMuFCYyc43wvtoYh_EokvLtRT3uQY/edit#

## GIF/Screenshots:
Screen Shot 2019-10-11 at 5 15 01 PM
Screen Shot 2019-10-11 at 5 15 26 PM

## Changes:
* All previous categories pages (ex: glitch.com/music) now use **collection** data instead of category data so that they are easier to manage by dev advocates. This will also fix an existing bug with Categories and My Stuff and help prevent future bugs with the new exploring collections project.
* Combines Art and Music into one category (since there is now going to be one collection for the two of them)
* Removes Learn to Code from the list

## How To Test:
* go to https://ancient-salmonberry.glitch.me/secret and enable the onboarding banners
* go to https://ancient-salmonberry.glitch.me/ and sign in with a new account that does not have any projects
* notice on the homepage that the onboarding banner has the new list of categories, make sure they all go to reasonable locations, try resizing the window to make sure it looks good
* notice on the user page that the onboarding banner has a new list of categories, make sure they also go to reasonable locations, try resizing the window to make sure it looks good
* go to the /create page and scroll to the bottom to the categories list and make sure they all go to reasonable locations, try resizing the window to make sure it looks good
* ensure that all old categories links go to an actively managed collection and it's not a dead link in case we have old tweets or youtubes referring to them:
    - https://ancient-salmonberry.glitch.me/hello-worlds
    - https://ancient-salmonberry.glitch.me/games
    - https://ancient-salmonberry.glitch.me/building-blocks
    - https://ancient-salmonberry.glitch.me/learn-to-code
    - https://ancient-salmonberry.glitch.me/tools-for-work
    - https://ancient-salmonberry.glitch.me/community-picks
    - https://ancient-salmonberry.glitch.me/handy-bots
[...]

PR #979, opened 2019-10-15
Gelatinous bestseller: 301 redirect for 

modernserf:
## Links
* Remix link: https://gelatinous-bestseller.glitch.me
* Issue-tracker link: https://app.zenhub.com/workspaces/community-5d07d1359dee6059a9688176/issues/fogcreek/glitch-community-work/537

## Changes:
* redirect legacy root team URLs to '@' URLs (eg. glitch.com/slack => glitch.com/@slack)

## How To Test:
* visit https://gelatinous-bestseller.glitch.me/twitter
* you should be redirected to https://gelatinous-bestseller.glitch.me/@twitter

## Feedback I'm looking for:
* Is the client-side routing at all necessary?


src/presenters/pages/router.js
```diff
@@ -16,7 +16,7 @@ import OauthSignIn from './signin';
 import JoinTeamPage from './join-team';
 import QuestionsPage from './questions';
 import ProjectPage from './project';
-import { TeamPage, UserPage, TeamOrUserPage } from './team-or-user';
```
GregWeil: `TeamPage` is dead code now

server/routes.js
```diff
@@ -170,6 +171,13 @@ module.exports = function(external) {
     await render(req, res);
   });
 
+  // redirect legacy root team URLs to '@' URLs (eg. glitch.com/slack => glitch.com/@slack)
+  Object.keys(rootTeams).forEach((teamName) => {
+    app.get(`/${teamName}`, (req, res) => {
+      res.redirect(301, `/@${teamName}`);
```
sarahzinger: as you mentioned redirects in this pr: https://github.com/FogCreek/Glitch-Community/pull/976 I was curious did you do this on the server because of SSR or for SEO reasons or both?

modernserf: both, I think? I don't think we were doing SSR on these routes before, and I'm not sure if we were SEO-ing these correctly before (e.g. using `<link rel="canonical">`)

PR #980, opened 2019-10-15
Sideways indigo: fix missing currentUser.projects bug

modernserf:
## Links
* Remix link: https://sideways-indigo.glitch.me/
* Issue-tracker link: https://app.zenhub.com/workspaces/community-5d07d1359dee6059a9688176/issues/fogcreek/glitch-community-work/516

## Changes:
From the Sentry error, it looks like this problem happens when there's an error loading a shared user -- this would result in the shared & cached users being nulled out, instead of replaced with a new anonymous user. I think it would go something like this:

- shared user is loaded from storage
- shared user exists, so no anonymous user is loaded
- cached user is requested
- cached user returns error, because sharedUser is bad
- getSharedUser is called, but API returns a 401 error,  and so returns `undefined`
- getCachedUser is called again on a null user, and returns `undefined` (instead of an error)
- the users are cleared from localstorage
- the loadedFresh action is dispatched with `undefined` as the value, which overwrites the currentUser state
- currentUser no longer has a projects field, and the app crashes

This PR basically removes every opportunity for the current user to be nulled, whether that's from the situation i'm describing above, or that somehow the stored shared user is missing a persistent token.

## How To Test:
* no idea


GregWeil:
I think `currentUser` isn't undefined, the projects field is. I can recreate the error in this remix with `localStorage['community-cachedUser'] = JSON.stringify({});`.

I don't know how someone actually gets in that state, but it might be enough to fall back on the default user fields when loading in from local storage.

modernserf:
this is why i think currentUser is being deleted: `Fixed shared cachedUser from 5180178 to null` https://sentry.io/organizations/fog-creek-software/issues/1227033335/?project=1246508&referrer=github_integration

GregWeil:
Ok, another way to force a repro is to load the site signed in, then edit `localStorage.cachedUser` with an invalid token and refresh the homepage. glitch.com gets a `currentUser.projects` error, this remix doesn't break the whole site but logs a `sharedUser is null` error and doesn't generate an anon user or do anything until you refresh again.

modernserf:
@GregWeil is this still relevant? I saw you had a PR open for some currentUser stuff recently

PR #984, opened 2019-10-15
Fix tests & add more & a11y fix

clottman:
## Links
* Remix link
* https://github.com/FogCreek/Glitch-Community-Work/issues/542

## GIF/Screenshots:
no visual changes

## Changes:
* Adds a package called mocha-jsdom which does jsdom setup + teardown 
  * This prevents a memory leak which can lead to a timeout of mocha
* Adds a unit test for the PrivateToggle component
* Adds a role to the children of hidden checkbox for greater assistive technology compatability
* makes BookmarkButton's child a span because labels can't semantically contain divs

## How To Test:
* Make sure the bookmark button looks/acts the same

### DIFF
--- a/package.json
+++ b/package.json
@@ -108,7 +108,6 @@
     "markdown-it-sanitizer": "^0.4.3",
     "memory-cache": "^0.2.0",
     "mini-css-extract-plugin": "^0.8.0",
-    "mocha-jsdom": "^2.0.0",
     "module-alias": "^2.2.0",
     "nodemon": "^1.19.3",
     "prop-types": "^15.7.2",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -108,6 +108,7 @@
     "markdown-it-sanitizer": "^0.4.3",
     "memory-cache": "^0.2.0",
     "mini-css-extract-plugin": "^0.8.0",
+    "mocha-jsdom": "^2.0.0",
```
keithk: Hmm we should clean up our packages.json and move things like mocha/etc to dev dependencies -- not for this PR though

clottman: Oh good catch, all the other mocha stuff is in dev dependencies

PR #985, opened 2019-10-15
Nova chance: Disable cloudfront caching of html

GregWeil:
## Links

## Changes:
- Changed `Cache-Control` to `no-cache` for all community pages (does not include proxied pages or js)
- Fixed a typo in the js cache time, doesn't actually change anything because the default was milliseconds

## How To Test:
- Load up the site, the html should have a `no-cache` header. JS and proxied pages should still be cached
- Pull up glitch.com and see how often you get `Hit from Cloudfront` in the response headers, it seems very rare for me (make sure the dev tools aren't disabling the browser cache)

This is a test for ssr optimizely, so I'll have to be ready to roll back in case this breaks the site

### DIFF
--- a/server/routes.js
+++ b/server/routes.js
@@ -39,7 +39,7 @@ module.exports = function(external) {
   initWebpack(app);
   const buildTime = dayjs();
 
-  const ms = dayjs.convert(7, 'days', 'miliseconds');
+  const ms = dayjs.convert(7, 'days', 'milliseconds');
   app.use(express.static('public', { index: false }));
   app.use(express.static('build/client', { index: false, maxAge: ms }));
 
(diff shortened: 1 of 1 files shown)


server/routes.js
```diff
@@ -128,7 +128,7 @@ module.exports = function(external) {
   );
 
   app.use(function(req, res, next) {
-    res.header('Cache-Control', 'public, max-age=1');
+    res.header('Cache-Control', 'no-cache');
```
sarahzinger: forgive me I'm not that well versed in caching, reading the docs now. Is the max-age we set above used with no-cache?

GregWeil: They apply to different requests. If a request ends up in the static folder it'll get a max age and stop there, otherwise it will continue on and get this no cache.

sarahzinger: ahh rad that makes a lot of sense, thanks!

PR #991, opened 2019-10-16
Humdrum-triangle: Add 2fa toggle

oliviacpu:
## Links
* Remix link: [https://humdrum-triangle.glitch.me](https://humdrum-triangle.glitch.me)

## Changes:
* Adds a 2FA dev toggle separate from user passwords

## How To Test:
* When "User Passwords is enabled and "Two Factor Auth" is not enabled
  * There should be an option to sign in with passwords
  * There should be a user settings option when signed in
  * Set/reset password should be the only option in user settings
* When both are enabled
  * There should be an extra page in the user settings dialog
  * Enable/disable 2FA and backup codes should be visible

### DIFF
--- a/src/components/account-settings-overlay/index.js
+++ b/src/components/account-settings-overlay/index.js
@@ -1,22 +1,21 @@
 import React, { useState } from 'react';
-import { Icon, SegmentedButton } from '@fogcreek/shared-components';
+import { Icon, SegmentedButton, Button } from '@fogcreek/shared-components';
 
 import Text from 'Components/text/text';
 import { Overlay, OverlaySection, OverlayTitle, OverlayBackground } from 'Components/overlays';
 import PopoverContainer from 'Components/popover/container';
 import { useCurrentUser } from 'State/current-user';
+import useDevToggle from 'State/dev-toggles';
 
 import PasswordSettings from './password-settings';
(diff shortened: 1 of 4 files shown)


src/components/account-settings-overlay/index.js
```diff
@@ -29,15 +28,26 @@ const AccountSettingsOverlay = () => {
...
+                    Password
+                  </Button>
+                  <Button variant="secondary" size="small" disabled={page === '2fa'} onClick={() => setPage('2fa')}>
+                    Two Factor Settings
+                  </Button>
+                </div>
+              )}
+              <div className={styles.accountSettingsContent}>
+                {page === 'password' ? <PasswordSettings /> : null}
+                {page === '2fa' ? <TwoFactorSettings /> : null}
```
sarahzinger: does this need to also be behind the dev toggle or is the concept that it'll never get to here

oliviacpu: Yeah, nobody should get to this unless they've turned on 2FA and at that point they'll need to complete the challenge to sign in. You could disable 2FA and turn off the dev toggle to get back to "just passwords"

oliviacpu:
@sarahzinger I changed the button types. I don't think they should have been secondary.

PR #992, opened 2019-10-16
Long city: Exploring Collections

sarahzinger:
## Links
* Remix link: https://long-city.glitch.me/@glitch/glitch-team-faves/play
* Issue-tracker link: Covers a lot of different tickets, but here's the board with all the related epic tickets: https://app.zenhub.com/workspaces/community-5d07d1359dee6059a9688176/board?epics=192398789_428&repos=117580951,192398789
* Specs/doc links: https://docs.google.com/document/d/1VMTVJE-ZEGA2dIqBknaM25pkODMyfYAEK5FEVPWkPvI/edit#heading=h.3im6lap0txtw

## GIF/Screenshots:
Screencapture: https://recordit.co/ZoSppzVn6v

## Changes:
* Adds a `Play` button to collections
* When you click that button it launches a new play view of collections where you can look at a project embed of each project one at a time.
* Doing this will also change the url so that a particular point in a collection is sharable as well. It'll add the `play/:projectId` to the url, if no projectId is passed in it will default to the first project in a collection

## How To Test:
There is no dev toggle, I'm just putting this all in one big pr, since it seemed relatively self contained to one page. Just go to any collection and check it out! Some particular edge cases to think about:
- how does it look when there are no projects in a collection
- how does it look when there's 1 project in a collection
- how is adding/removing notes
- does reordering a collection mess with anything?
- does the browser's back/forward work as expected
- can you add something to my stuff
- ?? probably other things! try to break it plz :)

## Feedback I'm looking for:
Anything!

- [x] add metrics
- [x] browser testing
- [x] a11y testing
- [x] not sure if we want to do a formal qa session for this mini project or not, will ask the pod.


sarahzinger:
ok seeing some state bugs here:
- if you add a note then click to the next project, then click back the note goes away
- if you reorder projects then press play, it has the old project order

will try to see if there's any easy fixes for that stuff

ttseng:
I'm noticing that the loading speed of the embeds is kind of slow when navigating between apps.

In `glitch-dex-org` we pinged all the projects in the collection so navigating between them would be faster
```
    // ping all the domains in the collection asynchronously ((:
    loadedProjects.forEach( (nextItem) => {
      let xhttp = new XMLHttpRequest();
      xhttp.addEventListener("readystatechange", function () {
        if (this.readyState === this.DONE) {
          // console.log(this.responseText);
        }
      });
      xhttp.open('GET', 'https://'+ nextItem.domain + '.glitch.me', true)
      xhttp.send();
    });
```
https://glitch.com/edit/#!/glitch-dex-org?path=public/new-client.js:101:0

do you think we can do something like that here?

src/components/collection/container.styl
```diff
@@ -37,24 +40,47 @@
     border-radius: 0 5px 5px 0
```
ttseng: I think this is meant to be `0 0 5px 5px`

sarahzinger:
@ttseng I thought i was waking up all the projects here: https://github.com/FogCreek/Glitch-Community/pull/992/files#diff-c97b90dcdd15dc8b4d5b111466dbfc06R32 but maybe it's not working? I'll take another look

src/components/collection/collection-projects-player.js
```diff
@@ -0,0 +1,202 @@
...
+      }
+    });
+  }
+  return currentIndex;
+};
+
+const wakeUpAllProjectsInACollection = (projects) => {
+  const chunkedProjects = chunk(projects, 10);
+  chunkedProjects.map(async (projectsBatch) => {
+    const promisedBatch = projectsBatch.map(async (project) => fetch(`https://${project.domain}.glitch.me`, { mode: 'no-cors' }));
```
oliviacpu: We can pair on this and figure out why it's not working how we expect... I've tried to wake up projects in a similar way and had problems

oliviacpu: So, we tried using the `xhttp` method that @ttseng suggested and are getting a bunch of CORS errors that make it seem like we're not succeeding either way. :(

keithk: Ohh interesting, Tiff wouldn't have had that problem because the glitch-dex-org was on a glitch.me domain.

src/components/collection/collection-projects-player.js
```diff
@@ -0,0 +1,206 @@
...
+  };
+
+  const onClickOnProject = (project, onClose) => {
+    const selectedProjectIndex = findIndex(projects, (p) => p.id === project.id);
+    changeSelectedProject(selectedProjectIndex);
+    onClose();
+  };
+
+  const back = () => {
+    if (currentProjectIndex > 0) {
```
clottman: If the number of projects in the collection changes (due to manipulating the collection in another window and then the user object refreshing), will this handle it gracefully? We did some messy logic in the pagination controller to handle those edge cases that you might want to re-use here.

src/components/collection/collection-projects-player.js
```diff
@@ -0,0 +1,206 @@
...
+    };
+    push(newLocation);
+    setSelectedPopoverProjectId(projects[newIndex].id);
+    setCurrentProjectIndex(newIndex);
+    setAnnouncement(`Showing project ${newIndex + 1} of ${projects.length}, ${projects[newIndex].domain}`);
+  };
+
+  const onClickOnProject = (project, onClose) => {
+    const selectedProjectIndex = findIndex(projects, (p) => p.id === project.id);
+    changeSelectedProject(selectedProjectIndex);
```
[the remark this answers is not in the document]
sarahzinger: this function calls `changeSelectedProject` which makes the setAnnouncement call :)

src/components/collection/collection-projects-player.js
```diff
@@ -0,0 +1,206 @@
...
+      <Popover
+        align="left"
+        renderLabel={({ onClick, ref }) => (
+          <UnstyledButton ref={ref} onClick={onClick}>
+            <span className={classnames(styles.popoverButton, isDarkColor(collection.coverColor) && styles.dark)}>
+              <span className={styles.projectAvatar}>
+                <ProjectAvatar project={featuredProject} />
+              </span>
+              <Text>{featuredProject.domain}</Text>
+              <Icon icon="chevronDown" alt="show project dropdown from collection" />
```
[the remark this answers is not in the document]
sarahzinger: great catch! will fix thanks :)

src/components/collection/collection-projects-player.js
```diff
@@ -0,0 +1,206 @@
...
+                  </ResultItem>
+                </div>
+              )}
+            </ResultsList>
+          </div>
+        )}
+      </Popover>
+
+      <div className={classnames(styles.playerDescription, isDarkColor(collection.coverColor) && styles.dark)}>
+        <Markdown length={80}>{featuredProject.description || ' '}</Markdown>
```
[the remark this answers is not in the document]
sarahzinger: ah interesting thanks!

src/components/collection/collection-projects-player.js
```diff
@@ -0,0 +1,206 @@
...
+  setCurrentProjectIndex: PropTypes.func.isRequired,
+  collection: PropTypes.object.isRequired,
+  push: PropTypes.func.isRequired,
+  params: PropTypes.object.isRequired,
+};
+
+const CollectionProjectsPlayer = withRouter(({ history, match, isAuthorized, funcs, collection }) => {
+  const { projects } = collection;
+  const [currentProjectIndex, setCurrentProjectIndex] = useState(getCurrentProjectIndexFromUrl(match.params.projectId, projects));
+  useEffect(() => wakeUpAllProjectsInACollection(projects), []);
```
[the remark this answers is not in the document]
sarahzinger: yeah I didn't do that on purpose because I thought it was unnecessary to try to wake up all the projects just because one of them changed in someway (such as adding a note for example), BUT this makes me think that maybe I could do more fancy diffing here to see if there are any projects that haven't been woken up yet that need to be like in the case of adding a new project for example

src/components/collection/collection-projects-player.styl
```diff
@@ -0,0 +1,80 @@
...
+
+.resultListWrapper
+  max-height: 400px
+  overflow-x: hidden
+  overflow-y: auto
+  button
+    &:hover, &:focus
+      background-color: general-link
+      color: #fff
+      outline: none
```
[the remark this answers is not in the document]
sarahzinger: @clottman this is the project dropdown on the /play page, if we keep the outline it looks like this:
Screen Shot 2019-10-31 at 1 09 11 PM
My thinking is that because we're making it clear visually which one is in focus, it should be ok if there's no outline

src/components/collection/container.js
```diff
@@ -72,121 +53,109 @@ const CollectionContainer = ({ collection, showFeaturedProject, isAuthorized, pr
...
+      newLocation.state = {
+        preventScroll: true,
+      };
+      setAnnouncement('Collection is in Play View');
+    } else {
+      newLocation.pathname = `/@${match.params.owner}/${match.params.name}/play`;
+      newLocation.state = {
+        preventScroll: true,
+      };
+      setAnnouncement('Collection in Grid View');
```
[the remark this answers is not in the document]
sarahzinger: haha yes thank you words are hard. I like view all mode and play mode :)

src/components/collection/collection-projects-player.styl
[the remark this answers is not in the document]
clottman: Hmm.. I would probably argue that the outline helps users know that it's truly in keyboard focus, but since we have alternate styling it might be okay to leave alone.

src/components/collection/collection-projects-player.js
sarahzinger: hey sorry @clottman just reviewing these comments now. are you asking about having like 2 tabs of the same content and you are editing the collection in one tab and then playing the collection in another tab? 

If so I think it handles it well enough. It essentially will preserve the state of whatever is currently in the window. Is there a specific circumstance in particular I should look out for?

clottman: yeah, that would be the scenario. Or it's a team collection and someone else is editing the collection while you're looking at it. Maybe it's not an issue here though? I don't remember what exactly we were trying to avoid with the pagination stuff on projects, maybe @GregWeil does?

sarahzinger: oh sorry I accidentally deleted the comment you were responding to :( for @GregWeil I was asking if the scenario that we were worried about was having 2 tabs open of the collection page, one we were editing and one we were viewing, and that in my testing it seemed like it worked "well enough" where the play view just had the out of date but functional view before the edits as it doesn't reload state until refresh

GregWeil: The main issue we were having was when the number of pages changed such that your current page no longer exists. I guess the equivalent would be if you can add/remove projects while in the explore view. I think an effect on collection.projects that makes sure the project is still in the collection (and updates the current index if it's been reordered) would be enough.

sarahzinger: yeah I'm trying to recreate that bug and cannot when adding/removing projects, it just preloads the collection and does not ping for updates, so it relies on the old collection data.

If i delete a project in one window while I have the player open it can't load the project in the iframe but otherwise because all the old data is in state it continues to act as if it was in there in terms of respecting the index and things like that. I'm going to leave as is if that's ok!

clottman: Sounds good to me, thanks for doing some testing around it!

GregWeil: Only the current user and one degree away gets shared between tabs, so they'll sync up what collections you have but not the contents of any of them

src/components/collection/collection-projects-player.js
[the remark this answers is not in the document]
sarahzinger: so because this will get called on mount and we remount this component every time we click between play and grid mode, and there's no current way to actually add a project to a collection while you're play mode, there's never a circumstance where you'd need to wake up a project and we weren't already calling this function. I think I'm going to leave as is for now!

Though it was fun to play around with it... I got to learn about lodash's xorBy which was fun: https://lodash.com/docs/4.17.15#xorBy

src/components/collection/collection-projects-player.styl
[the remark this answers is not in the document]
sarahzinger: yeah now that I'm adding yellow backgrounds for the private projects, it seems like the focus outline is very necessary! I'm going to add it back in, I think we have a minor style issue with the outlines, but we can address that in shared-components

ttseng:
The focus outline is getting cut off at the bottom - if you add a ~5px bottom margin to the container, then it looks fine.

image

src/components/collection/collection-projects-player.styl
```diff
@@ -0,0 +1,109 @@
...
+      background-color: general-link
+      
+.private
+  background-color: private
+  button
+    background-color: private
+
+.popoverItem
+  display: flex
+  min-width: 200px
```
ttseng: think we also want `width: 100%` here - without it, the background color doesn't stretch to the width of the menu item:
image

ttseng:
found a weird one!  
image

The dropdown caret becomes really small on medium viewports

src/components/collection/collection-projects-player.js
```diff
@@ -0,0 +1,214 @@
...
+    const newLocation = {
+      pathname: `/@${params.owner}/${params.name}/play/${projects[newIndex].id}`,
+      state: {
+        preventScroll: true,
+      },
+    };
+    push(newLocation);
+    setSelectedPopoverProjectId(projects[newIndex].id);
+    setCurrentProjectIndex(newIndex);
+    setAnnouncement(`Showing project ${newIndex + 1} of ${projects.length}, ${projects[newIndex].domain}`);
```
clottman: When I test this it seems to be doing the right thing, but the code makes it seem like it shouldn't be. Wouldn't `newIndex` be the new index we are moving to, not `newIndex + 1`?

clottman: Because arrays are zero indexed but humans don't tend to think in those terms. Never mind!

PR #995, opened 2019-10-16
Fallacious slice: fix two small bugs I found

GregWeil:
## Links

## Changes:
- Grab `location` before sending it to the join team page. This fixes an ssr error on it, the client just got a key of `undefined`
- Set cookies at the root path rather than on the current page

## How To Test:
Check your cookies on the site, try accepting a team join request (and check the server logs)

### DIFF
--- a/src/state/current-user.js
+++ b/src/state/current-user.js
@@ -16,22 +16,23 @@ const getFromStorage = (key) => readFromStorage(getStorageMemo(), key);
 const setStorage = (key, value) => writeToStorage(getStorageMemo(), key, value);
 
 function identifyUser(user) {
-  document.cookie = `hasLogin=; expires=${new Date()}`;
   if (user) {
     addBreadcrumb({
       level: 'info',
       message: `Current user is ${JSON.stringify(user)}`,
     });
-    if (user.login) {
-      const expires = new Date();
-      expires.setFullYear(expires.getFullYear() + 1);
-      document.cookie = `hasLogin=true; expires=${expires}`;
-    }
   } else {
     addBreadcrumb({
       level: 'info',
       message: 'logged out',
     });
+  }
+  if (user && user.login) {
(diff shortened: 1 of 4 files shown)


src/state/current-user.js
```diff
@@ -16,23 +16,24 @@ const getFromStorage = (key) => readFromStorage(getStorageMemo(), key);
...
   } else {
     addBreadcrumb({
       level: 'info',
       message: 'logged out',
     });
   }
+  if (user && user.login) {
+    const expires = new Date();
+    expires.setFullYear(expires.getFullYear() + 1);
+    document.cookie = `hasLogin=true; path=/; expires=${expires}`;
```
sarahzinger: why do we store this stuff has cookies in general? where is this used?

oliviacpu: Oh I would also like to know

GregWeil: It's used by SSR. Cloudfront splits the cache based on cookie value, and the server can use this to determine how the home page and user popover/sign in button render

sarahzinger: can you point me to the code for this? I'd love to better understand it and not sure where to look 🙏thanks!

GregWeil: `server/routes.js` reads the cookie in `render()` and sends it along as a constant value
`src/state/globals.js` stores the value in `SSR_SIGNED_IN` and spreads it via react context
`home-v2/index.js` reads it and uses it in place of currentUser.login for ssr and the initial render

sarahzinger: ah I see, do we fake sigin via cookies for server side rendering so that there isn't a flash of the sign in buttons?

GregWeil: Yup, the home page has a bunch of extra above the fold content for signed out users, so this keeps that from loading in then popping out a moment later.

PR #997, opened 2019-10-17
circular-snout: pull live homepage and pupdates data from cms.glitch.me

taravancil:
## Links
* [circular-snout.glitch.me](https://circular-snout.glitch.me)

## Changes:
This pulls live homepage and pupdates data from cms.glitch.me instead of from the ~community filesystem.

- Remove `src/curated/{page.json}`
- Remove part of merge script that handles live updates to home.json
- Remove routes `/api/{pupdates|home}` and the helpers in `server/curated` that wrote pupdates.json and home.json to the filesystem

## How To Test:
* The homepage should load with all the content we expect
* Open pupdates, they should all load

## Feedback I'm looking for:
- How do we make sure cms.glitch.me is always online? Is it reasonable to assume that ~community will ping it for data often enough that it will never go to sleep? That feels risky.
- Does this approach seem sound?


server/curated.js
```diff
@@ -1,36 +1,13 @@
...
-const readFile = util.promisify(fs.readFile);
-const writeFile = util.promisify(fs.writeFile);
-
 let pageCache = {};
 
 async function getData(page) {
   if (!pageCache[page]) {
-    const json = await readFile(path.join(__dirname, `../src/curated/${page}.json`));
-    pageCache[page] = JSON.parse(json);
+    const { data } = await axios.get(`https://cms.glitch.me/${page}.json`);
```
modernserf: Where is this cache being updated?

sh/merge.sh
```diff
@@ -8,10 +8,6 @@ if [ "$1" == "-h" ]; then
   exit 0
 fi
 
-curl https://community.glitch.me/api/home > src/curated/home.json
```
modernserf: Does this script need to exist at all if not for this curated content?

modernserf:
For preventing it from going to sleep, i imagine we could just add it to the Performance Boost team?

oliviacpu:
@modernserf Does Performance Boost also keep awake, or just add more memory/resources?

sh/merge.sh
oliviacpu: I think a version of it needs to exist for getting live changes onto a branch now, but not directly pushed to master. Once we're not doing "Glitch hosted" Community anymore then we won't have "live changes" and probably won't need this.

modernserf: what live changes are we doing besides CMS stuff?

server/curated.js
oliviacpu: I think there's another place we can put this that I'm going to try out

sh/merge.sh
oliviacpu: I think small fixes to deployed PRs have been considered fine "live changes" outside of the CMS stuff?

modernserf:
So now the homepage/pupdate cache will update every hour -- does that work for marketing? I know they get concerned when the 'zine takes a while to update.

oliviacpu:
@modernserf There is a 15 minute cache for zine content that we could use as well. I can change it to use that one.

keithk:
Yeah, can we try making this a 5m cache?

oliviacpu:
🙋 We're going to let the CMS push updates to us, so this is getting closed.

PR #1006, opened 2019-10-18
Add "resume coding" to home page

oliviacpu:
## Links
* Remix link: https://green-impatiens.glitch.me/

## Changes:
* Adds the "Resume Coding" button back, using the diff from this PR https://github.com/FogCreek/Glitch-Community/pull/955

## How To Test:
* When you're signed in and have projects you should see the resume coding button, and clicking it should take you to the editor

### DIFF
--- a/src/presenters/pages/login/reset-password.js
+++ b/src/presenters/pages/login/reset-password.js
@@ -5,7 +5,6 @@ import dayjs from 'dayjs';
 import { Button } from '@fogcreek/shared-components';
 
 import { useAPI } from 'State/api';
-import { useCurrentUser } from 'State/current-user';
 import useLocalStorage from 'State/local-storage';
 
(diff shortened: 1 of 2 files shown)


src/presenters/pages/login/reset-password.js
```diff
@@ -5,7 +5,6 @@ import dayjs from 'dayjs';
 import { Button } from '@fogcreek/shared-components';
 
 import { useAPI } from 'State/api';
-import { useCurrentUser } from 'State/current-user';
```
sarahzinger: @oliviacpu is this file change supposed to be there? Seems like reset password should be unrelated yeah?

oliviacpu: Good catch! Thank you! Maybe this should be in my other PR?

PR #1012, opened 2019-10-22
Profuse tea: Fix project page flicker, one path for 404 errors

GregWeil:
## Links
https://app.clubhouse.io/glitch/story/2985/project-page-more-by-section-flickers-a-bunch-on-page-load

## Changes:
- The 'more by' section on project pages memoizes args so it doesn't keep reloading every render
- Added a try/catch to handle 404s when loading project members

## How To Test:
- Try loading up a project page and make sure there is no flickering
- Pin a project on your user page, refresh, delete it, then refresh again. On glitch.com a 404 will appear in the console, on this remix it won't. The 404 happens because the site ssr caches the deleted project, and the browser makes the request for members before it realizes the project is gone.

### DIFF
--- a/src/components/related-projects/index.js
+++ b/src/components/related-projects/index.js
@@ -1,4 +1,4 @@
-import React, { useCallback } from 'react';
+import React, { useMemo } from 'react';
 import PropTypes from 'prop-types';
 import { sampleSize } from 'lodash';
 import { Icon } from '@fogcreek/shared-components';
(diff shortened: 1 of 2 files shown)


src/components/related-projects/index.js
```diff
@@ -41,10 +41,18 @@ async function getProjects(api, { type, id, ignoreProjectId }) {
   return [...sampledPins, ...sampledRecents];
 }
 
-function RelatedProjects({ project }) {
+const ProjectsDataLoader = ({ children, type, id, ignoreProjectId }) => {
```
oliviacpu: Neat! Would this be useful on any other components that load project data?

GregWeil:
Chrome always logs 404s even if they're caught. The error should go away in firefox and not get logged to sentry though.

src/components/related-projects/index.js
GregWeil: I don't think so. The `getProjects` function isn't used anywhere else.

PR #1013, opened 2019-10-23
Fix homepage bug

sarahzinger:
## Links
* Remix link: https://glass-mercury.glitch.me/
* Issue-tracker link: https://app.clubhouse.io/glitch/story/1680/mobile-version-of-homepage-is-not-properly-rendering-html

## GIF/Screenshots:
Screen Shot 2019-10-23 at 10 34 17 AM

## Changes:
* Fix a bug on the homepage on small width screens where we weren't rendering the Fresh apps descriptions the same way we were on wider screens https://github.com/FogCreek/Glitch-Community/blob/master/src/presenters/pages/home-v2/index.js#L109

## How To Test:
* look at the fresh apps section of the homepage on mobile and big screens, make sure we're rendering html where appropriate


sarahzinger:
I believe we are getting those from our CMS so I think we should be good!

PR #1014, opened 2019-10-23
cms-updates: Check for a secret instead of persistentToken and team membership

oliviacpu:
## Links
* Remix link: 
* Issue-tracker link
* Specs/doc links

## Changes:
* Moves the authorization check before writing curated content to disk up to the route level
* Checks for a secret that the CMS is sending rather than a specific user's `persistentToken`
* Renames a couple of methods to describe what they're doing more specifically

## How To Test:
* I need to set this up

## Feedback I'm looking for:
- Is this secret thing okay? It seems better than sending and checking someone's `persistentToken`.

- [ ] Add environment variables to staging before deploying, and again after swapping projects

### DIFF
--- a/server/curated.js
+++ b/server/curated.js
@@ -1,24 +1,13 @@
 const fs = require('fs');
 const util = require('util');
 const path = require('path');
-const axios = require('axios');
-
-const { API_URL } = require('./constants').current;
-const { getAllPages } = require('Shared/api');
-
-const GLITCH_TEAM_ID = 74;
-
-const api = axios.create({
-  baseURL: API_URL,
-  timeout: 5000,
-});
 
 const readFile = util.promisify(fs.readFile);
 const writeFile = util.promisify(fs.writeFile);
 
 let pageCache = {};
 
-async function getData(page) {
+async function readCuratedContent(page) {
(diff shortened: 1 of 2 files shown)


oliviacpu:
Semi-related to this PR, I think the CMS server should be sending updates to _both_  `community` and `community-staging` unless we're also managing merging in the latest CMS updates by a script or something.

Tara mentioned getting some merge conflicts in the files when that was attempted, but it seems like since these are hands-off now they can be .gitignored pretty safely?

oliviacpu:
@sarahzinger 
> Lingering questions, that would be good to be documented before deploying if possible!
> 
> * how frequently will this ping for new curated content from the cms? On refresh of the server so on every deploy?
> * what should we do about the merge script?

* This won't ping the CMS anymore. The CMS will send updates as soon as they're made to our `POST` route for this.
* I can check the merge/CMS script and make sure it does what we want in this PR before we deploy.

One more thing I was thinking is that the route might be better as `/curated/:page` instead of `/api/:page` since it isn't really part of the "API" as we know it

sarahzinger:
@oliviacpu ooh yeah all of that makes a lot of sense! thanks!

PR #1015, opened 2019-10-23
Glowing draw: Update Embed to optionally point directly to project

sarahzinger:
## Links
* Remix link: https://glowing-draw.glitch.me/@glitch/glitch-team-faves/play
* Issue-tracker link: https://app.clubhouse.io/glitch/story/1691/create-or-reuse-an-iframe-to-display-current-project-item-in-a-collection
* Specs/doc links: https://fogcreek.slack.com/archives/CNW257T7E/p1571759715050000

## GIF/Screenshots:
before/after
Screen Shot 2019-10-23 at 5 09 24 PM

## Changes:
* before we always pointed our embeds to glitch.com/embed/#!/embed/project-name which meant that as you tabbed through the different projects in a collection it was pretty slow, even waking the projects up all in one go on load of the page didn't quite work, as we still had to load the code, not just the project.
* now we optionally let the project embeds point directly to the project at projectname.glitch.me, which means we can wake them all up when we load the component, making them really easy to breeze through quickly. 
* the only issue there is that there's no embed bar that links to who made the project and stuff like that, so we have to mimic that on our end here. 

## How To Test:
* compare clicking around https://long-city.glitch.me/@glitch/glitch-team-faves/play with https://glowing-draw.glitch.me/@glitch/glitch-team-faves/play the glowing-draw one should be much faster

## Feedback I'm looking for:
anything!

not really a thing to do but a thing to note is that this pr is based off of long-city. We're not really ready to merge that pr yet, but I wanted to make it easier to look at pieces of the work to get better feedback, so I'll merge this and it won't get deployed for a bit.

### DIFF
--- a/src/components/project/embed.js
+++ b/src/components/project/embed.js
@@ -1,5 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import classnames from 'classnames';
+
 import { APP_URL } from 'Utils/constants';
 import Image from 'Components/images/image';
 import styles from './embed.styl';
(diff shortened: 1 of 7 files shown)


src/components/project/embed.js
```diff
@@ -24,19 +26,19 @@ const browserSatisfiesRequirements = (() => {
   return false;
 })();
 
-const Embed = ({ domain, loading }) => (
-  <div className={styles.embedContainer}>
+const Embed = ({ domain, loading, hideCode }) => (
```
oliviacpu: My only suggestion is that this might be more clear as `previewOnly` instead of `hideCode`

It feels more descriptive of what it is doing (loading only the preview/live app and not the embed that contains the code & preview)

PR #1022, opened 2019-10-25
Fallacious slice: Don't load stuff right away, wait until it's relevant

GregWeil:
## Links
https://fallacious-slice.glitch.me/@Performance-Boost
https://app.clubhouse.io/glitch/story/3199/investigate-why-the-site-is-making-so-many-network-requests

## Changes:
- Team pages only load projects owned by team members once you open their user pop, rather than loading every project owned by every team member right away
- Don't load project item teams/members until they are visible on the page

## How To Test:
- Try removing yourself from a team. It should act no different, except that it doesn't load all of your projects until you open your team user pop
- Clicking show all on pages with many projects does load everything eventually, and there are no 429 errors

## Feedback I'm looking for:
Team pages with a lot of projects slow down quite a bit as you scroll because every time a project loads the context causes every project to re render

### DIFF
--- a/shrinkwrap.yaml
+++ b/shrinkwrap.yaml
@@ -9,7 +9,7 @@ dependencies:
   '@babel/register': 7.5.5
   '@babel/runtime': 7.6.3
   '@babel/runtime-corejs3': 7.6.3
-  '@fogcreek/shared-components': 0.8.0
+  '@fogcreek/shared-components': 0.11.3
   '@jedmao/location': 2.0.4
   '@optimizely/optimizely-sdk': 3.3.0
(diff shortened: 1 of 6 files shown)


sarahzinger:
clicking show all and scrolling leads to a lot of blank space that doesn't seem to fix itself, do you experience that as well? Maybe instead of Show all, we can do show 50 more or something like that?

GregWeil:
If you wait a minute or two it'll sort itself out, but it is way too long. I'm trying to improve project item rendering performance in another remix, but my goal here was to fix the bugs causing api timeouts

PR #1023, opened 2019-10-25
Thinkable gold - Change Add to Collection Button Project Pages

ttseng:
## Links
* Remix link: https://thinkable-gold.glitch.me/~stencilfy
* Issue-tracker link: https://app.clubhouse.io/glitch/story/1738/updated-shared-component-button-to-allow-for-overflow-button-text

## GIF/Screenshots:

image
image

## Changes:
* Change "Add to Collection" button text to "Add" on small viewports so the buttons look more consistent (single-line)

## How To Test:
* go to any project page and see what it looks like on small viewports (mobile)


src/components/project/add-project-to-collection-pop.js
```diff
@@ -20,6 +21,8 @@ import { ProjectAvatar } from 'Components/images/avatar';
 import CollectionResultItem from 'Components/collection/collection-result-item';
 import { CreateCollectionWithProject } from 'Components/collection/create-collection-pop';
 import { AddProjectToCollectionMsg } from 'Components/notification';
+import { mediumSmallViewport } from 'Components/project/project-actions';
```
clottman: This feels like the wrong place for a constant that we can reasonable expect to use in lots of different components. Can we move it to maybe alongside the use-window-size hook, or to utils/constants.js?

src/components/project/add-project-to-collection-pop.js
```diff
@@ -193,35 +204,45 @@ AddProjectToCollectionBase.propTypes = {
...
-          createCollectionPopover: () => (
-            <CreateCollectionWithProject
-              addProjectToCollection={(...args) => {
-                addProjectToCollection(...args);
-                togglePopover();
-              }}
+const AddProjectToCollection = ({ project, addProjectToCollection }) => {
+  const [width] = useWindowSize();
+  let buttonText = null;
+  if (width && width < mediumSmallViewport) {
```
clottman: Since width is set by an effect, is it possible we're flashing the wrong text here momentarily? If so I think we can fix it by initializing the state with the actual values instead of 0s in the useWindowSize hook. 

```git
- const [windowSize, setWindowSize] = useState([0, 0]);
+ const [windowSize, setWindowSize] = useState([window.innerWidth, window.innerHeight]);
```

src/components/project/add-project-to-collection-pop.js
ttseng: good idea - moved to the use-window-size hook

src/components/project/add-project-to-collection-pop.js
ttseng: done!

shrinkwrap.yaml
```diff
@@ -11031,7 +11031,11 @@ specifiers:
   '@babel/register': 7.5.5
   '@babel/runtime': ^7.5.4
   '@babel/runtime-corejs3': ^7.5.5
+<<<<<<< HEAD
```
clottman: Still some merge conflicts here

PR #1024, opened 2019-10-28
Charmed albertonykus: create /settings route for user deletion

whimsicallyson:
## Links
https://charmed-albertonykus.glitch.me/settings
[ch3219]

## GIF/Screenshots:
Screen Shot 2019-10-28 at 2 58 42 PM

## Changes:
* Moved password and 2fa settings into a page (instead of a modal)
* Changed avatar menu so that the account settings button links to that page

## How To Test:
* Turn on some feature switches
* Go to /settings
* Make sure what you see jives with the feature switches you have set
* Make sure you can use the features

## Feedback I'm looking for:
Should I remove the (disabled) privacy button?
The modal has a "Email notifications are sent to {email}" message, should we have that here too?


sarahzinger:
its coming together :) One thing that comes to mind that still needs to be done is making sure you can't get to this route unless the auth stuff is enabled with the dev toggle

sarahzinger:
Also since privacy and notifications don't exist yet maybe we should just remove the button?

whimsicallyson:
ok I think that's all updated!

sarahzinger:
cool another thing! it probably doesn't make sense to go to the /settings page unless you're authenticated yeah? do you want to ensure you have to be logged in to see the route?

src/presenters/pages/router.js
```diff
@@ -167,16 +179,19 @@ const Router = () => {
...
+          <Route key={category.url} path={`/${category.url}`} exact render={() => <Redirect to={`/@glitch/${category.collectionName}`} />} />
         ))}
 
         <Route path="/secret" exact render={({ location }) => <SecretPage key={location.key} />} />
 
+        <Route
+          path="/settings"
+          exact
+          render={({ location }) =>
+            <>{ settingsPageEnabled ? <SettingsPage key={location.key} /> : <NotFoundPage key={location.key} />}</>
```
sarahzinger: this is totes fine as is, but I remember you expressed a desire for this to be a one liner and now that it's not `settingsPageEnabled && <SettingsPage />` it can be! If you want you don't have to use the fragments and can instead do a one liner:
`render={({ location }) => settingsPageEnabled ? <SettingsPage key={location.key} /> : <NotFoundPage key={location.key} />}`

but also that's just stylistic stuffs so no need to change if you don't want to

sarahzinger:
cool this is looking good to me! The flash is annoying but I don't see a super easy way around it right now and it won't actually be an issue when we release to customers :) 

Not sure what we'll want to do about this account button, but we can think about that when we're closer to release and chat with Tiff. Maybe we'll just end up taking out the left sidebar, but we can do that later :)

Before we merge let's just double check with Olivia when she gets in that this doesn't impact anything she is doing with the signin project. I'm pretty sure it does not because that's for logging in/signing up, and this is for changing your password and stuff like that. But just in case!

One ticket bites the dust :)

src/presenters/pages/router.js
```diff
@@ -167,16 +179,17 @@ const Router = () => {
...
-            path={`/${category.url}`}
-            exact
-            render={() => <Redirect to={`/@glitch/${category.collectionName}`} />}
-          />
+          <Route key={category.url} path={`/${category.url}`} exact render={() => <Redirect to={`/@glitch/${category.collectionName}`} />} />
         ))}
 
         <Route path="/secret" exact render={({ location }) => <SecretPage key={location.key} />} />
 
+        <Route
```
clottman: Would it be cleaner to make the conditional be on whether this Route is defined at all? That way if we for some reason make changes to how 404s are displayed, we won't need to remember to update it here. 

So `{settingsPageEnabled && <Route... render={<SettingsPage>}/>}`

src/presenters/pages/settings.js
```diff
@@ -0,0 +1,39 @@
...
+import styles from './settings.styl';
+import { emoji } from '../../components/global.styl';
+
+const Settings = () => {
+  const tagline = 'Account Settings';
+  const userPasswordEnabled = useDevToggle('User Passwords');
+  const tfaEnabled = useDevToggle('Two Factor Auth');
+
+  return (
+    <main>
```
clottman: Need to add `id='main'` here so the skip link works.

whimsicallyson:
@oliviacpu do these changes have any effect on your sign in work? my hunch is that it shouldn't, but I want to check before deploying and breaking anything!

PR #1029, opened 2019-10-29
Nova chance: Clean up ssr context handling

GregWeil:
## Links
https://app.clubhouse.io/glitch/story/1593/a-b-testing-with-optimizely (am adding optimizely stuff in a separate remix and there are too many places that needed the var added)

## Changes:
- Define the context vars once rather than three times
- Removed the try/catch around ssr because the caching layer catches errors
- Set a cache fallback value for the zine and ssr, don't use `val || default` everywhere
- Don't rename vars when putting them in the context, name them correctly in the first place

## How To Test:
Try loading up a bunch of pages and refreshing to make sure ssr works


views/index.ejs
```diff
@@ -25,7 +25,7 @@
     <!-- styles from styled-components -->
     <%- styleTags %>
     
-    <% if (!rendered) { %>
+    <% if (!html) { %>
```
clottman: Where did the rename for this happen? I'm not seeing where `rendered` came from in the first place.

GregWeil: Line 93 in the old routes.js

PR #1031, opened 2019-10-30
Various dart: update deploying steps/scripts

whimsicallyson:
## Changes:
* update `deploying.md` and `merge.sh` to a flow that won't run afoul of github's merge requirements

## Feedback I'm looking for:
Since master is currently up to date with all of my code, I can't exactly test this. So hoping someone with lots of confidence in their Git knowledge can confirm that this is what we want!

### DIFF
--- a/DEPLOYING.md
+++ b/DEPLOYING.md
@@ -2,8 +2,6 @@
 
 Only employees of Glitch will be able to do this step, and here it is! 
 
-You should perform these steps immediately after clicking the "Merge pull request" button on your PR in Github. 
-
 ### Announce the Deployment
 
 Be present in our `#community-deploy` chat room and say that you're starting a deploy and what you're deploying.
@@ -18,17 +16,15 @@ In your local git repository:
 
 ```
   # Make sure we're in the right place and up to date.
-  git checkout master
-  git pull
+  git pull master
 
   # Pull in any live changes that aren't yet in Master:
   git pull live master
   
   # Merge if needed
-  # [ This is on you. 🐉 ]
   
   # Then push these changes back up to github
-  git push origin master
+  git push origin my-remix
   
 ``` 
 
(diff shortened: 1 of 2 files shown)


sh/merge.sh
```diff
@@ -12,7 +12,6 @@ curl https://community.glitch.me/api/home > src/curated/home.json
 git add src/curated/home.json
 git commit -m "update home page content"
 
-git checkout master
-git pull
+git pull master
```
clottman: This command, when run from my local branch, results in 
```
fatal: 'master' does not appear to be a git repository
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

I think what we want is `git pull origin master` on line 15.

DEPLOYING.md
```diff
@@ -2,8 +2,6 @@
 
 Only employees of Glitch will be able to do this step, and here it is! 
 
-You should perform these steps immediately after clicking the "Merge pull request" button on your PR in Github.
```
clottman: Can we replace this line with adding (somewhere) that the merge step should now ideally happen before merging the pull request in github?

clottman: Yeah - could even make it more explicit by pointing out that the `live` remote is the part that we need merged in and passing Circle CI.

> merge in all changes, including those from the live remote (the ~community project on Glitch), before...

DEPLOYING.md
```diff
@@ -17,23 +17,21 @@ We need to make sure any live changes are merged and happy.  This is a Glitch si
 In your local git repository:
 
 ```
-  # Make sure we're in the right place and up to date.
-  git checkout master
-  git pull
-
-  # Pull in any live changes that aren't yet in Master:
+  # Make sure we're up to date, including any live changes.
+  git pull master
```
clottman: This should be `git pull origin master`

PR #1035, opened 2019-10-30
Fix a couple popover styling bugs

whimsicallyson:
## Links
* https://cookie-ricotta.glitch.me

## GIF/Screenshots:
before components: 
Screen Shot 2019-10-30 at 4 04 59 PM
current prod:
Screen Shot 2019-10-30 at 4 05 20 PM
fixed:
Screen Shot 2019-10-30 at 4 05 31 PM

## Changes:
* Long titles weren't wrapping nicely, now they should be!
* There was also a bit of extra padding on the search results list in add-project-to-collection

## Additional Changes!
* removed some extra padding from the new-project-pop
* Fixed an error where the notification would claim you weren't able to create a collection with a project

## How To Test:
* Check out popovers related to long names (be it your username, or a project/collection)
* Create a new collection from a project page


src/components/global.styl
```diff
@@ -63,8 +63,8 @@ section
 div[data-module="Popover"]
```
sarahzinger: I do wonder in general why we have global styles for popovers? are all of our popover supposed to have a fixed with?

whimsicallyson: we had a `wide` prop that we would pass in in the past, it seemed the easiest/best way to keep that was with a globally available class name?

sarahzinger: this is not a globally available class name tho, this is going to style every single popover

whimsicallyson:
temporarily closing to deal with other things

src/components/global.styl
whimsicallyson: ohhh misunderstood your question. most of our popovers do not have a fixed width, because of long collection names and such. when the popovers are filled with standard html tags, it works fine, but when we're combining text nodes within react it gets a bit confused, so adding the `max-content` fixes that. There's an additional classname below to mimic the `wide` prop

PR #1044, opened 2019-11-01
Superb guide: Optimizely refactors + support for ssr based on user id

GregWeil:
## Links
https://app.clubhouse.io/glitch/story/1593/a-b-testing-with-optimizely

## Screenshots
image

## Changes:
- Track a new cookie `optimizely-id` which is set to the current user id if you are logged in. The server assigns a random number if you don't already have an id
- SSR caches based on the current user's feature flags (so it doesn't have to render every request)
- Added a table to the secret page that shows what flags are active for your user id

## How To Test:
Try loading up the site and check what features are enabled, they should stay across reloads. Sign in, and notice that the enabled features might have changed. Sign out, they should stay the same as when you were signed in.

## Feedback I'm looking for:
The analytics feature is based on the team id, but the secret page uses your user id. Should it do anything to indicate that?

- I need to get the `optimizely-id` cookie whitelisted in cloudfront
- (not blocking deploy) there should be a way to force enable/disable features


GregWeil:
Note: this remix is using a debug optimizely config. The analytics and test feature are both set to 50%, which is not the case in prod

package.json
```diff
@@ -49,7 +49,7 @@
     "@babel/runtime-corejs3": "^7.5.5",
     "@fogcreek/shared-components": "^0.13.2",
     "@jedmao/location": "^2.0.4",
-    "@optimizely/optimizely-sdk": "^3.2.2",
+    "@optimizely/optimizely-sdk": "^3.3.0",
```
sarahzinger: do we have to do anything for the yaml file?

GregWeil: There is a change to shrinkwrap, but it's smaller than the usual package update. Maybe it already got silently updated?

PR #1045, opened 2019-11-01
Charmed albertonykus

whimsicallyson:
## Links
* https://charmed-albertonykus.glitch.me/settings
* https://app.clubhouse.io/glitch/story/3221/add-a-delete-account-section-to-the-new-settings-page-with-a-button-to-delete-the-account
* https://docs.google.com/document/d/1ya5WQraU0OPXf22uCsTGhaNnFU-3XyVLHSqGHzMkTPk/edit#heading=h.3im6lap0txtw

## Changes:
* Adds a button to delete your account, hidden behind a dev toggle

## How To Test:
* It doesn't actually _do_ anything yet, so...just make sure its a button? And it only shows up with the dev toggle on, when you're already logged in?

## Feedback I'm looking for:
Should this be broken out into a separate file and imported?

- [x] Have Tiff add a coffin image to shared-components
- [x] Update/publish shared-components to be able to source the coffin icon


sarahzinger:
@whimsicallyson I think something is up with your remix when I go there it tells me my browser doesn't support js :) 

Also how about to make this pr just slightly beefier we make it so when you click on the button it opens a modal, (and we can add the cool insides for the modal in a separate pr) :)

sarahzinger:
jkjk my browser apparently had some ancient broken code cached somehow? (which seems like a bug! but not for this pr lol)

sarahzinger:
if I don't have 2fa enabled i see this, is that what you see? 
Screen Shot 2019-11-05 at 1 26 20 PM

looks like we need a background color or something, also feels like the padding is a little off right?

whimsicallyson:
fixed the padding, and https://github.com/FogCreek/shared-components/pull/64 should fix the z-index issue

src/presenters/pages/settings.styl
```diff
@@ -14,4 +14,7 @@
   flex-basis: 80%
   padding-left: 10px
   .contentSection
-    padding-bottom: 100px
\ No newline at end of file
+    padding-bottom: 100px
+
+[data-module="Button"].modalButton
```
sarahzinger: does .modalButton not work here?

whimsicallyson: it does not, there was a bit of discussion here yesterday: https://fogcreek.slack.com/archives/CB4EHAJ2J/p1572972710440800

sarahzinger: hmm well what do you think? I can see a few options:
- we leave as is and fix another day
- we try to get to root of figuring out why we can't add classnames direction
- we wrap it in a div or a span or something (i think in the case of something like a margin that makes a lot of sense to me)
- probably other options too!

whimsicallyson: I think a combo of 1&2--this is definitely not the only place with this setup, so fixing it just here seems wrong. If (as Justin says) this isn't a pattern we want to be using, let's take some time to figure out why just adding a classname won't work and we can remove all instances of the anti-pattern at that point.

sarahzinger: sounds great! you wanna make a ticket and we can try to figure it out next sprint?

PR #1052, opened 2019-11-06
Thinkable-Gold: Empty /Search Page Design

ttseng:
## Links
* Remix link: https://thinkable-gold.glitch.me/search
* Issue-tracker link: https://app.clubhouse.io/glitch/story/1682/search-page-has-no-default-landing-page

## GIF/Screenshots:
image

## Changes:
* Added search field to /search - note that autocomplete results are not displayed on this page automatically as you type.
* Added Surprise Me button that takes you to a random project from https://glitch.com/@glitch/community-picks

- [x] Confirm with dev advocates and marketing about whether community-picks is the best place to draw projects from


ttseng:
note - have already ran this by marketing https://fogcreek.slack.com/archives/CJ6HUS7L6/p1573064938009100

ttseng:
Also added it to the no results page
image

src/components/buttons/random-project-button.js
```diff
@@ -0,0 +1,25 @@
...
+  const owner = 'glitch';
+  const name = 'community-picks';
+  const { value: collection } = useCachedCollection(`${owner}/${name}`);
+
+  return (
+    <>
+      {collection && (
+        <ProjectLink project={collection.projects[Math.floor(Math.random() * collection.projects.length)]}>
+          <Button as="span">
+            Surprise Me <Icon icon="diamondSmall" />
```
sarahzinger: i wonder if we need an alt tag here... maybe not?

src/presenters/pages/search.js
```diff
@@ -3,31 +3,41 @@ import PropTypes from 'prop-types';
...
   const searchResults = useAlgoliaSearch(query);
   const setActiveFilter = (filter) => {
     history.push(`/search?q=${encodeURIComponent(query)}&activeFilter=${filter}`);
   };
 
   return (
-    <AnalyticsContext properties={{ origin: 'search', query }}>
-      <Layout searchQuery={query}>
+    <Layout searchQuery={query}>
+      <AnalyticsContext properties={{ origin: 'search', query }}>
```
sarahzinger: I do worry this will break some analytics event, like if someone clicks new project or login or something in layout it might not know they did so from the search page? maybe that's not too important tho?

src/components/buttons/random-project-button.js
ttseng: added empty alt-text just in case

ttseng:
I added a slight animation to the surprise me button

surprise-me-btn

src/components/buttons/random-project-button.styl
```diff
@@ -0,0 +1,36 @@
+.randomProjectButtonWrapper
+    margin: 2em 0
+    animation: wiggle ease-in 2.5s infinite
```
oliviacpu: Is this subtle enough to leave on if "Reduce motion" is preferred? I'm not sure how much motion there should be before we consider that?

src/presenters/pages/search.js
oliviacpu: Hmm... would we have a way to check this at all?

ttseng:
I removed the `surprise me` button and will leave additional design work re: suggesting apps for another ticket.

For now, the default search page just has an empty search bar, instead of returning an error:

image

PR #1056, opened 2019-11-07
hissing-gibbon: Update "create team" to use new routes

oliviacpu:
## Links
* Remix link: [https://hissing-gibbon.glitch.me/](https://hissing-gibbon.glitch.me/)
* Issue-tracker link: https://app.clubhouse.io/glitch/story/3159/port-teamid-byurl-url-and-userid-bylogin-login-to-v1-routes

## Changes:
* Changes two routes to use new V1 routes

## How To Test:
* Try to create a team, enter a known username and team name that are taken, you should see an error that says the name is unavailable
* An unused username and teamname should be allowed when creating a team


src/components/create-team-pop/index.js
```diff
@@ -35,8 +35,8 @@ const CreateTeamPop = withRouter(({ onBack, history }) => {
       let error = null;
 
       try {
-        const { data } = await api.get(`userId/byLogin/${url}`);
-        if (data !== 'NOT FOUND') {
+        const { data } = await api.get(`v1/users/by/login?login=${url}`);
+        if (Object.keys(data).length > 0) {
```
sarahzinger: is this the only way to check that a name is in use? I could imagine getting back a 401/500/some other error for example and there being no data as well right? seems like the server should return the reason?

sarahzinger:
I also notice this happens when I create a team now: https://recordit.co/hhKb9NnMRT it fixes itself once i hard refresh, i think it's new to this pr. maybe we have some out of date caching logic or something?

oliviacpu:
@sarahzinger Updating with the latest changes seems to have fixed it.

sarahzinger:
hey hey! we talked about this on slack, but just a reminder I'm seeing the 404 when we create a new team after hitting an error first, because we cache the error or something. If it's an easy fix lets fix it here! if not I'm chill to just deploy as is, as long as we make a ticket to fix it someday!

PR #1060, opened 2019-11-11
Superb guide: Force a restart if the server uses more than 2gb

GregWeil:
## Links

## Changes:
Check every second, if the site is using more than 2gb then force exit (forever will then restart it, without having webpack/babel also restart like a glitch memory limit would)

## How To Test:
Add an interval that keeps doubling the length of an array and wait until it restarts


GregWeil:
Abandoning this pr

PR #1067, opened 2019-11-13
Bring Back Exploring Collections aka Play Mode

sarahzinger:
Reverts FogCreek/Glitch-Community#1066

This pr is a revert of a revert, with one more commit and a merge of the latest master. 

remix with code available: https://flying-sidecar.glitch.me/@glitch/glitch-team-faves/play

Brings back the Exploring Collections feature, but this time only wakes up a few projects at a time.
To see what projects it wakes up I'd suggest looking at the app with the network tab open and filter results to urls containing `.glitch.me`, as you tab through you can see what projects we wake up
Screen Shot 2019-11-18 at 11 54 46 AM

I currently have it set to wake up 6 projects at a time, and I make this request every 5th project to ensure we're always viewing one project "behind" the ones we're waking up. Here's a weird diagram I attempted to make to show what I'm talking about:
Screen Shot 2019-11-18 at 11 29 21 AM
Each number represents a project index within a collection. Everytime we hit an index with an arrow underneath we wake up the next 6 projects (represented by boxes)

The main changes can be seen here: https://github.com/FogCreek/Glitch-Community/pull/1067/files#diff-c97b90dcdd15dc8b4d5b111466dbfc06R143

The downside of this approach is if I share with someone a link to a particular project within a collection and that index is not evenly divisible by 5, then no projects will wake up until someone tabs to the project where it's index is divisible by 5. It also servers no benefits for users who are tabbing backwards. But I think for most people they'll start with 0 and tab mostly forward, and for those users they'll have a fairly seamless experience. Perhaps if this ships well we can go back to waking all the projects up at once (or at least the first 20 projects or something like that)


sarahzinger:
Yasi feels we're better off not waking up any projects right now: https://fogcreek.slack.com/archives/CB4EHAJ2J/p1574096692099000

so I updated this pr not to wake up any projects. Hopefully after we release this we can iterate and wake up more and more over time.

PR #1070, opened 2019-11-14
Move to PM2 to manage community

keithk:
## Links
[ch4155]
https://nettle-halloumi.glitch.me/
https://glitch.com/edit/#!/nettle-halloumi

## Changes:
* Removes nodemon/forever, and replaces it with PM2
* _Enables the node cluster mode, which will run multiple instances of community at once. It determines this itself by the number of cores available_ . - Temporarily disabled
* _If any of those instances run into 1GB of memory (we have 4GB total) it will gracefully restart, while the other instances will continue serving traffic_ Set to 3GB for now

This should be a fine bandaid for our current issues and downtime. The downside is that it will spin up 2-3 instances, and those will have their own caches to warm up so it's likely the site will be slower for a little bit if the pm2 gets restarted entirely (if the container itself forces a restart)

## Some cool things
- You can run `pm2 monit` from the console to get this view:
Screen Shot 2019-11-14 at 12 02 30 PM

Which lets you see the instances running, and the amount of memory being used. In addition the **additional metrics** section shows new tracking stats like:

Screen Shot 2019-11-14 at 12 06 57 PM

Which should help us in debugging going forward

## Feedback I'm looking for:
- Does this make sense? Does it work?
Most of the changes are in `ecosystem.config.js`


.pm2/touch
```diff
@@ -0,0 +1 @@
+1573761130557
```
GregWeil: This looks like a runtime artifact. If so, could it get added to the gitignore

package.json
```diff
@@ -21,6 +21,7 @@
     "build": "npm run build:webpack && npm run build:babel",
     "build:babel": "npx babel ./src --config-file ./.babelrc.node.js --copy-files -d ./build/node",
     "build:webpack": "npx webpack",
+    "build:webpack:prod": "NODE_ENV=production npx webpack && npm run build:babel",
```
GregWeil: This seems equivalent to `NODE_ENV=production npm run build`. Should it just be called `build:prod`?

ecosystem.config.js
```diff
@@ -0,0 +1,16 @@
+module.exports = {
+    apps : [{
+      name: "glitch-community",
+      script: "./server/server.js",
+      instances : "1",
+      max_memory_restart: "1G", // This will restart gracefully the service if it takes 1GB
+      instances : "max", // This will spin up 2-3 instances
```
clottman: Why do we have `instances: 1` as well as `instances: max`? I assume the latter one takes priority but might be worth cleaning up for clarity.

keithk: Oops, that was from testing, cleaning up!

ecosystem.config.js
```diff
@@ -0,0 +1,16 @@
...
+      exec_mode : "cluster",
+      env: {
+        NODE_ENV: "development",
```
clottman: I'm not 100% clear on how these variables work - do we need to have all the variables from `.env` listed here?

keithk: Nope, we can also probably remove this, I was using it for testing but we have htis in our .env (which I forgot to port over when I made this remix 💀 )

ecosystem.config.js
```diff
@@ -0,0 +1,16 @@
...
+      },
+      env_production: {
+        NODE_ENV: "production",
+      }
```
GregWeil: How does this interact with `.env`, or will this replace it?

package.json
```diff
@@ -29,7 +30,7 @@
     "lint:server": "eslint --config server/.eslintrc.server.js server shared webpack.config.js",
     "lint:src": "eslint --config src/.eslintrc.js src",
     "prestart": "npm run watch:lint:server & \n if if-env BUILD_TYPE=watcher; then \n npm run watch:babel & npm run watch:webpack & \n fi",
-    "start": "NODE_OPTIONS='--max-old-space-size=4096' forever --minUptime 5000 --spinSleepTime 1000 --killSignal=SIGTERM -c 'nodemon --watch server --watch shared --watch views --exitcrash' ./server/server.js",
+    "start": "pm2-runtime start ecosystem.config.js --env production",
```
clottman: By passing in the env variable here, does this mean all community remixes will be building in production mode by default? Why not use the env variable from .env?

package.json
```diff
@@ -142,10 +143,11 @@
     "url-join": "^4.0.1",
     "url-search-params-polyfill": "^7.0.0",
     "wcag-contrast": "^2.1.1",
-    "webpack": "^4.35.0",
-    "webpack-cli": "^3.3.5",
+    "webpack": "^4.41.2",
```
clottman: Any reason in particular for upgrading webpack? (Just for knowledge sharing's sake). I looked through the releases and see some bug fixes and new features; no breaking changes as expected by semver.

keithk: Mostly just for the bug fixes, I'm going to make a branch later to update our packages to the extent that we can as well

keithk:
Okay, fixed all of the package.json / node ENV stuff

clottman:
With 4 different webpack builds, is there a way to know, without just watching the logs to see them all go by, that all 4 processes are ready?

keithk:
@clottman So it **should** only build once -- the `prestart` in package.json is what runs the build, or should, and then server.js should pick up on those

GregWeil:
If you're in memory mode each instance will run its own build. Personally I'd be in favor of removing memory mode if it doesn't play well with pm2, though

clottman:
@keithk I could be misunderstanding but that's not what I'm seeing. 
image

On my local machine it started up 12 different processes and output a webpack build for each one. 😱 My fans are going crazy.

clottman:
Ahh @GregWeil that must be the problem, when I remixed Keith's project to test it out, I didn't alter the .env file at all. Same with running on my local machine. Should static mode be the default instead maybe?

keithk:
@clottman @GregWeil Yeah, we should change the default -- should it be static or watcher?

ecosystem.config.js
```diff
@@ -0,0 +1,9 @@
+module.exports = {
+    apps : [{
+      name: "glitch-community",
+      script: "./server/server.js",
+      max_memory_restart: "1G", // This will restart gracefully the service if it takes 1GB
+      instances : "max", // This will spin up 2-3 instances
+      exec_mode : "cluster",
+    }]
```
GregWeil: I think this loses nodemon's file system watching. It looks like pm2 does have the capability https://pm2.keymetrics.io/docs/usage/watch-and-restart/

keithk: Cool, added it to this file!

ecosystem.config.js
```diff
@@ -0,0 +1,9 @@
+module.exports = {
+    apps : [{
+      name: "glitch-community",
+      script: "./server/server.js",
+      max_memory_restart: "1G", // This will restart gracefully the service if it takes 1GB
+      instances : "max", // This will spin up 2-3 instances
```
keithk: On a glitch container this is *3* instances

ecosystem.config.js
```diff
@@ -0,0 +1,16 @@
...
+      exec_mode : "cluster",
+      watch: ["server", "shared", "views"],
```
keithk: Should this be watching anything else @GregWeil ?

ecosystem.config.js
clottman: On my local machine it's 12. 😆

server/webpack.js
```diff
@@ -21,7 +21,7 @@ function webpackExpressMiddleware() {
 }
 
 module.exports = function(app) {
-  if (!process.env.BUILD_TYPE || process.env.BUILD_TYPE === 'memory') {
+  if (!process.env.BUILD_TYPE || process.env.BUILD_TYPE === 'memory' && process.env.NODE_ENV !==  'production') {
```
keithk: This should stop it from running memory mode in production

ecosystem.config.js
keithk: @clottman haha yeah, it'll spin up depending on the number of CPUs it has available!

ecosystem.config.js
GregWeil: Nope, that looks like it

server/webpack.js
GregWeil: server/render.js also needs something like this, so babel doesn't run in the same file

package.json
```diff
@@ -21,6 +21,7 @@
     "build": "npm run build:webpack && npm run build:babel",
     "build:babel": "npx babel ./src --config-file ./.babelrc.node.js --copy-files -d ./build/node",
     "build:webpack": "npx webpack",
+    "build:prod": "NODE_ENV=production npx webpack && npm run build:babel",
```
vrk: This might be an obvious question, but when does this `build:prod` script get called?

server/render.js
```diff
@@ -8,7 +8,7 @@ const { getOptimizelyClient, getOptimizelyData } = require('./optimizely');
 const setup = () => {
   const src = path.join(__dirname, '../src');
   const build = path.join(__dirname, '../build/node/');
-  if (!process.env.BUILD_TYPE || process.env.BUILD_TYPE === 'memory') {
+  if (!process.env.BUILD_TYPE || process.env.BUILD_TYPE === 'memory' && process.env.NODE_ENV !== 'production') {
```
vrk: Actually a few more basic questions:
* What are the valid values for `BUILD_TYPE`, and where are they configured? 
* What would be the effect of running this code (or the similarly guarded code below) in production? Would it cause a major performance problem? If so, seems like this is a little scary, even if temporary - would it be easy to add a unit test?
* Should this _ever_ be run in production? If not, should this be `||` instead of `&&`?

ecosystem.config.js
```diff
@@ -0,0 +1,16 @@
+module.exports = {
+    apps : [{
+      name: "glitch-community",
+      script: "./server/server.js",
+      max_memory_restart: "1G", // This will restart gracefully the service if it takes 1GB
+      instances : "max", // This will spin up 2-3 instances
```
vrk: (+1 on your suggestion from Slack to set this explicitly to 2 or 3; seems like it'd be mildly better to have some determinism here)

dluxemburg:
This pull came up in the incident zoom and I think it might be trouble if the underlying problem is overuse of a shared resource in the container. 

Specifically, it will mean making many more outbound requests to the API for projects because the 3 subprocess won't share the in-memory cache. There's a plausible theory that the memory pressure is caused by those backing up and now the container as a whole will be sending more and each process will have greater memory pressure. So if outbound requests or incoming data is restricted at the container level this might make the problem worse.

It's possible that even if the above is true, there won't be enough of an effect to matter and/or juggling the processes as they run out of memory will make it irrelevant and there will always be at least one that's doing okay while the bad ones restart fresh (that one might have a rough time of it though with the lower memory cap).

Not saying we shouldn't try it, just that there are plausible reasons it wouldn't work so if it doesn't we should probably back off quickly rather than expect to necessarily be able to tweak things until it's stable. Also that even if we do get it to work, it might be more sensitive to volume and distribution or traffic than the current version of things so unless we definitely see an improvement we should be ready to roll this back before layering additional fixes on top.

Also! We bumped the size of performance boost from 4 to 6, so once we can confirm that both community projects are running in containers with the new limit we'd be able to set the max memory per process to at least 1.5gb.

keithk:
@dluxemburg At the least we'll be able to see it restarting the processes, if that does end up backing up.

Otherwise, we can set it to just "1" for the number of instances, and have it still restart automatically at the memory limit (4G or something) instead of waiting on the container to restart (which goes way past the time that it becomes unresponsive)

dluxemburg:
Being able to go back to 1 instance and avoid container restarts is a good point. That would definitely avoid the shared resource concern (at least as I was suggesting it).

keithk:
Yup - I think starting with one and bumping it up if we feel confident is good, and just relying on PM2 for the restart capability

ecosystem.config.js
```diff
@@ -0,0 +1,16 @@
+module.exports = {
+    apps : [{
+      name: "glitch-community",
+      script: "./server/server.js",
+      max_memory_restart: "3G", // This will restart gracefully the service if it takes 1GB
```
clottman: This comment is misleading now

package.json
clottman: Bumping this question - I don't see it referenced anywhere.

keithk:
Alright can I get a 👍  from @GregWeil / @clottman 

This now:
- Runs in fork mode (only one instance)
- Has a limit of 3GB, we can patch this to 5GB once we're on a performance boost with a 6GB limit
- Will restart automatically if it hits that memory limit

server/webpack.js
GregWeil: I think this check is broken, if BUILD_TYPE is unset it'll use memory mode even in production

ecosystem.config.js
keithk: Removed!

package.json
GregWeil: Also wondering this, is there a situation where we want to build prod without using env vars?

server/webpack.js
keithk: @GregWeil so it should be:
`if ((!process.env.BUILD_TYPE || process.env.BUILD_TYPE === 'memory') && process.env.NODE_ENV !==  'production')`

right?

keithk: 👍  Updated in webpack & render

PR #1071, opened 2019-11-14
Superb guide: Move ssr build into webpack for faster server startup

GregWeil:
## Links

## Changes:
- Get rid of the separate `babel` call, now ssr is prepared as part of the webpack build
  - Less confusing output: there's a single 'build complete' message
  - The server only reloads once the entire build is done, not every time babel writes out a single file
- Don't watch for file changes when `BUILD_TYPE=static`, so chokidar and require.cache manipulation don't even come into play

## How To Test:
Try running the site with `BUILD_TYPE` set to `memory` `watcher` and `static`. In watcher mode it should only ssr load once on startup and again after the build finishes, rather than on every request until babel finishes.


keithk:
This looks great.

I wonder if this is something that we should step away from trying to do ourselves in the long term, though. Something like [Razzle](https://github.com/jaredpalmer/razzle) which just enables SSR/frontend for any type of application could be a good avenue for us to investigate so that we aren't trying to maintain all of the code to serve our builds.

package.json
```diff
@@ -18,8 +18,7 @@
     "./src/utils/lever/lever.styl"
   ],
   "scripts": {
-    "build": "npm run build:webpack && npm run build:babel",
-    "build:babel": "npx babel ./src --config-file ./.babelrc.node.js --copy-files -d ./build/node",
+    "build": "npm run build:webpack",
```
clottman: Since this is just doing one thing now, can we simplify it to `build: npx webpack` and get rid of the extra script entry? We would need to replace the other uses of that script though too, like in `watch:webpack`

package.json
```diff
@@ -145,6 +142,7 @@
     "webpack": "^4.35.0",
     "webpack-cli": "^3.3.5",
     "webpack-dev-middleware": "^3.7.0",
+    "webpack-node-externals": "^1.7.2",
```
clottman: Why did this need to move? Do we need it in both dependencies and devDependencies (line 162)?

server/render.js
```diff
@@ -5,64 +5,68 @@ const { captureException } = require('@sentry/node');
...
-});
-
-const requireClient = () => {
-  if (needsTranspile) {
-    // remove everything in the src directory
-    Object.keys(require.cache).forEach((location) => {
-      if (location.startsWith(directory)) delete require.cache[location];
+const watch = (location, entry, log) => {
+  let needsReload = true;
+  requireClient = () => {
```
clottman: What are we gaining by making this a mutating function identifier? Are there different cases when each version would get called?

webpack.config.js
```diff
@@ -174,4 +173,64 @@ module.exports = smp.wrap({
...
+              },
+              'stylus-loader',
+            ],
+          },
+        ],
+      },
+    ],
+  },
+  externals: [
+    NodeExternals(),
```
clottman: How were we getting by without this before?

webpack.config.js
```diff
@@ -174,4 +173,64 @@ module.exports = smp.wrap({
...
+    (context, request, callback) => /^Shared[\\/]/.test(request) ? callback(null, `commonjs ${request}`) : callback(),
```
clottman: What's this line doing?

webpack.config.js
```diff
@@ -174,4 +173,64 @@ module.exports = smp.wrap({
...
+          {
+            test: /\.styl/,
+            use: [
+              {
+                loader: 'css-loader',
+                options: {
+                  modules: {
+                    localIdentName: '[name]__[local]___[hash:base64:5]',
+                  },
+                  onlyLocals: true,
```
clottman: I see [the docs for this option](https://github.com/webpack-contrib/css-loader#onlylocals) but I don't fully understand them. Will classes written as `:global` still work?

server/render.js
```diff
@@ -5,64 +5,68 @@ const { captureException } = require('@sentry/node');
...
-  clearCache(); // clear the server rendering cache
-});
-
-const requireClient = () => {
-  if (needsTranspile) {
-    // remove everything in the src directory
-    Object.keys(require.cache).forEach((location) => {
-      if (location.startsWith(directory)) delete require.cache[location];
+const watch = (location, entry, log) => {
+  let needsReload = true;
```
clottman: I'm having a hard time following the logic around setting this/using it. Is there a way we can refactor to make control flow rely less on this variable?

webpack.config.js
GregWeil: The babel cli builder compiled each file separately, so the resulting build was a mirrored src directory but with node compatible code. That meant that node `require('module')` calls (transpiled from `import from 'module'` statements) worked without any special consideration. Webpack writes out a single build/server.js bundle, and by default would pack in all of the external dependencies like it does for the client build.

package.json
GregWeil: Oh, I didn't notice that it was already there. I guess it could stay in `devDependencies` since it's only used in memory/watcher mode and isn't needed for static builds

webpack.config.js
GregWeil: That's telling webpack not to include anything from the shared directory in the bundle. There are many ways to config this, but this was the only way I could get to preserve the call as is (and is pretty much how the node externals lib does it, except matching against node_modules)

GregWeil: https://webpack.js.org/configuration/externals/#function

webpack.config.js
GregWeil: They will still be included in the client build which writes the actual css. They won't be exposed via `import styles from './styles.styl'` calls, but I don't think global classes were anyway since they don't get renamed.

webpack.config.js
clottman: Why should shared not go in the bundle?

GregWeil: If it's included in the bundle all src code will get a different copy of the shared files (the bundled copy) than the server code, which will read the original files. Having two copies of the same module running can cause issues with module-level state being duplicated.

server/render.js
GregWeil: `watch` doesn't get called in static mode, so `requireClient` is just a direct call to require. I'll see how I can improve readability there

server/render.js
GregWeil: Probably, I'll see what I can do

GregWeil: I renamed things and shuffled them around a bit, is this an improvement?

PR #1072, opened 2019-11-15
Don't handle the error twice for invite user

sarahzinger:
## Links
* Remix link: https://joyous-nutria.glitch.me/
* Issue-tracker link: https://app.clubhouse.io/glitch/story/1702/weird-ui-messaging-in-response-to-a-team-invite-failing-due-to-rate-limiting

## GIF/Screenshots:
before:
Screen Shot 2019-11-15 at 3 18 56 PM

after:
Screen Shot 2019-11-15 at 3 19 10 PM

## Changes:
When we run into an error inviting a user to a team we seem to be handling it in 2 places so I'm leaving this one: https://github.com/FogCreek/Glitch-Community/blob/master/src/components/team-users/index.js#L219 and removing the generic one in team.js

## How To Test:
* create a team and invite a test user, then remove them, then invite them again you should hit the error

## Feedback I'm looking for:
should the error handling belong in team.js... some part of me wonders if this solution is too lazy, maybe I should put move all of the errors there. This just seemed easier

### DIFF
--- a/src/state/team.js
+++ b/src/state/team.js
@@ -97,8 +97,8 @@ export function useTeamEditor(initialTeam) {
         updateCurrentUser({ teams: [...teams, team] });
       }
     }, handleError),
-    inviteEmail: (email) => inviteEmailToTeam({ team }, email).catch(handleError),
-    inviteUser: (user) => inviteUserToTeam({ user, team }).catch(handleError),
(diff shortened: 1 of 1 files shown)


sarahzinger:
A change to the backend was recently deployed that returns a 429 again, so I'm gunna try handling that and returning a more descriptive error message.

Going to close this unmerged and reopen again later.

PR #1073, opened 2019-11-15
Fail quietly if there's weird stuff in the address bar after a #

clottman:
## Links
* usefocusfirst.glitch.me
* https://app.clubhouse.io/glitch/story/1617/usefocusfirst-bug

## How To Test:
Compare: [working link in remix with weird URL](https://usefocusfirst.glitch.me/~soulpatch#:~:targetText=Bootstrap%204.0%20Album&targetText=Bootstrap%20is%20the%20world's%20most,of%20the%20official%20Bootstrap%20examples.)

and [broken page on prod with weird URL](https://glitch.com/~soulpatch#:~:targetText=Bootstrap%204.0%20Album&targetText=Bootstrap%20is%20the%20world's%20most,of%20the%20official%20Bootstrap%20examples.)


src/hooks/use-focus-first.js
```diff
@@ -2,13 +2,17 @@ import React from 'react';
...
-      const firstHeading = document.querySelectorAll(`#${hash} h1:first-of-type, #${hash} h2:first-of-type`)[0];
-      if (firstHeading) {
-        firstHeading.setAttribute('tabindex', -1);
-        firstHeading.focus();
+    try {
+      const [, hash] = window.location.href.split('#');
+      if (hash) {
+        const firstHeading = document.querySelectorAll(`#${hash} h1:first-of-type, #${hash} h2:first-of-type`)[0];
+        if (firstHeading) {
+          firstHeading.setAttribute('tabindex', -1);
```
sarahzinger: I see we already had this so not disagreeing with but more just curious: why do we set the tabindex to -1 here?

clottman: Headings aren't normally focusable, but they (like other elements) can be made focusable programmatically if you set tabindex to -1. (-1 also means that they still won't be in the normal tab order, which is what we want.)

sarahzinger: that makes a lot of sense, thanks!!

PR #1077, opened 2019-11-18
Create /delete route in community

sarahzinger:
## Links
* Remix link: https://humane-clam.glitch.me/
* Issue-tracker link: https://app.clubhouse.io/glitch/story/3229/create-a-delete-token-token-route-in-community

## GIF/Screenshots:
Error view:
Screen Shot 2019-11-18 at 3 48 30 PM
Success view:
Screen Shot 2019-11-18 at 3 48 44 PM

## Changes:
Adds a success and error state for when users click the link in their email to delete their account. Right now those backend endpoints don't exist but eventually they'll be able to:
- click a button in community to delete their account
- this sends them an email
- this email contains a link to glitch.com/delete?token=sometoken
- we then use that token to delete their account
- if the token is valid we show them a success page, with a form to explain why they are leaving glitch
- if the token is invalid we show them an error page

Since the backend is not complete I've just worked on the visual aspects.

## How To Test:
* enable deleting on the /secret page
* go to https://humane-clam.glitch.me/delete?token=good to see the success page and any other token to see the error page. When you submit feedback you should be able to see it in segment's debugger. Here's an example of what I see, seems like the string is sanitized to me?:
```
analytics.track('Submitted User Deletion Feedback', {
  reasonForLeaving: 'this is an example of a longer response that we could hypothetically receive however this is not deployed as of yet so you can definitely disregard it if you\'d like. It has not bearing on our current churn rate, I just wanted to see what is a reasonable max length for the input box you know? I\'m thinking 500 characters is plenty to say that a user is deleting their profile on glitch.com because they woke up on the wrong side of the bed, but I don\'t know if it\'s too many characters for segment.'
});
```

## Feedback I'm looking for:
visually how's this looking? Also right now I'm sending why folks are leaving to segment, which seems...ok? I'm worried segment is flakey. I also worry about if segment can handle long response and stuff so I limited the textarea to 500 characters, let me know if that seems ok?

### DIFF
--- a/src/presenters/pages/delete.js
+++ b/src/presenters/pages/delete.js
@@ -3,7 +3,7 @@ import { Helmet } from 'react-helmet-async';
 import Layout from 'Components/layout';
 import { TextArea, Button, Icon } from '@fogcreek/shared-components';
 import Notification from 'Components/notification';
-import { useTracker } from 'State/segment-analytics';
+import { useTracker, useIsAnalyticsInitialized } from 'State/segment-analytics';
 import useDevToggle from 'State/dev-toggles';
 
 import { emoji } from 'Components/global.styl';
@@ -17,6 +17,7 @@ const ValidToken = () => {
     ...inherited,
     reasonForLeaving,
   }));
+  const isInitialized = useIsAnalyticsInitialized();
 
   const submitFeedback = () => {
     setShowNotification(true);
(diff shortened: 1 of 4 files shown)


clottman:
Worth calling out that if we're only sending responses to segment, users with tracking-preventing browsers or browser add-ons likely won't have their submissions recorded at all. Is that okay? @sarahzinger

sarahzinger:
@clottman great point thank you!! Talked it over in the pod channel and I think the thing to do is to see:
-- can we detect if a user is blocking segment, if so lets hide the form (and/or possibly show an email)
-- and if not that then can we detect if a segment event did not succeed and if so show that to the user.

Going to move this back to WIP and try to update accordingly thanks for bringing this to my attention!

sarahzinger:
Now featuring an email for folks blocking segment :) 

Screen Shot 2019-11-22 at 3 12 29 PM

sarahzinger:
@clottman want to take another look for me? ty!

src/presenters/pages/delete.js
```diff
@@ -0,0 +1,95 @@
...
+            <Notification type="success" persistent>
+              Thanks for sharing your thoughts!
+            </Notification>
+          )}
+        </>
+      ) : (
+        <p>
+          If you'd like to share any feedback, feel free to leave us a note at{' '}
+          <Button as="a" href="mailto:support@glitch.com">
+            <img src="https://cdn.glitch.com/99defeae-ef08-4bb7-ba01-385a81949100%2FemailBlack.svg?1537835934697" alt="" width="18" />{' '}
```
clottman: Would the Icon component be able to be used here? Mostly to reduce places we sprinkle the CDN urls in case we change out the emoji later on.

src/presenters/pages/delete.js
```diff
@@ -0,0 +1,95 @@
...
+
+  const submitFeedback = () => {
+    setShowNotification(true);
+    tracker();
+    setReason('');
+  };
+
+  return (
+    <div>
+      <h1>Your account has been deleted</h1>
```
clottman: This is super minor but we use a `.` at the end of the header on the error page ("Sorry, we were unable to delete your account.") and not one here. I don't know what our design/content guidelines would say about what the preferred way is.

PR #1082, opened 2019-11-19
Sponge-Credit: Make experience renaming project smoother, Option 1

sarahzinger:
## Links
* Remix link: https://sponge-credit.glitch.me/
* Issue-tracker link: https://app.clubhouse.io/glitch/story/1700/if-you-edit-the-name-of-a-project-while-on-the-project-page-typing-a-little-bit-slowly-the-focus-is-lost-and-it-correlates

## GIF/Screenshots:
screen recording: https://recordit.co/Wqx0AhIfdg

## Changes:
Before when you would change the name of the project it would make the request to the backend, then update the state throughout the app, which would cause everything to rerender with the new project name, which would steal back focus and make the embed flicker, which was a pretty unpleasant experience.

Now it will make the change to the backend so as to determine if it's an ok name or not, and it will only update the state internally to the input field. On blur we then update the state for the rest of the app which will make the embed flicker etc, but this way it only happens once.

## How To Test:
* use https://sponge-credit.glitch.me/ to change a project name, try typing slowing, make sure it feels smooth to use.

## Feedback I'm looking for:
Please compare with https://github.com/FogCreek/Glitch-Community/pull/1093 to see which we like better!


sarahzinger:
we talked about this in pr walk through, I'm going to try building another version where we don't save until the onblur and then compare the two and see which one feels right

ttseng:
per our discussion:

- we should update the URL as the user types, if the domain does not already exist
- however, we should not update the embed until onBlur, so the embed doesn't keep flashing every time someone is editing the project name

sarahzinger:
@ttseng wanna take another look for me?

PR #1083, opened 2019-11-20
Live changes fix

sarahzinger:
not sure why this made a merge conflict, but I think this is up to date now, it seems to be adding new apps we love


sarahzinger:
not merging this since I reverted the pr

PR #1069, opened 2019-11-14
WIP: Various dart: user deletion modal

whimsicallyson:
in case you feel up for taking a look and noticing anything that would be causing issues. no pressure, getting healthy is more important! (https://various-dart.glitch.me/)

PR #1089, opened 2019-11-21
Fix rerender bug

sarahzinger:
## Links
* Remix link: https://tremendous-handstand.glitch.me/settings

## GIF/Screenshots:
after/before
Screen Shot 2019-11-21 at 2 17 43 PM

## Changes:
While working on: https://github.com/FogCreek/Glitch-Community/pull/1069 we noticed that the /settings route kept re-rendering forever. 

This behavior matches up with what we saw back when we made /settings a conditional route and it actually caused `<Router/>` to re-render indefinitely. We made a change to put the conditional inside the Settings page component, but never really got to the bottom of why re-renders were happening in the first place, we just isolated the issue to a page that only glitch devs are using right now.

@GregWeil and @whimsicallyson and I looked at it again today and realized that when we used the dev toggle and when we had 2 versions of glitch open at one time we could see the re-renders happening. We learned that we have a storage event that fires everytime local storage is updated:
https://github.com/FogCreek/Glitch-Community/blob/master/src/state/local-storage.js#L80
and that that storage event will save a new cached object, even if the old cached object is the same as before:
https://github.com/FogCreek/Glitch-Community/blob/master/src/state/local-storage.js#L70
doing so will then update the value we then pass to our localStorageProvider:
https://github.com/FogCreek/Glitch-Community/blob/master/src/state/local-storage.js#L101
which then causes the whole page to rerender

As it turns out segment regularly updates our localStorage, about once every few seconds are so, and this is expected behavior. So instead we're only interested in updating the cache if it's a key we already know about (not any of the segment event keys for example).

This seems to prevent it from rerendering forever, and yet still seems to update the app across multiple tabs :)

## How To Test:
* It's a bit challenging to see the effects of the lack of forced rerenders without adding a bunch of console.logs (but you could locally!) For example you could console.log("hello") in the settings page and see it rerender forever in a remix of prod, but you'll see it a reasonable amount of times in this remix. I did try using the profiler react-dev tools thing and recorded the results for 30 seconds and show the screenshots up above. You can see the one on the right which is a remix of prod, seems to have many more re-renders than the one on the left which is this remix.
* I think in terms of bugs we'd be interested in seeing if anything else that is relying on this localstorage provider is broken. I'm not entirely sure what is interested in it, but I did test that signing in/out seems to work fine across tabs, as does using different dev toggles on the secret page. If you think of other things to look out for feel free to call them out here/qa them :)

## Feedback I'm looking for:
anything! this is pretty confusing stuff, so feel free to shout if something doesn't look right/make sense, its possible we're misunderstanding stuff, but this does seem to make the bug go away.

### DIFF
--- a/src/state/local-storage.js
+++ b/src/state/local-storage.js
@@ -68,7 +68,7 @@ const LocalStorageProvider = ({ children }) => {
         if (event.key) {
           setCache((oldCache) => {
             // segment updates localstorage but that doesn't need to affect our cache
-            if (event.key.includes("segment")) return oldCache;
+            if (event.key.includes('segment')) return oldCache;
 
             const newCache = new Map(oldCache);
             newCache.delete(event.key);
--- a/src/state/local-storage.js
+++ b/src/state/local-storage.js
@@ -67,6 +67,9 @@ const LocalStorageProvider = ({ children }) => {
       if (event.storageArea === storageRef.current) {
         if (event.key) {
           setCache((oldCache) => {
+            // segment (and maybe other things) could update localstorage, no need to update cache
+            if (!oldCache.has(event.key)) return oldCache;
+
             const newCache = new Map(oldCache);
             newCache.delete(event.key);
             return newCache;
--- a/src/state/local-storage.js
+++ b/src/state/local-storage.js
@@ -67,8 +67,8 @@ const LocalStorageProvider = ({ children }) => {
       if (event.storageArea === storageRef.current) {
         if (event.key) {
           setCache((oldCache) => {
-            // segment (and maybe other things) could update localstorage, no need to update cache
-            if (!oldCache.has(event.key)) return oldCache;
+            // segment updates localstorage but that doesn't need to affect our cache
+            if (event.key.includes("segment")) return oldCache;
 
             const newCache = new Map(oldCache);
             newCache.delete(event.key);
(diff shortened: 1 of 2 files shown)


src/state/local-storage.js
```diff
@@ -67,6 +67,9 @@ const LocalStorageProvider = ({ children }) => {
       if (event.storageArea === storageRef.current) {
         if (event.key) {
           setCache((oldCache) => {
+            // segment (and maybe other things) could update localstorage, no need to update cache
+            if (!oldCache.has(event.key)) return oldCache;
```
clottman: How does oldCache get initialized for a new user to the site? 

I'm imagining a user that has never visited Glitch.com before, after this update goes out - would they ever have a localStorage cache?

sarahzinger:
You know what I don't think we make these segment calls every 7 seconds anymore so I don't see the bug, things seem to rerender a reasonable amount of times now. I suspect it's possible we can improve this behavior but for right now let's not touch it until there's an actual need

sarahzinger:
ok jk this is still happening I can recreate it now lol, gunna investigate

src/state/local-storage.js
GregWeil: It gets initialized as an empty Map on every fresh page load, then is populated as the page reads stuff out of localstorage.

sarahzinger:
@clottman so I edited this to be more conservative for now and only reuse the cache if it's a segment event, but is it possible that what you were running into was that you can't view the /setting page unless you're logged in? https://github.com/FogCreek/Glitch-Community/blob/master/src/presenters/pages/settings.js#L53

PR #1093, opened 2019-11-22
Historical mochi: Renaming Project, Option 2

sarahzinger:
## Links
* Remix link: https://historical-mochi.glitch.me/
* Issue-tracker link: https://app.clubhouse.io/glitch/story/1700/if-you-edit-the-name-of-a-project-while-on-the-project-page-typing-a-little-bit-slowly-the-focus-is-lost-and-it-correlates

## GIF/Screenshots:
screen recording here: https://recordit.co/8YQsa4SG9M

## Changes:
Very similar to https://github.com/FogCreek/Glitch-Community/pull/1082 but in this option we never save anything until blur. We do however verify the validity of the name.

if you blur on a bad name, it goes to the last saved good state.

## How To Test:
* use the remix and try renaming a project

## Feedback I'm looking for:
Which do we like better? The auto save that saves as you type or the save on blur? Either way I think we shouldn't update the rest of the app state until on blur.


sarahzinger:
gunna close this, sounds like we generally prefer option 1

PR #1095, opened 2019-11-25
Pickle suggestion: wake up some projects in a collection while in play mode

sarahzinger:
## Links
* Remix link: https://pickle-suggestion.glitch.me/
* Issue-tracker link: https://app.clubhouse.io/glitch/story/4600/wake-up-projects-a-few-at-a-time

## Changes:
We decided to deploy Exploring Collection without waking up projects since we didn't want to put any additional pressure on infrastructure. That said we looked into the outage on 11-12-19, and we don't think it was related to the exploring collections feature so now that it's out and doing fine, we'd like to deploy a change to wake up some of the projects again so that the embeds transition better from one project to another as you tab through. 

to be extra conservative I've added some logic here to ensure a client only requests to wake up a few projects at a time.

## How To Test:
To test you can go to the remix and look at a collection in play mode and open the network tab and search for requests that contain `glitch.me` in them.
69072849-4d34c780-09fa-11ea-93cf-7c5dc64a5a64

You should see that on every 5th project we wake up 6 projects (something like this:)
69073032-9be26180-09fa-11ea-8470-88a4a96f4a00

We should not re-wake up projects until you refresh the page.

## Feedback I'm looking for:
any reasons not to do this or alternatives folks can think of that they like better?

### DIFF
--- a/src/components/collection/collection-projects-player.js
+++ b/src/components/collection/collection-projects-player.js
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React, { useState, useEffect } from 'react';
 import PropTypes from 'prop-types';
 
 import { withRouter } from 'react-router-dom';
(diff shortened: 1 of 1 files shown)


src/components/collection/collection-projects-player.js
```diff
@@ -140,13 +140,42 @@ PlayerControls.propTypes = {
...
 
+const wakeUpNextBatchOfProjects = (projects) => {
+  projects.map(async (project) => fetch(`https://${project.domain}.glitch.me`, { mode: 'no-cors' }));
+};
 
 const CollectionProjectsPlayer = withRouter(({ history, match, isAuthorized, funcs, collection }) => {
   const { projects } = collection;
   const [currentProjectIndex, setCurrentProjectIndex] = useState(getCurrentProjectIndexFromUrl(match.params.projectId, projects));
-
+  const [wokeProjects, setWokeProjects] = useState({});
```
ttseng: wokeProjects 😂

ttseng:
awesome, this works great!

PR #1098, opened 2019-11-26
Profuse tea: Don't write cms data to the file system

GregWeil:
## Links
https://app.clubhouse.io/glitch/story/4593/prismic-handling-research-implementation

## Changes:
- The home data and pupdates get loaded form the cms on startup, and new updates cause a reload
- Content update requests fall back on the last good data rather than using a generic fallback

## How To Test:
- Try loading the site and make sure all of the content is there
- Throw some garbage code in one of the getters and watch the site not explode or lose content

## Feedback I'm looking for:
- This still expects the curated home/pupdates files to be there, even if the aren't used outside of failure cases. We need some way for them to be updated, either as part of the build process or in the prismic webhook.
- The api updaters discard the POST content and reload from the cms. Do they still need the secret/security?


GregWeil:
TODO: Update cms.glitch.me documentation

server/routes.js
```diff
@@ -162,21 +162,32 @@ module.exports = function(EXTERNAL_ROUTES) {
...
   app.post('/api/:page', async (req, res) => {
     const { page } = req.params;
-    if (!['home', 'pupdates'].includes(page)) return res.sendStatus(400);
-
-    if (req.headers.authorization === process.env.CMS_POST_SECRET) {
-      await writeCuratedContent({ page, data: req.body });
+    if (req.headers.authorization !== process.env.CMS_POST_SECRET) {
+      res.sendStatus(403);
+    } else if (page === 'home') {
+      await reloadHomeData();
```
sarahzinger: hi hi, so when we call reloadHomeData, I see we return the data, but then in this function we never do anything with it right? 

I don't see us storing that anywhere in a cache or something... do we? I feel like I'm missing it

GregWeil: The reload functions store the new content in the `promise` var. The POST /api/home routes don't do anything with the data, though.

sarahzinger: ah now i see it thanks, then when we call getData we get that promise, sorry my eyes missed it!

PR #1100, opened 2019-11-27
Quixotic rocket: Send 'Experiment Viewed' events to segment

GregWeil:
## Links
https://app.clubhouse.io/glitch/story/4829/add-an-experiment-viewed-event-for-optimizely-rollouts
https://app.clubhouse.io/glitch/story/2956/error-securityerror-the-operation-is-insecure

## Changes:
- Checking for an optimizely value sends an 'Experiment Viewed' event to segment with info about the feature and status
- Cleaned up team analytics a very small amount so it only reads from optimizely once
- Changed the secret page so it doesn't trigger tracking events
- Block optimizely from sending yet another error message to sentry

## How To Test:
Load up a team page you are not a member of and notice that there is no track request. Load up a team page you are a member of, and there should be a request with info about whether analytics are enabled

### DIFF
--- a/src/state/rollouts.js
+++ b/src/state/rollouts.js
@@ -4,6 +4,11 @@ import { useTracker } from './segment-analytics';
 import { useCurrentUser } from './current-user';
 import useUserPref from './user-prefs';
 
+// human readable rollout state to include in analytics
+const DEFAULT_DESCRIPTION = {
+  true: ['variant', 'showing the new form'],
+  false: ['control', 'showing the original form'],
+};
 const ROLLOUT_DESCRIPTIONS = {
   analytics: {
     true: ['enabled', 'analytics are shown on team pages'],
@@ -17,10 +22,6 @@ const ROLLOUT_DESCRIPTIONS = {
     true: ['swapped', 'create is shown at the index'],
     false: ['control', 'see the existing homepage'],
   },
-};
-const DEFAULT_DESCRIPTION = {
-  true: ['variant', 'showing the new form'],
-  false: ['control', 'showing the original form'],
 };
 
 const Context = createContext();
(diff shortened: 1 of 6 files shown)


src/state/rollouts.js
```diff
@@ -41,7 +62,26 @@ export const useFeatureEnabledForEntity = (whichToggle, entityId) => {
...
-  return overrides[whichToggle] !== undefined ? !!overrides[whichToggle] : raw;
+  const enabled = overrides[whichToggle] !== undefined ? !!overrides[whichToggle] : raw;
+
+  const track = useTracker('Experiment Viewed');
+  const { id } = useOptimizelyValue(
+    (optimizely) => optimizely.projectConfigManager.getConfig().featureKeyMap[whichToggle],
+    [whichToggle],
+  );
+  useEffect(() => {
+    console.log(id, whichToggle, enabled);
```
clottman: Did you want this left in here?

src/state/segment-analytics.js
```diff
@@ -40,9 +43,9 @@ AnalyticsContext.defaultProps = {
 
 export const useTracker = (name, properties, context) => {
   const inherited = React.useContext(Context);
-  return () => {
+  return (callProperties) => {
     try {
-      analytics.track(name, resolveProperties(properties, inherited.properties), resolveProperties(context, inherited.context));
+      analytics.track(name, resolveProperties(callProperties, properties, inherited.properties), resolveProperties(context, inherited.context));
```
clottman: Do you have a preference between this and 
```
export const useTracker = (name, properties, context) => (extraProperties) => {
  try {
    analytics.track(name, resolveProperties(extraProperties, properties), context);
``` 
, which leaves the resolveProperties function alone? 

I need to accomplish the same thing in a PR for the P1 analytics events I'm about to open.

src/state/rollouts.js
GregWeil: I did not!

src/state/segment-analytics.js
GregWeil: Not really, this makes the call a bit cleaner which is nice but recursive calls are not great

PR #1101, opened 2019-11-27
WIP: Polar-aura: P1 Analytics Events

clottman:
## Links
* polar-aura.glitch.me
* https://app.clubhouse.io/glitch/story/4563/community-p1-events

## Changes:
* Connected to the WebDev source in Segment
* Includes events listed as P1 from the spreadsheet
* Removes all the old events
* Removes AnalyticsContext as the new analytics schema doesn't require this nor the origin property
* Renames TrackedButtonGroup to just ButtonGroup because we don't need to track the things the buttons were doing any more

## How To Test:
* Refer to the spreadsheet linked in the clubhouse ticket. Dedicated QA time will be held next week. 

## Feedback I'm looking for:
- Did the assumptions I made here match the intended functionality of the events? I'm worried some context got lost in the discussion around what various properties are supposed to represent. 

- [x] Change the source for events back to the prod source
- [x] Add in the CTA events (still waiting on a list of what the relevant calls to action are)


src/presenters/pages/router.js
```diff
@@ -47,9 +47,18 @@ function ExternalPageReloader() {
...
 function track() {
   try {
-    const { analytics } = window;
+    const { analytics, location } = window;
     if (analytics) {
-      analytics.page({}, { groupId: '0' });
+      const searchParams = new URLSearchParams(location.search);
+      const properties = {
+        search: location.search.slice(1),
+        utmMedium: searchParams.get('utm_medium'),
```
sheridanvk: I noticed that utm fields are automatically collected by Segment https://segment.com/docs/connections/spec/common/#context-fields-automatically-collected (you can see for yourself by passing in utm_campaign, etc, and look at the page event that gets sent by Segment; it includes all the info in the campaign section of the request body). It seems that these are also passed along to e.g. RedShift and such https://segment.com/docs/connections/sources/catalog/libraries/website/javascript/#context--traits. Do you think we could test this to see if we don't need to add this additional code and Segment just takes care of it for us?

clottman: Good catch - removed the manual setters.

src/presenters/pages/create.js
```diff
@@ -343,6 +354,15 @@ function ScreencapSection({ title, description, video, smallVideos, blob, image,
 }
 
 function Help() {
+  const { location } = useGlobals();
+  const helpCenterHref = 'https://glitch.com/help';
```
sarahzinger: unrelated but maybe this should be a relative link

clottman:
The shrinkwrap.yaml.swp file is junk, I think it's created when you try to open shrinkwrap.yaml in the glitch editor in a console-based text editor while the container is writing to that file? 

User ID is set because we called `window.analytics.identify` in current-user.js. 

The `analytics.page` stuff is kinda black-boxy to me; [the section on Default Properties (scroll down) in the docs](https://segment.com/docs/connections/sources/catalog/libraries/website/javascript/#page) talks about what properties are set by default.

sarahzinger:
cool :) thanks for the explanation!

src/models/project.js
```diff
@@ -74,3 +74,22 @@ export function userIsOnlyProjectAdmin({ project, user }) {
   if (adminCount > 1) return false;
   return userIsProjectAdmin({ project, user });
 }
+
+// To be kept in sync with source/data/base-project/domains.js in the editor repo
+const baseProjectIds = Object.freeze([
```
sarahzinger: TIL about .freeze thanks!

views/index.ejs
```diff
@@ -138,8 +138,11 @@
         // js failed to parse, odds are the css would be broken too
         showFallback('noscript');
       }
-      // TODO change me before deploying 
-      analytics.load("Oa74eP053IozA1ouL7my4svHW36qGGmM");
+      if (window.location.hostname === "glitch.com") {
+        analytics.load("N8NetCAuTbHYGqtQtb44FFBSPUXwsAu2"); /* WebProd source in Segment */
+      } else {
+        analytics.load("Oa74eP053IozA1ouL7my4svHW36qGGmM"); /* WebDev source in Segment */
```
sarahzinger: this is cool! so from now on we should look at dev segment for remixes yeah?

sarahzinger: it looks like hostname is not supported on opera https://developer.mozilla.org/en-US/docs/Web/API/Location but maybe that's fine? or we could use .host? seems not that important probably

clottman: Yep dev segment will have the data for remixes and localhost. 

Great catch on hostname not being supported, I looked at that page but didn't see it :( Unfortunately it looks like host isn't supported either. How would you feel about defaulting to the prod source if `!window.location.hostname` in addition to the current condition? I don't think we'll ever be doing development on opera, and if we are I think analytics impact will be minimal.

src/models/project.js
clottman: All thanks to Sheridan, I copied from the editor side. :)

views/index.ejs
sarahzinger: oh dang that browser compatibility chart on mozilla is not consistent weird! on the location page it says host is supported, but on the host page it says it isn't :(

yeah your solution sounds perfect to me!

PR #1106, opened 2019-12-03
State updates for user deletion

whimsicallyson:
## Links
* https://cookie-ricotta.glitch.me/
* https://app.clubhouse.io/glitch/story/3863/team-ownership-transfer-highlights-and-links-out-to-which-teams-have-multiple-members-but-only-one-admin-step-2a
* https://app.clubhouse.io/glitch/story/3864/project-ownership-transfer-highlights-and-links-out-to-which-projects-have-multiple-members-but-only-one-admin-step-2a
* https://docs.google.com/document/d/17O0aLlf0znuCN4-LuVihAkGzsUtdFpv4uV7dQpouvWI/edit#heading=h.c3jzht6x35ee (tech spec)

## Changes:
* In order to have the delete account modal properly know what projects and teams are preventing account deletion, we need to update currentUser when changes happen. Now we do!

## How To Test:
* throw some `console.log`s in to make sure the currentUser updates with the proper info on a project or team delete, or a admin change. 
* make sure that the project/team has properly deleted or the admin has changed

things that have changed (but should still work the same from a user POV):
  * Changing the admin of a project
  * Deleting a project
  * Changing the admin status of another person on a team
  * Changing the admin status of yourself on a team (only possible if there are other admins left)
  * Deleting a team

## Feedback I'm looking for:
um, anything you've got! 
if you'd like to reference it, #1069 is the oversized "everything in once place" WIP PR that this is pulled from

### DIFF
--- a/src/state/project.js
+++ b/src/state/project.js
@@ -5,6 +5,7 @@ import { useAPI, useAPIHandlers } from 'State/api';
 import useErrorHandlers from 'State/error-handlers';
 import * as assets from 'Utils/assets';
 import { getAllPages } from 'Shared/api';
+import { useCurrentUser } from 'State/current-user';
 import { MEMBER_ACCESS_LEVEL, ADMIN_ACCESS_LEVEL } from 'Models/project';
 
 async function getMembers(api, projectId, withCacheBust) {
@@ -93,6 +94,7 @@ export function useProjectReload() {
 }
 
 export function useProjectEditor(initialProject) {
+  const { currentUser, update: updateCurrentUser } = useCurrentUser();
   const [project, setProject] = useState(initialProject);
   const { uploadAsset } = useUploader();
   const { handleError, handleErrorForInput, handleImageUploadError } = useErrorHandlers();
@@ -106,14 +108,26 @@ export function useProjectEditor(initialProject) {
       ...prev,
       ...changes,
     }));
+    updateCurrentUser({
+      projects: currentUser.projects.map((p) => {
+        if (p.id === project.id) {
+          return { ...p, ...changes };
+        }
+        return p;
+      }),
+    });
   }
 
   const withErrorHandler = (fn, handler) => (...args) => fn(...args).catch(handler);
 
   const funcs = {
-    deleteProject: () => deleteItem({ project }).catch(handleError),
+    deleteProject: () => {
+      const projects = currentUser.projects.filter((p) => p.id !== project.id);
+      updateCurrentUser({ projects });
+      return deleteItem({ project }).catch(handleError);
+    },
     updateDomainBackend: withErrorHandler(async (domain) => {
-      await updateItem({ project }, { domain });
+      await updateFields({ domain });
       // don't await this because the project domain has already changed and I don't want to delay other things updating
       updateProjectDomain({ project });
     }, handleErrorForInput),
@@ -140,7 +154,7 @@ export function useProjectEditor(initialProject) {
           }));
         }, handleImageUploadError),
       ),
-    reassignAdmin: async ({ currentUser, user }) => {
+    reassignAdmin: async ({ user }) => {
       await updateProjectMemberAccessLevel({ project, user, accessLevel: ADMIN_ACCESS_LEVEL });
       await updateProjectMemberAccessLevel({ project, user: currentUser, accessLevel: MEMBER_ACCESS_LEVEL });
       const newPermissions = [...project.permissions].map((permission) => {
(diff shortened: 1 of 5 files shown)


sarahzinger:
Hey bud, nice work breaking up the giant pr into smaller ones :)

I think next step is to put this on a remix so someone can qa it, and try to list out all the things a person should try doing to see that these state changes working as expected so that someone reviewing this pr knows what to test (ex: delete a team, etc)

src/state/project.js
```diff
@@ -106,14 +108,26 @@ export function useProjectEditor(initialProject) {
...
   }
 
   const withErrorHandler = (fn, handler) => (...args) => fn(...args).catch(handler);
 
   const funcs = {
-    deleteProject: () => deleteItem({ project }).catch(handleError),
+    deleteProject: () => {
+      const projects = currentUser.projects.filter((p) => p.id !== project.id);
+      updateCurrentUser({ projects });
+      return deleteItem({ project }).catch(handleError);
```
sarahzinger: as i look at this i wonder if we should make the backend update first and then if that's successful update state whatcha think?

src/state/project.js
```diff
@@ -106,14 +108,26 @@ export function useProjectEditor(initialProject) {
...
+    },
     updateDomainBackend: withErrorHandler(async (domain) => {
-      await updateItem({ project }, { domain });
+      await updateFields({ domain });
```
sarahzinger: do we have to have this change or was this a funky git merge? I notice it brings back this bug: https://app.clubhouse.io/glitch/story/1700/if-you-edit-the-name-of-a-project-while-on-the-project-page-typing-a-little-bit-slowly-the-focus-is-lost-and-it-correlates

src/state/team.js
```diff
@@ -82,6 +93,11 @@ export function useTeamEditor(initialTeam) {
   const withErrorHandler = (fn, handler) => (...args) => fn(...args).catch(handler);
 
   const funcs = {
+    deleteTeam: () => {
+      const teams = currentUser.teams.filter((t) => t.id !== team.id);
+      updateCurrentUser({ teams });
+      return deleteItem({ team }).catch(handleError);
```
sarahzinger: similarly here i wonder if we should await the deleteItem and then once the backend is done then update state

src/state/project.js
whimsicallyson: yup we can change that back!

src/state/project.js
whimsicallyson: like this?

src/state/project.js
```diff
@@ -123,11 +123,17 @@ export function useProjectEditor(initialProject) {
   const funcs = {
     deleteProject: () => {
       const projects = currentUser.projects.filter((p) => p.id !== project.id);
-      updateCurrentUser({ projects });
-      return deleteItem({ project }).catch(handleError);
+      const deleteStatus = deleteItem({ project }).catch(handleError).then(
```
sarahzinger: interesting! I'm not 100% sure that would work but it might :) Can you test it to see for sure?

I think in the way you've written it .catch would handle an error and then it would continue on to the .then as if it had succeeded, and instead of returning a response it would return an error. But maybe the error doesn't have a .data property so it would work lol. 

If you want to read more about .catch and .then I'd 100% recommend checking out https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/catch

It might be a bit simpler to do something like this:
```
deleteProject: withErrorHandler(async() => {
      await deleteItem({ project });
      const projects = currentUser.projects.filter((p) => p.id !== project.id);
      updateCurrentUser({ projects });
      },handleError)
```
I believe in this instance if deleteItem returns an error, it will throw and not move forward with the rest of the code, and instead get caught in handleError. Or at least that's a pattern I'm seeing elsewhere in our code!

But let me know if you can test and see! I've been very curious about our error handling as I don't think we've quite settled on a pattern that works well and would be interested to see what you learn! Happy to pair on this if you want too

src/state/project.js
```diff
@@ -106,14 +108,26 @@ export function useProjectEditor(initialProject) {
...
 
   const funcs = {
-    deleteProject: () => deleteItem({ project }).catch(handleError),
+    deleteProject: withErrorHandler(async () => {
+      await deleteItem({ project });
+      const projects = currentUser.projects.filter((p) => p.id !== project.id);
+      updateCurrentUser({ projects });
+    }, handleError),
     updateDomainBackend: withErrorHandler(async (domain) => {
-      await updateItem({ project }, { domain });
```
sarahzinger: lets add back this await like we talked about :)

sarahzinger:
hmm getting an error when i try to delete a team, i get: `deleteTeam TypeError: funcs.deleteItem is not a function`

sarahzinger:
hey hey mind updating cookie-ricotta with the latest? I'm still getting the same error

PR #1107, opened 2019-12-03
Bring back dividers and private backgrounds on results lists

sarahzinger:
## Links
* Remix link: https://special-hen.glitch.me/
* Issue-tracker link: https://app.clubhouse.io/glitch/story/5041/dividers-on-results-lists-have-gone-missing

## GIF/Screenshots:
before/after
Screen Shot 2019-12-03 at 2 13 44 PM

## Changes:
* adds back some dividers and privacy backgrounds on resultslists.

## How To Test:
* try adding a project to a collection via the project-options dropdown
* try adding a project to a collection on the collection page with the Add Project button
* try adding a project to a team
* try filtering by project on team analytics

## Feedback I'm looking for:
anything


clottman:
Shouldn't the fix for the borders be in the shared component? I know @ehmorris was working on border colors in the shared component for ResultItems previously and was going to come back to that work at some point.

ehmorris:
I agree with @clottman that we should add the borders back in shared-components. I just looked at my PR again because I don't remember intending to remove the borders. I think I accidentally broke them with this change (on this line): https://github.com/FogCreek/shared-components/pull/66/files#diff-aaa6a56963e3b44fefcb1c6a66a2771eR145

I can see in my screenshots that the borders were missing, so I didn't notice, but I think a one-character fix will add them back everywhere, and I don't have anything against borders for RowItems.

sarahzinger:
cool makes sense! will close this out and make the changes necessary in shared-components

PR #1108, opened 2019-12-03
Quixotic-rocket: Ignore the cached user if it doesn't match the signed in user

GregWeil:
## Links
https://app.clubhouse.io/glitch/story/5179/the-community-site-flickers-when-logged-out-via-editor

## Changes:
Check that the cachedUser matches the sharedUser before using it

## How To Test:
Sign into the remix, wait for it to fully load, then run `localStorage.removeItem('cachedUser')` and refresh the page. On glitch.com it will show you as signed in for a moment then correct to signed out. On this remix it won't show the signed out user, but it will still server render the signed in homepage.

### DIFF
--- a/src/state/current-user.js
+++ b/src/state/current-user.js
@@ -15,16 +15,6 @@ const getStorageMemo = memoize(getStorage);
 const getFromStorage = (key) => readFromStorage(getStorageMemo(), key);
 const setStorage = (key, value) => writeToStorage(getStorageMemo(), key, value);
 
-function setCookie(name, value) {
-  if (value) {
-    const expires = new Date();
-    expires.setFullYear(expires.getFullYear() + 1);
-    document.cookie = `${name}=true; path=/; expires=${expires}`;
-  } else {
-    document.cookie = `${name}=; path=/; expires=${new Date()}`;
-  }
-}
-
 function identifyUser(user) {
   if (user) {
     addBreadcrumb({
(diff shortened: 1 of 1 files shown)


GregWeil:
Naming refresher because it's downright misleading here (sorry 😢)
`sharedUser` is stored as `cachedUser` in localstorage, ad is the source of truth shared with the editor
`cachedUser` is stored as `community-cachedUser` and is used when rendering the community site

Maybe an extra q is should I expand this tiny fix into a lot of changes to rename the whole file to match the local storage keys?

sarahzinger:
@GregWeil haha yes was just leaving a comment about this in another pr. if you don't mind I think renaming stuff so we're consistent is worth it, also if you wouldn't mind leaving like a quick glossary either in the readme or in a comment next to our first definition of each type of user that would be 💯 I'd give you many bravos! :)

sarahzinger:
Hmm seeing weird behavior on this:
- I login
- I run removeItem
- I don't see any changes
- I refresh page I'm not logged in as an anonymous user, when I click on my grey avatar it takes me to /user/0 and it never loads

GregWeil:
I think you caught me right in the middle of swapping cachedUser and sharedUser, could you try again?

sarahzinger:
hmm yeah it's different now, though I still see a flash of an anonymous user: https://recordit.co/8q4LYGmAxf but I see a similar but different thing on prod so I guess it doesn't matter

src/state/current-user.js
```diff
@@ -125,23 +120,23 @@ async function getCachedUser(sharedUser) {
...
 };
 
-// sharedUser syncs with the editor and is authoritative on id and persistentToken
-const sharedUserKey = 'cachedUser';
-// cachedUser mirrors the v1 API data and is what we actually display
-const cachedUserKey = 'community-cachedUser';
+// cachedUser syncs with the editor and is authoritative on id and persistentToken
+const cachedUserKey = 'cachedUser';
+// communityCachedUser mirrors the v1 API data and is what we actually display
+const communityCachedUserKey = 'community-cachedUser';
```
sarahzinger: ty for these comments they are very helpful!

GregWeil:
> 
> 
> hmm yeah it's different now, though I still see a flash of an anonymous user: https://recordit.co/8q4LYGmAxf but I see a similar but different thing on prod so I guess it doesn't matter

That happens because the server relies on the `hasLogin` cookie to tell if you're signed in or not, but it doesn't get updated until you refresh. (It's updated right away if you use the sign out button)

PR #1111, opened 2019-12-05
Astonishing orchestra: Projects modal pane

whimsicallyson:
## Links
* https://astonishing-orchestra.glitch.me/
* https://app.clubhouse.io/glitch/story/3864/project-ownership-transfer-highlights-and-links-out-to-which-projects-have-multiple-members-but-only-one-admin-step-2a
https://docs.google.com/document/d/17O0aLlf0znuCN4-LuVihAkGzsUtdFpv4uV7dQpouvWI/edit#heading=h.c3jzht6x35ee (tech spec)

## GIF/Screenshots:
Screen Shot 2019-12-05 at 2 40 05 PM
Screen Shot 2019-12-05 at 2 44 18 PM

## Changes:
* Adds "isALink" prop to project-result-item so that clicking on results will take you to the project for updating
* Looks for projects where you're the only admin and there are multiple users, and not letting you continue till they've been fixed (as deleting your account would leave these in a bad state)
* Updates list automatically when project changes are made, no need for a refresh button!

## How To Test:
* Be logged in, turn on Account Deletion and at least one other toggle
* Go to /settings, hit Delete Account, Continue to Delete Account
* See the projects you need to take action on, try changing admins and/or deleting, make sure they disappear

## Feedback I'm looking for:
whatever you've got 😄 

* not a blocker to deploying (since its behind a dev toggle) but we should have the help links go somewhere.

### DIFF
--- a/src/components/project/project-result-item.js
+++ b/src/components/project/project-result-item.js
@@ -24,22 +24,25 @@ const ProfileListWrap = ({ project, asLinks }) => (
   </div>
 );
 
-const ProjectResultItem = ({ project, onClick, profileListAsLinks, buttonProps }) => (
-  <ResultItem className={classnames(project.private && styles.private)} onClick={() => onClick(project)} {...buttonProps}>
-    <div>
-      <ProjectAvatar project={project} />
-    </div>
-    <ResultInfo>
-      <ResultName>{project.domain}</ResultName>
-      {project.description.length > 0 && (
-        <ResultDescription>
-          <Markdown renderAsPlaintext>{project.description}</Markdown>
-        </ResultDescription>
-      )}
-      <ProfileListWrap project={project} asLinks={profileListAsLinks} />
-    </ResultInfo>
-  </ResultItem>
-);
+const ProjectResultItem = ({ project, onClick, profileListAsLinks, buttonProps, isALink }) => {
+  const linkProps = isALink ? { as: 'a', href: `/~${project.domain}`, target: '_blank' } : {};
+  return (
+    <ResultItem className={classnames(project.private && styles.private)} onClick={() => onClick(project)} {...buttonProps} {...linkProps}>
+      <div>
+        <ProjectAvatar project={project} />
+      </div>
+      <ResultInfo>
+        <ResultName>{project.domain}</ResultName>
+        {project.description.length > 0 && (
+          <ResultDescription>
+            <Markdown renderAsPlaintext>{project.description}</Markdown>
+          </ResultDescription>
+        )}
+        <ProfileListWrap project={project} asLinks={profileListAsLinks} />
+      </ResultInfo>
+    </ResultItem>
+  );
+};
 
 ProjectResultItem.propTypes = {
   project: PropTypes.shape({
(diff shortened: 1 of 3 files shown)


src/components/project/project-result-item.styl
```diff
@@ -1,6 +1,6 @@
 @import '../constants.styl'
 
-.private
+.private, a.private
```
clottman: Was this necessary? `.private` should select a superset of what `a.private` selects.

src/components/project/project-result-item.js
```diff
@@ -24,22 +24,25 @@ const ProfileListWrap = ({ project, asLinks }) => (
...
-        <ResultDescription>
-          <Markdown renderAsPlaintext>{project.description}</Markdown>
-        </ResultDescription>
-      )}
-      <ProfileListWrap project={project} asLinks={profileListAsLinks} />
-    </ResultInfo>
-  </ResultItem>
-);
+const ProjectResultItem = ({ project, onClick, profileListAsLinks, buttonProps, isALink }) => {
+  const linkProps = isALink ? { as: 'a', href: `/~${project.domain}`, target: '_blank' } : {};
```
clottman: We should probably use the `getProjectLink` function from models/project.js (instead of hardcoding `/~project.domain`) so we're not duplicating logic that could someday change.

src/components/project/project-result-item.js
```diff
@@ -50,10 +53,12 @@ ProjectResultItem.propTypes = {
   }).isRequired,
   onClick: PropTypes.func.isRequired,
   profileListAsLinks: PropTypes.bool,
+  isALink: PropTypes.bool,
 };
 
 ProjectResultItem.defaultProps = {
   profileListAsLinks: true,
+  isALink: false,
```
clottman: It feels kinda weird to me to add this special-case prop. 

What if instead we left the ProjectResultItem component alone, and passed in a function like this to the onClick prop in DeleteAccountModal?
```
<ProjectResultItem ... onClick={() => window.open(`${domain}/${getProjectUrl(project)}`, '_blank')}>
```

src/components/project/project-result-item.styl
whimsicallyson: yes, its another specificity issue with shared-components :(

src/components/project/project-result-item.js
whimsicallyson: isn't this bad for a11y, since its a `<button>` functioning as an `<a>`? (I also haven't gotten it to work yet, but that could be overcome)

clottman: It's not great but I think that concern is probably overriden by the fact that it looks like a button anyways (or at least, it's not clear that it's a link). As it is, we have a mix of behaviors across the site for buttons that have the same UI - some are buttons, some are links.

src/components/project/project-result-item.styl
sarahzinger: this will not be necessary in the future! https://github.com/FogCreek/shared-components/pull/69#issuecomment-561737209

we just need to upgrade the shared-components lib in the community site, I haven't done that because I figured tiff was still testing her popover changes

sarahzinger: oh btw if this is merged: https://github.com/FogCreek/Glitch-Community/pull/1116 we can remove this private stuff and add in the new isPrivate props to resultItem and resultname

PR #1112, opened 2019-12-05
Debonair quokka: Teams modal pane

whimsicallyson:
## Links
* https://debonair-quokka.glitch.me/
* https://app.clubhouse.io/glitch/story/3863/team-ownership-transfer-highlights-and-links-out-to-which-teams-have-multiple-members-but-only-one-admin-step-2a
https://docs.google.com/document/d/17O0aLlf0znuCN4-LuVihAkGzsUtdFpv4uV7dQpouvWI/edit#heading=h.c3jzht6x35ee (tech spec)

## GIF/Screenshots:
Screen Shot 2019-12-05 at 3 01 01 PM
Screen Shot 2019-12-05 at 3 01 27 PM

## Changes:
* Creates a team-result-item component so that teams can be shown in a ResultsList
* Looks for teams where you're the only admin and there are multiple users, and not letting you continue till they've been fixed (as deleting your account would leave these in a bad state)
* Updates list automatically when project changes are made, no need for a refresh button!

## How To Test:
* Be logged in, turn on Account Deletion and at least one other toggle
* Go to /settings, hit Delete Account, Continue to Delete Account, Continue to Delete Account (again)
* See the teams you need to take action on, try changing admins and/or deleting, make sure they disappear

## Feedback I'm looking for:
whatever you've got 😄

* not a blocker to deploying (since its behind a dev toggle) but we should have the help links go somewhere.

### DIFF
--- a/src/components/delete-account/delete-account-modal.js
+++ b/src/components/delete-account/delete-account-modal.js
@@ -1,21 +1,15 @@
 import React, { useState, useEffect } from 'react';
-<<<<<<< HEAD
 import PropTypes from 'prop-types';
-=======
->>>>>>> 895988ced4dec94b70885c16b67d0d0bf5df1b5f
 import Pluralize from 'react-pluralize';
 import { Actions, Badge, Button, DangerZone, Icon, Info, Overlay, ResultsList, Title, useOverlay, mergeRefs } from '@fogcreek/shared-components';
 
 import { useCurrentUser } from 'State/current-user';
 
 import Link from 'Components/link';
-<<<<<<< HEAD
 import TeamResultItem from 'Components/team/team-result-item';
-import { getTeamLink } from 'Models/team';
-=======
 import ProjectResultItem from 'Components/project/project-result-item';
+import { getTeamLink } from 'Models/team';
 import { getProjectLink } from 'Models/project';
->>>>>>> 895988ced4dec94b70885c16b67d0d0bf5df1b5f
 import MultiPage from '../layout/multi-page';
 
 import styles from './delete-account-modal.styl';
@@ -125,24 +119,6 @@ TeamTransfer.propTypes = {
   last: PropTypes.object.isRequired,
 };
 
-<<<<<<< HEAD
-const ProjectTransfer = ({ setPage, onClose, first, focusedOnMount, last }) => (
-  <>
-    <Title>Transfer Project Ownership</Title>
-    <Info>
-      You must <Link ref={mergeRefs(first, focusedOnMount)} to="/">transfer ownership</Link> or <Link to="/">delete</Link> these projects before you can delete your account.
-    </Info>
-    <Actions>
-      <Button className={styles.actionButton} onClick={() => setPage('teamOwnerTransfer')}>
-        Continue to Delete Account
-      </Button>
-      <Button className={styles.actionButton} variant="secondary" onClick={onClose} ref={last}>
-        Close
-      </Button>
-    </Actions>
-  </>
-);
-=======
 const ProjectTransfer = ({ setPage, onClose, first, focusedOnMount, last }) => {
   const { currentUser } = useCurrentUser();
   const [currentlyFocusedProject, onSelectProject] = useState(null);
@@ -191,7 +167,6 @@ const ProjectTransfer = ({ setPage, onClose, first, focusedOnMount, last }) => {
     </>
   );
 };
->>>>>>> 895988ced4dec94b70885c16b67d0d0bf5df1b5f
 
 ProjectTransfer.propTypes = {
   setPage: PropTypes.func.isRequired,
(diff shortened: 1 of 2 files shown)


src/components/team/team-result-item.js
```diff
@@ -0,0 +1,57 @@
...
+  );
+};
+
+TeamResultItem.propTypes = {
+  team: PropTypes.shape({
+    id: PropTypes.number.isRequired,
+    name: PropTypes.string.isRequired,
+    description: PropTypes.string.isRequired,
+  }).isRequired,
+  onClick: PropTypes.func.isRequired,
```
sarahzinger: looks like you're missing a few proptypes

src/components/delete-account/delete-account-modal.js
```diff
@@ -35,22 +36,66 @@ const DeleteInfo = ({ setPage, onClose, first, focusedOnMount, last }) => (
...
+        <Button disabled={singleAdminTeams.length > 0} className={styles.actionButton} onClick={() => setPage('emailConfirm')}>
+          Continue to Delete Account
+        </Button>
+        <Button className={styles.actionButton} variant="secondary" ref={last} onClick={onClose}>
+          Close
+        </Button>
+      </Actions>
+    </>
+  );
+};
```
sarahzinger: want to add proptypes for this baby too?

src/components/delete-account/delete-account-modal.js
```diff
@@ -35,22 +36,66 @@ const DeleteInfo = ({ setPage, onClose, first, focusedOnMount, last }) => (
...
+        </ResultsList>
+      ) : (
+        <Actions>
+          <Icon className={emoji} icon="victoryHand" />
+          All Done
+        </Actions>
+      )}
+      <Info className={styles.remaining}>
+        <p>
+          <Badge>{singleAdminTeams.length}</Badge> teams to update
```
sarahzinger: should we use that plauralize thingy so that it never says `1 teams to update` ?

src/components/delete-account/delete-account-modal.js
```diff
@@ -35,22 +36,66 @@ const DeleteInfo = ({ setPage, onClose, first, focusedOnMount, last }) => (
...
+        </p>
+      </Info>
+      <Info>
+        If you delete your account, you will automatically be removed from these teams:{' '}
```
sarahzinger: does this `{' '}` do anything for you?
When I look at it I see:
Screen Shot 2019-12-06 at 10 14 32 AM

It might make sense to put other teams into a span and style the span instead

whimsicallyson: that's probably because I forgot to pull the change into the remix 🤦‍♀

src/components/team/team-result-item.js
```diff
@@ -0,0 +1,57 @@
...
+  <div>
+    <VisibilityContainer>
+      {({ wasEverVisible }) =>
+        wasEverVisible ? <ProfileListWithData team={team} asLinks={asLinks} /> : <ProfileList layout="row" size="small" asLinks={asLinks} />
+      }
+    </VisibilityContainer>
+  </div>
+);
+
+const TeamResultItem = ({ team, onClick, buttonProps, profileListAsLinks }) => (
```
whimsicallyson: I updated this to match Cassey's suggestion about not having a special-case prop from #1111

src/components/delete-account/delete-account-modal.js
```diff
@@ -35,22 +38,82 @@ const DeleteInfo = ({ setPage, onClose, first, focusedOnMount, last }) => (
...
-      </Button>
-      <Button className={styles.actionButton} variant="secondary" ref={last} onClick={onClose}>
-        Close
-      </Button>
-    </Actions>
-  </>
-);
+
+DeleteInfo.propTypes = {
+  setPage: PropTypes.string.isRequired,
```
sarahzinger: PropTypes.func.isRequired

src/components/delete-account/delete-account-modal.js
```diff
@@ -69,6 +132,14 @@ const ProjectTransfer = ({ setPage, onClose, first, focusedOnMount, last }) => (
   </>
 );
 
+ProjectTransfer.propTypes = {
+  setPage: PropTypes.string.isRequired,
```
sarahzinger: same deal

sarahzinger: don't forget to pass in first/focusedOnMount (or remove it if you don't want/need to use it)

src/components/delete-account/delete-account-modal.js
```diff
@@ -35,22 +38,82 @@ const DeleteInfo = ({ setPage, onClose, first, focusedOnMount, last }) => (
...
+        <Button className={styles.actionButton} variant="secondary" ref={last} onClick={onClose}>
+          Close
+        </Button>
+      </Actions>
+    </>
+  );
+};
+
+TeamTransfer.propTypes = {
+  setPage: PropTypes.string.isRequired,
```
sarahzinger: not a string

whimsicallyson:
> i think the teams count is wrong
what makes you think that? wondering if I have the wrong understanding of when a team will be deleted

sarahzinger:
@whimsicallyson just checked again and this time it's right! I think what happened is that I had the delete modal up, and then created a team and it didn't update and then I was confused lol. But that's a weird edge case I think.

PR #1115, opened 2019-12-06
Add jsconfig.json

oliviacpu:
## Changes:
I’ve been keeping this `jsconfig.json` in the root of Glitch-Community so that I can use Cmd+Click in VS Code to navigate through the aliases (`import whatever from ‘Components/whatever’`) to where the files are.

## How To Test:
* See that you can use Cmd+Click in VS Code to navigate through aliases.
* See that VS Code is not showing errors that were not there before and are not real.

## Feedback I'm looking for:
Anything!

- [ ] It shouldn't change anything about the site.

### DIFF
--- a/jsconfig.json
+++ b/jsconfig.json
@@ -1,0 +1,16 @@
+{
+  "compilerOptions": {
+    "allowSyntheticDefaultImports": false,
+    "baseUrl": "./",
+    "paths": {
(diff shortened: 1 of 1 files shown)


oliviacpu:
@sarahzinger I will try it on staging!

PR #1118, opened 2019-12-10
Revert "Quixotic-rocket: Ignore the cached user if it doesn't match the signed in user"

GregWeil:
Reverts FogCreek/Glitch-Community#1108


GregWeil:
Figured out the cause and fixed in #1119

PR #1120, opened 2019-12-10
Persistent papaya: Show the create page on glitch.com if the feature switch is enabled

GregWeil:
## Links
https://persistent-papaya.glitch.me/
https://app.clubhouse.io/glitch/story/3502/a-b-test-a-swap-of-glitch-com-and-glitch-com-create

## Changes:
- `/` and `/index.html` shows the homepage or create page based on the `swap_index_create` feature
  - The feature is 50% in dev, restricted to signed out users with no projects
- The `Start Remixing` button on the create page says `Start Coding`
- The create page tweaks the window title if it is shown on the homepage

## How To Test:
- You can manually enable it at /secret, but it will cause flicker on page load
1. Go to https://persistent-papaya.glitch.me/ and make sure you are signed out
2. Go to /secret and check if the feature is enabled (it's a 50% chance for anon users with no projects)
3. If it isn't, run `document.cookie = 'optimizely-id='` in the console and refresh the page until it is
- Signed out users with no projects should see either the homepage or the create page
- Signed out users with projects should always see the homepage
- Signed in users with any number of projects should always see the homepage

- https://github.com/FogCreek/Glitch-Editor/pull/625 fixes flicker from signin/out and project management via the editor, by making the editor aware of the ssr cookies


GregWeil:
Updated with fixes for tests

PR #1121, opened 2019-12-11
Community Infra connections for staging

cori:
## Links
* [README](ci/README.md)
* [diagram](https://docs.google.com/drawings/d/1KmJhEIkqhS2VBxhiZNYJSiO0iUX8IrVQW4soiHcE_0k/edit)

## Changes:
* updated `.circleci/config.yml` to add deploy steps
* added scripts in `ci` to handle CircleCI-based and local-to-the-community-hosts deploy scripts

## How To Test:
* merge changes into staging
* watch CircleCI job execute (and, hopefully, succeed!)
* check staging.glitch.com for update

## Feedback I'm looking for:
* is this understandable?
* is it clear how development will change?
* what additional docs or guidance is needed?
* is it stable?
* what else do we need to address (aside from moving into prod)?

* i think this can be merged and we can start testing it - it shouldn't affect anyone not merging to `staging`, which is new, so....


.circleci/config.yml
```diff
@@ -3,75 +3,140 @@
...
       - checkout
+      - run: 
+          name: prep-for-artifacts
+          command: mkdir -p $CIRCLE_ARTIFACTS
       - run:
-          name: update-pnpm
-          command: "sudo npm install -g pnpm@3.x"
+          name: setup-ssh
+          command: cp ./ci/ssh.config ~/.ssh/config
+      # - run:
```
clottman: Any reason we still need this commented out? Or should we just axe it?

keithk:
Haven't had a chance to pour through this yet, but wondering if when we're ready to deploy the comments (especially to notes to self) will be removed, or if there are any of them that remain questions that we should answer?

ci/README.md
```diff
@@ -0,0 +1,59 @@
+# Glitch.com Continuous Integration Configuration
+
+## Overall Process
+
+1. local dev &#x2672;
```
clottman: What's this character?

ci/README.md
```diff
@@ -0,0 +1,59 @@
...
+1. push to `staging`
```
clottman: Can you clarify what this means? Like is the expected flow that we merge a pull request to staging? Or push from the command line?

ci/README.md
```diff
@@ -0,0 +1,59 @@
...
+   1. CircleCI picks up the commit
+      1. lints, builds, runs tests, just like it does right now
```
clottman: This is minor but can we make this an actual numbered list? So when discussing instructions we can say ie, "I just followed step 3 and then this happened"

ci/README.md
```diff
@@ -0,0 +1,59 @@
...
+      1. checks if the asset is in s3 [remotely executed on a worker]
```
clottman: When you say "checks if the asset is.." - what asset are you talking about?

ci/README.md
```diff
@@ -0,0 +1,59 @@
...
+      1. if not, sends asset for upload [remotely executed on a worker]
+      1. sends `deploy` command, which gets the asset from S3 and deploys it to the local box [remotely executed on each worker]
+1. merge to `master`
+   1. CircleCI picks up the commit
+      1. same process as for staging @ 2.1.1 above, but deploying to the production boxes
```
clottman: Same comment here re: numbering the list. Also an example of why - which "2.1.1" is this referring to?

ci/README.md
```diff
@@ -0,0 +1,59 @@
...
+      1. there's nothing in this process that indicates that staging has to be involved. If you want to have folks PR directly against / merge directly into `master` that should also work just as well.
+
+## Details
+
+### No secrets in the Glitch-Community repo
+
+A central tenet to the community CI work is that the community repo shouldn't house any secrets (even encrypted ones). This means that any information we need to get from somewhere else (AWS, primarily) needs to be proxied through something that has the appropriate permissions. To avoid additional complexity, I just relied on the built-in access that the community host machines already have by virtue of them living in our infrastructure and already having the Glitch repo in place. The downside to this is that this CI process is tightly tied to the GLitch CI process in a number of ways, so when that process changes this process will have to undergo corresponding changes. There are a few touchpoints in the Glitch code that represent this link, so I don't think there'll be major changes to the Glitch deploy process without some attention paid here, but it's worth noting.
```
clottman: I don't understand what this means. In the past if I needed to add a new .env value (maybe an API key to reach Prismic or a new service), I'd add it to .env. What should I do in the new system if I need to add a key? Where do I find the current key, for example if I need to cycle an API key?

ci/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+* I think it might be worthwhile to add some level of tracing to this process in the future
+* finish production infra (s3 bootstrapping bucket and the like)
+* connect `master` to prod infra and test some deploys
+* flip to prod infra
+
+### Things I chose _not_ to do
+
+* prevent overlapping CircleCI deploys
+  * there are so few boxes, deploys should happen so uickly, and the number of simulatneous deploys is so low this didn't seem worth any effort right now
```
clottman: Minor - typo here on 'uickly'

ci/README.md
```diff
@@ -0,0 +1,59 @@
...
+* implement `no-deploy` or `no-test` flags
+  8 again this didn't seem needed right away; we can implement them later if we care to
```
clottman: 8?

ci/check-deploy-source.sh
```diff
@@ -0,0 +1,27 @@
...
+set -x  #   we don't want -e or -o pipefail; we want to handle the results of the grep at the end
+
+#####
+# THIS SCRIPT RUNS ON COMMUNITY WORKERS
+# It checks for the requested asset (based on the commit sha) in S3 
+# and tells the caller who is responsible for the asset
+#####
+
+# check req params - we need the env and a sha to use for file manipulation
+if [ 2 -ne "$#" ]; then
```
clottman: I'm no shell wizard and this is hard to google - what does `$#` mean? At what point would it be equal to 2?

ci/check-deploy-source.sh
```diff
@@ -0,0 +1,27 @@
...
+  exit 1
+fi
+
+export ENVIRONMENT=$1
+export CIRCLE_SHA=$2
+
+source /opt/glitch-community/ci/env
+
+# check S3 for the asset; tell caller the result
+# aws s3 ls s3://community-bootstrap-bucket20191205165831056600000001 | grep "$CIRCLE_SHA"; code=$?
```
clottman: what's the commented out part doing here? Do we need it?

ci/local-deploy.sh
```diff
@@ -0,0 +1,47 @@
...
+fi
+
+export ENVIRONMENT=$1
+export CIRCLE_SHA=$2
+
+cd /opt/glitch-community
+
+source ci/env
+
+# we run npm i here in case pm2 is not available but we could probably just swallow that error
```
clottman: What does "in case pm2 is not available" mean? What conditions would cause that, and how do we recover?

ci/local-deploy.sh
```diff
@@ -0,0 +1,47 @@
...
+npm i && npm run stop && wait
+
+# avoid cruft like deleted files from hanging around; currently removed folders will still persist
```
clottman: minor - the 'from' in the first part of the sentence confuses me a bit. Maybe it's a typo?

ci/local-deploy.sh
```diff
@@ -0,0 +1,47 @@
...
+rm "$CIRCLE_SHA.tar.gz"
+
+#   retrieve the stored .env file
+aws s3 cp --quiet s3://$BOOTSTRAP_BUCKET/.env .
+source .env
+
+#   install the deps, run the app
+npm i && wait
+
+#   one can temporarily add a pseudo-random pause here
```
clottman: When would this be necessary?

package.json
```diff
@@ -26,8 +26,9 @@
     "lint": "npm run lint:server && npm run lint:src",
     "lint:server": "eslint --config server/.eslintrc.server.js server shared webpack.config.js",
     "lint:src": "eslint --config src/.eslintrc.js src",
-    "prestart": "npm run watch:lint:server & \n if if-env BUILD_TYPE=watcher; then \n npm run watch:webpack & \n fi",
-    "start": "pm2-runtime start ecosystem.config.js",
+    "prestart": "if if-env BUILD_TYPE=memory; then \n npm run watch:lint:server & \n fi & \n if if-env BUILD_TYPE=watcher; then \n npm run watch:webpack \n npm run watch:lint:server & \n fi",
```
clottman: We probably want to clean this up and remove BUILD_TYPE=watcher - is that on your todo list @cori or something we should make a ticket for? 

Also - what's this change doing?

.circleci/config.yml
cori: i removed this one. There were others in some of the testing jobs that I don't recall modifying (although I probably did) so I reset them - those were outside my scope.

.circleci/config.yml
```diff
@@ -3,75 +3,140 @@
...
+      #     name: update-pnpm
+      #     command: "sudo npm install -g pnpm@3.x"
       - restore_cache:
           keys:
-            - deps-{{ checksum "shrinkwrap.yaml" }}
+            # in Glitch we key on .Branch, but this is probably just as good
+            # diff branches will probably have different checksums
```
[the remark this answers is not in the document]
cori: that was an out-of-date comment, sorry!

ci/README.md
cori: This and the list numbering were intended to be viewed in the rendered file, but nevertheless they weren't adding anything so I removed them.

ci/README.md
cori: I think there is some discussion to be had on this front right now, and I'm more or less agnostic about it. Either would work from the technical side of things, as would skipping staging entirely (once we've got master wired up)

ci/README.md
cori: done

ci/README.md
cori: updated, but let me know if still unclear

ci/check-deploy-source.sh
cori: that was a left-over - gone now

ci/local-deploy.sh
cori: hopefully more clear now

ci/local-deploy.sh
cori: clearer?

ci/local-deploy.sh
cori: I used this extensively in testing long-running deploys, and in case we need to run some more tests of that nature I didn't want to remove it yet.

package.json
cori: I had to make sure that we did _not_ run the lint watcher in "serving" ("static") mode (in either staging or production) because the CircleCI job would never exit. This may not be the right or best way to handle this, and any other BUILD_TYPE changes are outside my remit - if there's a better change that would accomplish what I was looking for I'm happy to hear it!

cori:
> Haven't had a chance to pour through this yet, but wondering if when we're ready to deploy the comments (especially to notes to self) will be removed, or if there are any of them that remain questions that we should answer?
> 
> There also appears to be a good amount of code that is entirely commented out, so once we're sure that it's not required can we remove it (so it's not dead code)

I mostly cleaned up comments, but I didn't get back to `.circleci/config.yml` before I sent the PR, so that one didn't get cleaned up. The subsequent commit should address much of that, so I don't think there should be much a) dead code or b) overly-personal comments left, but let me know what you find!

ci/README.md
cori: see prev.

ci/README.md
cori: As the new deploy is currently structured, the `.env` file is stored in the s3 bucket and would need to be updated there, which (again as things stand on 2019-12-12) will require coordination with someone with access - potentially via an Infrastructure story in ClubHouse, but TBD).

In going that route I looked at the current `.env` file and its history and it seemed like it changed only rarely so this seemed like a fairly small burden. There's only one key in there right now that seemed like it should be protected (which is what this configuration is attempting to retain) but it seemed likely that number was only going to go up. I'm open to discussions about whether this is the correct, or even a workable approach.

ci/README.md
cori: *

ci/check-deploy-source.sh
cori: `$#` is the number of arguments passed in to the script, but you're right, bash is very hard to google. Hopefully the updated comment is clearer.

ci/README.md
keithk: I'm thinking that we open PRs to staging, verify it on staging.glitch.com (because it's on the same infrastructure as the production site, different than testing locally), and then git merge staging into master

ci/README.md
```diff
@@ -0,0 +1,59 @@
...
+- .circleci
+  |
+  - config.yaml - CircleCI configuration file, enhanced with the deploy job
+- ci
+  |
+  - deploy.sh - script run on the CircleCI executor; marshalls the deploy behavior that takes place on the community hosts themselves
+  |
+  - (these files all run on the hosts in the Glitch Infrastructure)
+  - check-deploy-source.sh - checks to see if the asset for the current sha is in S3
+  - upload-asset.sh - handles uploading the provided asset to the appropriate bucket and marking the correct sha
```
mads-hartmann: Would it be possible to remove `check-deploy-source.sh` entirely and instead invoke `upload-asset.sh` in `config.yaml` if you generated AWS access key/secret that only allowed reading/writing to the community bucket and added it to the CirleCI environment?

If that's feasible I think that would make the deploy.sh script easier to read :)

ci/deploy.sh
```diff
@@ -0,0 +1,72 @@
...
+echo "${HOSTNAMES[@]}"
+
+for name in ${HOSTNAMES[*]}
+do
+
+  # if we run into a problem with the sequence of remote call we fall back to 
+  # shoving the asset to the destination server and brute forcing the deploy
+  # this is super-naive, and we mightr want to handle errors more explicitly
+  catch() {
+    echo "something bad happened; let us see if just pushing the asset will work"
```
keithk: What happens if this doesn't work?

ci/env
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env bash
+
+if [ -n "${CIRCLE_BRANCH:-}" ]
```
mads-hartmann: You could collapse all of this into a single switch statement:

```sh
case "${CIRCLE_BRANCH}" in 
    master)
        export ENVIRONMENT=production
        export BOOTSTRAP_BUCKET=
        export JUMP_DOMAIN=jump.glitch.com
        ;;
    staging)
        export ENVIRONMENT=staging
        export BOOTSTRAP_BUCKET=community-bootstrap-bucket20191205165831056600000001
        export JUMP_DOMAIN=jump.staging.glitch.com
        ;;
    *)
        echo "Unknown branch '$CIRCLE_BRANCH'"
        exit 1
esac
```

ci/README.md
cori: We could definitely do that; one of my unifying principles was not to have any secrets in the Community configuration, but we could achieve this in CircleCI itself without secrets in the _repo_, that's just not something I see us doing elsewhere right now, so I didn't want to open up a new layer of dependencies or a new pattern if I could avoid it, but maybe I bent over too far?

I guess it's a matter of preference and priority. I can implement that change without too much difficulty, I think, if we agree that it would be a substantial improvement - it would make the script easier to read and probably a _bit_ less fragile.

ci/deploy.sh
cori: That represents a serious problem that probably needs investigation. Probably the easiest way forward to fix an ongoing deploy would be to terminate the host, but if it happens often we'd want to see if something in the process is broken.

ci/env
cori: in some places we know the `ENVIRONMENT` (because we set it) but not the `CIRCLE_BRANCH` (because that's only available inside `deploy.sh` which runs on the executor). There's probably a better way, though.

ci/README.md
mads-hartmann: I think we upload our bootstrap scripts directly from CIrcleCI and the editor does it for their assets I think. For example we have both  `STAGING_AWS_BOOTSTRAP_SECRET` and `PRODUCTION_AWS_BOOTSTRAP_SECRET` in our circle environment variables right now :)

ci/env
mads-hartmann: Ah! I didn't catch that. If you could express that with a comment in the script that would be helpful :) 

I personally prefer switches of if's but that's just a personal preference so feel free to ignore this version:

```sh
# If the ENVIRONMENT isn't known we'll infer it from the branch
if [[ -z "${ENVIRONMENT}" ]]
then
case "${CIRCLE_BRANCH}" in 
    master)
        export ENVIRONMENT=production
        ;;
    staging)
        export ENVIRONMENT=staging
        ;;
    *)
        echo "Can't infer ENVIRONMNET from branch '$CIRCLE_BRANCH'"
        exit 1
esac
fi

# Set the right BOOTSTRAP_BUCKET and JUMP_DOMAIN based on the active ENVIRONMENT
case "${ENVIRONMENT}" in 
    production)
        export BOOTSTRAP_BUCKET=
        export JUMP_DOMAIN=jump.glitch.com
        ;;
    staging)
        export BOOTSTRAP_BUCKET=community-bootstrap-bucket20191205165831056600000001
        export JUMP_DOMAIN=jump.staging.glitch.com
        ;;
    *)
        echo "Unknown environmnet '$ENVIRONMENT'"
        exit 1
esac
```

package.json
GregWeil: This changes the default behavior when `BUILD_TYPE` isn't set. It's supposed to act like `memory` but this won't run the server linter.

I would like to keep BUILD_TYPE=watcher around, it doesn't restart webpack on server restart which is helpful when making server changes.

ci/README.md
cori: Yep! Those are encrypted and stored in the repo at `/ci/secrets/secure.env`. which was what I was working to avoid doing in the Glitch-Community repo.

We could use CircleCI builtins to set up AWS access and manage the secrets in CircleCI, which was the alternative I was referring to (that I don't think we do anywhere else)

ci/local-deploy.sh
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env bash
+set -xeuo pipefail
+
+#####
+# THIS SCRIPT RUNS ON ALL COMMUNITY WORKERS
+# It manages the running site, gets the requested asset, and deploys and installs it. 
+# This script suffers from a bootstrap problem - if you change this script
```
mads-hartmann: I think the way we get around the bootstrap problem in our `backend-local.sh` scripts is that the `parallel` command will transfer the script to the host before executing it.

ci/README.md
mads-hartmann: The are also set in the CircleCI environments variable config :) https://circleci.com/gh/FogCreek/Glitch/edit#env-vars

package.json
cori: I can try to replicate that logic - is there somewhere I should be looking for what the expected behavior is so I can try to match it? Is it basically just that unset `BUILD_TYPE` should behave like `memory`?

ci/README.md
mads-hartmann: I have a feeling they don't actually need to be in `/ci/secrets/secure.env` but I haven't verified this

ci/local-deploy.sh
cori: ah using `basefile` and `cleanup`. That's a good tip. I'm not using `parallel` anywhere here since we're explicitly counting on hosts being deployed to one by one, but if this bootstrap problem becomes onerous it might be worth considering (or something similar)

ci/README.md
cori: yup i totally missed those - I saw them being set via the top of `deploy.sh`, but this might make it more explicit; thanks!

package.json
GregWeil: The readme has the default value for each var, but it doesn't go that in depth on their effects. Unset BUILD_TYPE should be equivalent to memory.

ci/README.md
eestolano: I just verified it myself.

There are a couple of scripts (like `hostnames-by-role`) that get secrets from the CircleCI env if they're run in CircleCI, but get the secrets from `/ci/secrets/secure.env` when run outside of CircleCI. CircleCI itself doesn't get the key for the repo.

ci/local-deploy.sh
eestolano: I'd advise solving the bootstrap problem now; it's likely to cause problems at an extremely inconvenient time. I think the easiest way is to `scp` this script over into `/tmp`, and run that copy of it. I don't think it's important to remove it afterwards.

(If you want to use `parallel` with `basefile` and `cleanup`, you could set a parallelism level of 1.)

package.json
cori: I believe the most recent commit handles the unset BUILD_TYPE appropriately, or it seemed to in my local testing.

ci/env
cori: Updated!

ci/deploy.sh
```diff
@@ -0,0 +1,72 @@
...
+    # do the local deploy stuff
+    ssh -o "ProxyJump $JUMP_DOMAIN" -o StrictHostKeyChecking=no "$1.$ENVIRONMENT" "bash --login -c \"cd /opt/glitch-community && ci/local-deploy.sh $2 $3\""; code=$?
+
+  }
+
+  echo $name
+
+  #check if the asset is already in S3
+  ssh -o "ProxyJump $JUMP_DOMAIN" -o StrictHostKeyChecking=no "$name.$ENVIRONMENT" "bash --login -c \"cd /opt/glitch-community && ci/check-deploy-source.sh $ENVIRONMENT $CIRCLE_SHA\"" > /dev/null 2>&1; code=$?
+  S3_LOOKUP_RESULT="$code"
```
eestolano: This may be a moot point if you end up getting rid of `check-deploy-source` entirely, but you could simplify a couple of things here:

* `-e` doesn't exit on errors for commands in `if` statements.
* You already have CircleCI deploying ~/.ssh/config in advance; with a suitable SSH config file, you won't need those SSH options in the command line.
* I could be wrong, but it _looks_ like `check-deploy-source` cares neither what directory it's in nor whether  you're using a login session

So you could do something like the following, and also set `-e`:

```
if ssh "$name.$ENVIRONMENT" "/opt/glitch-community/ci/check-deploy-source.sh $ENVIRONMENT $CIRCLE_SHA" > /dev/null 2>&1; then
   # &c &c &c
fi
```

Or, if `-e` doesn't work for more complicated reasons, you can always `set +e` immediately before the problematic commands, and `set -e` again afterwards.

ci/deploy.sh
```diff
@@ -0,0 +1,72 @@
...
+fi
+
+export CIRCLE_SHA=$1
+source ci/env
+
+# TODO
+# * connect env and branch to remove hard-coded vals
+# * should this connect to Honeycomb? I think it would be valuable.
+
+# first get the list of hostnames - we could do this on any host, but we know the worker has the code
```
eestolano: You can run `hostnames-by-role` directly from the CircleCI executor, if you set `AWS_DESCRIBE_INSTANCES_KEY` and `AWS_DESCRIBE_INSTANCES_SECRET` appropriately in the CircleCI env. That's how we do it for the `Glitch` repo deploys.

ci/README.md
eestolano: And, zooming out, I agree with Mads' suggestion to get rid of `check-deploy-source.sh` and run `upload-asset.sh` directly from CircleCI itself. It would be much simpler.

package.json
```diff
@@ -26,8 +26,9 @@
     "lint": "npm run lint:server && npm run lint:src",
     "lint:server": "eslint --config server/.eslintrc.server.js server shared webpack.config.js",
     "lint:src": "eslint --config src/.eslintrc.js src",
-    "prestart": "npm run watch:lint:server & \n if if-env BUILD_TYPE=watcher; then \n npm run watch:webpack & \n fi",
-    "start": "pm2-runtime start ecosystem.config.js",
+    "prestart": "if if-env BUILD_TYPE=\"\"; then \n npm run watch:lint:server & \n fi & \n if if-env BUILD_TYPE=memory; then \n npm run watch:lint:server & \n fi & \n if if-env BUILD_TYPE=watcher; then \n npm run watch:webpack \n npm run watch:lint:server & \n fi",
```
eestolano: Maybe the prestart could be a tiny script? This one-liner is hard to read.

src/components/collection/collection-result-item.styl
```diff
@@ -11,9 +11,3 @@
   position: relative
   padding-top: 6px
   z-index: 1
-
-.private
-  > div
-    background-color: private
-  > div:hover
-      background-color: general-link
```
GregWeil: This doesn't seem like an intentional part of your pr, but this code was already removed from ~community. Maybe you got a bad merge/git doesn't see this branch as merged?

ci/deploy.sh
cori: I suppose having that call here would save an http call and associated latency and point of failure. You think that's worth the overhead of keeping a separate copy of that script here (i don't have the Glitch repo available here so I'd need to copy the script if I understand what you mean).

ci/README.md
keithk: Just checking on the status of these suggestions, and if we're moving forward w/ them @cori ?

cori: Yep I'm still working on these changes. hope to have them out by EoD today.

ci/deploy.sh
eestolano: Oh, right, you'd have to copy the script. I missed that somehow.

Avoiding the single point of failure is worthwhile, but you can do that without copying the script: use the jump box fleet instead of the worker box (`jump.staging.glitch.com` for staging and `jump.glitch.com` for prod)

(Also, the ssh command can be a bit simpler; the `hostnames-by-role` command doesn't care about having a bash login session or being in any particular directory.)

ci/deploy.sh
cori: I do think this is moot with recent changes.

package.json
cori: I think this might be a useful change. In the long run `package.json` is something the community team will manage, so I'll defer to you on this @keithk.

ci/local-deploy.sh
cori: This is addressed in the latest set of commits.

ci/deploy.sh
cori: I cleaned these up a little; I'm going to resolve this and you can take another look if you care to.

ci/deploy.sh
cori: Let me know if we need to discuss this further, @keithk!

package.json
cori: This is now working as expected, I _think_, but we may want to extract this complicated command into a tiny script. Deferring that for the moment until we talk about https://github.com/FogCreek/Glitch-Community/pull/1121#discussion_r359072657

cori:
All right! After a mess of commits I think I've addressed the central concerns / suggestions here. We are now:

* uploading assets to s3 directly from CircleCI
* avoiding problems with updating local-deploy.sh by sending it to the host's `/tmp` directory and executing from there

that's _most_ of what I addressed here, along with a few other minor changes.

cori:
@clottman I'm not sure where you requested changes and whether or not they've been addressed, but maybe you can take another look? I couldn't re-request a review from you.

ci/README.md
```diff
@@ -0,0 +1,59 @@
+# Glitch.com Continuous Integration Configuration
+
+## Overall Process
+
+1. local dev
+2. push to `staging`
+   i. CircleCI picks up the commit
+      a. lints, builds, runs tests, just like it does right now
+      b. queries s3 to determine if the build archive is already present [remotely executed on a worker]
```
clottman: Given the changes you mentioned in one of your last comments on this PR, is this doc line still accurate?

cori: Nope, that’s a good catch! I plan on updating the readme and associated diagram,  but figured I’d wait to see if more changes were needed; should have made that clear.

ci/deploy.sh
```diff
@@ -0,0 +1,48 @@
...
+
+  # do the "local" deploy stuff
+  ssh "$name.${ENVIRONMENT}" "bash --login -c \"cd /tmp && ./local-deploy.sh ${ENVIRONMENT} ${CIRCLE_SHA} ${COMMUNITY_AWS_BOOTSTRAP_SECRET}\""; code=$?
+
+done
+
+if [ ${code} -ne 0 ]; then
+  echo "Deploy failed"
+  exit ${code}
+fi
```
eestolano: This looks like the error codes for the `scp` commands get ignored, and only a failure of the `ssh` command counts as a failure.

ci/env
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/env bash
+
+set -x
+set +u  # some vars will be unset here depending on run context (CircleCI or community worker)
+
+# If we have CIRCLE_BRANCH this is running in CircleCI; set ENVIRONMENT appropriately
+# and use the CircleCI env for secrets
+if [[ -n "${CIRCLE_BRANCH}" ]];
+then
```
eestolano: Is this `CIRCLE_BRANCH` check the only case where you expect to read an unset variable? You can test for its existence in a `set -u`-compatible way by replacing `if [[ -n "${CIRCLE_BRANCH}" ]];` with `if [[ -v CIRCLE_BRANCH ]];`

ci/local-deploy.sh
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env bash
+set -xeuo pipefail
+
+#####
+# THIS SCRIPT RUNS ON ALL COMMUNITY WORKERS
+# It manages the running site, gets the requested asset, and deploys and installs it. 
+# This script suffers from a bootstrap problem - if you change this script 
+# a simple deploy will probably not suffice and may require some special management
```
eestolano: (Now that the bootstrap problem is resolved, this comment can go away, right?)

ci/deploy.sh
cori: Good catch, thanks!

in https://github.com/FogCreek/Glitch-Community/blob/48dbb2095542e89237ceba7907fde9823ab4bb31/ci/publish-build-asset.sh#L36 I'm summing exit codes just to get success/failure, and the same would work here, although in means the exit codes themselves are useless (although as long at `set -x` is active you can see each step's exit code in the logs). Do you think that's a useful way to handle this, or do you know of a better one?

FWIW I pushed the change I mentioned above to test against the deploy.

ci/env
cori: Ah, great tip! Thanks! Pushed!

ci/local-deploy.sh
cori: fixed

src/components/collection/collection-result-item.styl
cori: Oh I forgot to respond to this but Mads asked about "all the html file changes" as well this morning.

I'm pretty sure this is just related to the fact that the staging branch is new and I've only intermittently been merging changes from `Glitch-Community/master`. I think when this is merged into `staging` and we merge `master` in again it'll be easier to make sure everything looks as it should. But if it's concerning I can dig deeper first.

GregWeil: I think if master gets pulled into staging these changes will stop showing up without having to pull anything into this pr.

ci/deploy.sh
eestolano: Huh, I hadn't seen that technique before. That technique actually proceeds and runs subsequent commands if a command fails, and only fails out in the end. I think you're better off failing immediately -- both here and in `publish-build-asset.sh`. Otherwise, you could have some weird behavior -- for instance, if `scp`ing the deploy script to a box fails,, but the subsequent SSH command succeeds, you'll end up running a deploy on that box with a previous version of the deploy script.

ci/publish-build-asset.sh
```diff
@@ -0,0 +1,47 @@
+#!/usr/bin/env bash
+set -x  #   no -e or -o pipefail; we need to handle the exit code of the s3api call
```
eestolano: If you put the `s3api` call in an `if` statement, then a failure won't make it exit with -e

ci/publish-build-asset.sh
```diff
@@ -0,0 +1,47 @@
...
+aws s3api head-object --bucket "$COMMUNITY_BOOTSTRAP_BUCKET" --key "$CIRCLE_SHA.tar.gz" > /dev/null 2>&1; code=$?
+
+#   check to see if we have the package
+if [[ -f "/home/circleci/$CIRCLE_SHA.tar.gz" ]]; then
+
+  if [[ "$code" -ne 0 ]]; then
+    #   no file in s3
+    aws s3 cp --quiet "/home/circleci/$CIRCLE_SHA.tar.gz" "s3://$COMMUNITY_BOOTSTRAP_BUCKET"; code=$?
+    echo "$CIRCLE_SHA" > /home/circleci/LAST_DEPLOYED_SHA
+    aws s3 cp --quiet /home/circleci/LAST_DEPLOYED_SHA "s3://$COMMUNITY_BOOTSTRAP_BUCKET"; code=$(($code+$?)) # sum exit codes for a final success value
```
eestolano: I imagine you don't actually want to update `LAST_DEPLOYED_SHA` if you failed to upload the tarball?

ci/deploy.sh
cori: updated!

ci/publish-build-asset.sh
cori: I was resisting this because that syntax feels awkward to me, but better to _have_ the `-e`.

cori: oh, and I think resolved

ci/publish-build-asset.sh
cori: nope, thanks! fixed.

ci/deploy.sh
```diff
@@ -0,0 +1,51 @@
...
+echo "${HOSTNAMES[@]}"
+
+for name in ${HOSTNAMES[*]}
+do
+
+  echo "${name}"
+
+  # try to avoid potential bootstrap problems with updates to the local deploy by uploading it 
+  # (and the corresponding env file) to /tmp and executing from there.
+  scp /home/circleci/project/ci/local-deploy.sh /home/circleci/project/ci/env "deploy@${name}.$ENVIRONMENT":/tmp; code=$?
```
eestolano: whoops, missed this earlier. If this fails, the script will simply exit because of the `set -e`. Not the worst thing, since you're exiting anyway, but you miss out on the 'Deploy command failed' message.

Some options:
* stick it in an `if` statement
* `set +e` before the command, then `set -e` afterwards

PR #1124, opened 2019-12-13
Superb guide: Expand bootstrap error messages, fix for static builds

GregWeil:
## Links
https://app.clubhouse.io/glitch/story/5246/investigate-ssr-rendering-for-google-crawler

## Changes:
- Static builds use build/server.js instead of src/server.js
- Allow no-js ssr pages when running in the google cache
- Check for network errors loading the js code, and show a unique error if that happens

## How To Test:
- Load up the site (it's in a static build right now)
- Join the project and delete build/client.hash.js, notice that the 'your browser does not support js' message does not appear, and instead it says something failed to load

### DIFF
--- a/views/index.ejs
+++ b/views/index.ejs
@@ -51,6 +51,7 @@
       var OPTIMIZELY_ID = <%- JSON.stringify(OPTIMIZELY_ID) %>;
       var BUILD_COMPLETE = <%- BUILD_COMPLETE %>;
       var BUILD_TIMESTAMP = '<%= BUILD_TIMESTAMP %>';
+      var scriptLoadError = false;
     </script>
     <div id="fallback" style="display: none">
       <div class="content">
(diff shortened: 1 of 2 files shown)


views/index.ejs
```diff
@@ -98,24 +107,33 @@
...
-      var isCodeRunningInsideGoogleTranslateIframe = window.location.origin === "https://translate.googleusercontent.com"; 
+      var isCodeRunningInsideGoogleTranslateIframe = window.location.origin === "https://translate.googleusercontent.com";
+      var isCodeRunningInsideGoogleSearchCache = window.location.origin === "https://webcache.googleusercontent.com";
       
       if (isIe) {
         // ie doesn't support our css
         showFallback('ie');
       } else if (isCodeRunningInsideGoogleTranslateIframe) {
-        // only translate server rendered pages
+        // only allow translation of server rendered pages
```
clottman: What's the goal of this section? Is it still necessary since everything is getting SSR'd? (not a blocker on this PR)

GregWeil: Everything tries to ssr, but if the page throws an error the server will send an empty page and have the client render from scratch

PR #1130, opened 2019-12-17
various-dart: fix shared-components CSS dominance

whimsicallyson:
## Links
* https://various-dart.glitch.me
* https://app.clubhouse.io/glitch/story/3820/investigate-shared-components-css-dominance

## Changes:
* Move app CSS to top of `<body>` to fix cascade issues
* spent a lot of time looking into react hydration, am pretty sure that we're not actually using it in the case of the CSS files (they're not in helmet)

## How To Test:
* Honestly nothing should have changed.
* I'd recommend throttling, and make sure that you don't get giant icons on page load (shows that we're still properly pulling in shared-components CSS for SSR rendering)
* You can also check the page source and/or the HTML in the devtools for proof that app CSS is always showing up after shared-components `<style>` tags?

## Feedback I'm looking for:
Do you have a better way to do this?

- [ ] update shared-components for a couple global overrides I hadn't added in previously


src/components/project/project-embed.js
```diff
@@ -61,7 +61,7 @@ const ProjectEmbed = ({ project, top, addProjectToCollection, loading, previewOn
           <div className={styles.embedBottomBar}>
             <span className={styles.embedLeft}>
               <ProjectLink project={project}>
-                <ProjectAvatar project={project} />
+                <ProjectAvatar className={styles.avatar} project={project} />
```
GregWeil: This class doesn't make it into the dom

src/components/collection/edit-collection-color-pop.styl
```diff
@@ -1,4 +1,4 @@
-.container div[data-module="Popover"]
+.container > div
```
GregWeil: Using `div` here seems a bit fragile if we end up changing popovers to use a span or something. If you can't put a class on it, I think using `.container [data-module="Popover"]` might be better.

src/components/project/project-embed.js
whimsicallyson: this is the avatar in the lower left of the embed when you're on a project page, yes? it doesn't seem big...

GregWeil: No, this is for project embeds when playing a collection

GregWeil: This classname isn't referencing anything now

whimsicallyson: 🤦‍♀

PR #1131, opened 2019-12-17
Nova chance: Swap button text using media queries instead of react

GregWeil:
## Links
https://app.clubhouse.io/glitch/story/3489/replace-usewindowsize-with-css-media-queries

## Changes:
Added ResponsiveButton, which shows either `children` or `shortText` based on the window width

## How To Test:
- Resize the window on a project page and see how it does


whimsicallyson:
code looks good, waiting on glitch to be back up to test it out :)

PR #1134, opened 2019-12-19
Remove sign-in-pop and redirect to /login

oliviacpu:
## Links
* Remix link: https://hissing-gibbon.glitch.me/
* Issue-tracker link: https://app.clubhouse.io/glitch/story/5180/redirect-to-glitch-com-login-from-community

## Changes:
* This removes the sign-in-pop component and uses a small button as an href to redirect to `/login` in order to handle sign-in. We're doing this to stay consistent with Editor and with the launch of a similar feature there.
* This also uses the sign-in-layout component to populate the OAuth sign-in page that is currently at `/signin`. I'm going to swap the `/login` and `/signin` routes in a second PR.

## How To Test:
* Be signed out. Click "Sign In" on the header (and recent projects) and see that you are taken to `/login` and not the popover.

- [ ] ???


oliviacpu:
🙋 We might also swap `/login` and `/signin` in this PR before deploying. I'm coordinating with the Editor team to make sure these happen together.

shared/constants.js
```diff
@@ -25,27 +27,31 @@ const envs = {
...
     FACEBOOK_CLIENT_ID: '1121393391305429',
     OPTIMIZELY_KEY: 'GZMqH7Aou8QmcPV2hm16wj',
     PROJECTS_DOMAIN: 'glitch.development',
   },
-	local: {
+  local: {
     APP_URL: 'https://glitch.com',
     API_URL: process.env.FWD_SUBDOMAIN_PREFIX ? `https://${process.env.FWD_SUBDOMAIN_PREFIX}-glitch.fwd.wf/` : null,
     EDITOR_URL: 'https://glitch.com/edit/',
+    SIGN_IN_URL: 'https://glitch.com/login/',
```
clottman: What's the process for logging in to localhost now? Can we make this a relative url somehow (so it works both on community remixes on glitch and when running locally)?

oliviacpu: @clottman Oh yeah, if you just use `/login` it leaves the domain as whatever the browser has. I was being overly-complicated with the constants maybe. I did make a ticket to clarify that file in the future though, because I think it could use a little care. https://app.clubhouse.io/glitch/story/6033/clarify-shared-constants-environments

src/components/header/index.js
```diff
@@ -54,7 +53,9 @@ const Header = ({ searchQuery, showAccountSettingsOverlay, showNewStuffOverlay,
             )}
             {signedOut && (
               <li className={styles.buttonWrap}>
-                <SignInPop align="right" />
+                <Button size="small" as="a" href="/signin">
```
GregWeil: Could you change this to use `as={Link} to="/signin"` taking `Link` from `Components/link`? That way it'll load in app.

src/components/recent-projects/index.js
```diff
@@ -15,7 +14,10 @@ import { emoji } from '../global.styl';
 const SignInNotice = () => (
   <div className={styles.anonUserSignUp}>
     <span>
-      <SignInPop /> to keep your projects.
+      <Button size="small" as="a" href="/signin">
+        Sign in
+      </Button>{' '}
```
GregWeil: The same thing about using `Link` instead of `a` here

PR #1138, opened 2019-12-19
Profuse tea: Several styling fixes and a hopeful ssr fix

GregWeil:
## Links
https://app.clubhouse.io/glitch/story/5366/incorrect-button-style-for-unfeaturing-project
https://app.clubhouse.io/glitch/story/5269/user-avatars-are-stretched-in-add-team-member-pop
https://app.clubhouse.io/glitch/story/6051/the-privacy-toggle-is-off-center
https://app.clubhouse.io/glitch/story/4531/typeerror-resetstate-is-not-a-function

## GIF/Screenshots:
image
image
image

## Changes:
- Converted the add note and unfeature buttons to size=small variant=secondary
- Fixed alignment of the bookmark button and dropdown for featured projects
- The featured project text wraps instead of overlapping the bookmark button
- User avatars don't get squished in the add user to team pop
- Centered the privacy toggle icon within the button
- Added a hopeful fix for no resetState, if the code fails to load it will try again on the next render rather than staying broken

## How To Test:
- Look at all of the things that got their styling fixed and make sure they look good
- Look at a private project item to make sure the non button private icon isn't broken


GregWeil:
Not sure why linting is failing, I didn't change that file

whimsicallyson:
it seems to me that the lock icon is still one pixel too far to the right, and that the user items in the add team pop are taller than they used to be 😕

GregWeil:
Better centered the private toggle. The user avatars are the same height in firefox, though they are wider because they aren't squished (but if you close the popover and search the same thing again so it doesn't get squished it looks identical)

GregWeil:
Removed the resetState change, I'll put that in a separate pr for after vaca when I can keep an eye on it

whimsicallyson:
💯🚢

PR #1141, opened 2019-12-26
docs updated for https://github.com/FogCreek/Glitch-Community/pull/1121

cori:
## Changes:

* updated to match the changes in the [referenced PR](https://github.com/FogCreek/Glitch-Community/pull/1121), primarily the interaction between the various 3rd party services. Linked-to diagram is also correspondingly updated.
* I switched back to Markdown style numbering because GitHub-flavored Markdown (and possibly others) did not like my attempt at numbering it manually. My apologies, but please view the preview (not the diff) for the numbering.

## Feedback I'm looking for:

* is this sufficiently clear?
* should any additional guidance or information be included?
* is any of it extraneous?

## Things left to do before deploying (pulled from README.md for posterity):

* I think it might be worthwhile to add some level of tracing to this process in the future
* ~finish production infra (s3 bootstrapping bucket and the like)~ (in https://github.com/FogCreek/Glitch/pull/2444)
* connect `master` to prod infra and test some deploys
* flip to prod infra
* ~clean up expired assets~ (in https://github.com/FogCreek/Glitch/pull/2444)

### Things I chose _not_ to do (also pulled from README.md)

* prevent overlapping CircleCI deploys
  * there are so few boxes, deploys should happen so quickly, and the number of simultaneous deploys is so low this didn't seem worth any effort right now
* implement `no-deploy` or `no-test` flags
  * again this didn't seem needed right away; we can implement them later if we care to

### DIFF
--- a/ci/README.md
+++ b/ci/README.md
@@ -1,24 +1,27 @@
 # Glitch.com Continuous Integration Configuration
 
-## Overall Process
+## Overall Development Process
 
-1. local dev
-2. push to `staging`
-   i. CircleCI picks up the commit
-      a. lints, builds, runs tests, just like it does right now
-      b. queries s3 to determine if the build archive is already present [remotely executed on a worker]
-      c. if not, sends asset for upload [remotely executed on a worker]
-      d. sends `deploy` command, which gets the asset from S3 and deploys it to the local box [remotely executed on each worker]
+1. Do your development work locally on a branch (nothing here changes, I think)
+1. merge and push to `staging`
+      1. CircleCI picks up the commit
+            1. lints, builds, runs tests, just like it does right now
+            1. `deploy.sh`
+                  1. queries s3 to determine if the current commit's build archive is already present
+                  1. if not, sends asset for upload
+                  1. sends `deploy` command to each community worker, which runs `local-deploy.sh`, gets the asset from S3 and deploys it to the local box [remotely executed on each worker]
 3. merge to `master`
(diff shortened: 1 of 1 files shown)


ci/README.md
```diff
@@ -29,31 +32,11 @@ The overall structure, therefore, is this:
...
-
-* I think it might be worthwhile to add some level of tracing to this process in the future
-* finish production infra (s3 bootstrapping bucket and the like)
-* connect `master` to prod infra and test some deploys
-* flip to prod infra
-
-### Things I chose _not_ to do
+  - deploy.sh - script run on the CircleCI executor that marshalls all the CircleCI-hosted deploy work
+  - publish-build-asset.sh - script run on the CircleCI executor that checks for and uploads (if indicated) the SHA1-identified tarball of the build result for the matching commit 
+  - local-deploy.sh - does the _local_ deploy work - stops the running site, cleans up the old code, pulls and unzips the new code, npm installs and runs the site. Executed on each community worker in turn via an ssh session started from the CircleCI executor.
```
keithk: So this happening in turn means that we shouldn't end up with all of them going down at the same time to do an update, right?

cori: that's correct - each host's deploy should wait until the previous one is complete before starting on the next. We may need to tweak a few of the health check values to make sure that we're giving the newly-deployed-to hosts time to start up properly - the values in the load balancer right now are the result of my deploy testing in staging.

PR #1147, opened 2020-01-06
add production settings (prep for merging to master and starting prod deploys)

cori:
## Changes:
* adds the production S3 bucket and access key ids
* updates the staging access key to one already managed by terraform - this requires some CircleCI changes to match. future deploys will test that the key works as expected.

## How To Test:
* merge and deploy

## Feedback I'm looking for:
* sanity check
* timeline for getting these changes into prod so Circle can start using them.

- i'll merge this to staging to get the production info and the corrected staging key there. Once this is in staging, when we are ready to start deploys to the prod infra just merging staging to master (and adding the proper env var to CircleCI) should get things running.

### DIFF
--- a/ci/env
+++ b/ci/env
@@ -24,14 +24,14 @@ fi
 # set the COMMUNITY_BOOTSTRAP_BUCKET, JUMP_DOMAIN, and key id appropriately
 case "${ENVIRONMENT}" in
   production)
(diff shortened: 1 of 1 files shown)


cori:
This will need to be lightly coordinated with https://github.com/FogCreek/Glitch/pull/2462

PR #1148, opened 2020-01-06
various dart: completes account closing work

whimsicallyson:
## Links
* https://various-dart.glitch.me
* https://app.clubhouse.io/glitch/story/4033/update-frontend-to-use-new-updated-endpoints
* https://docs.google.com/document/d/1ya5WQraU0OPXf22uCsTGhaNnFU-3XyVLHSqGHzMkTPk/edit#heading=h.tcauhdj77y1d

## Changes:
* Hooks modal up to backend
* Sends email to complete deletion when you have transferred all projects/teams
* On clicking link in email, checks token validity and closes account
* Logs you out on account close
* Changes wording from "delete" to "close" (since not everything is deleted on our end)
* If you don't have projects/teams to transfer, skip those dialogs

## How To Test:
* Sign into various-dart
* Turn on Account Deletion switch
* Go to `/settings`, walk through the steps to close your account (use a test account! make sure it has some projects and/or teams with other users, where your account is the admin)
* Get an email, click the link
* You'll need to change the link to be various-dart.glitch.me/delete?token=[whatever]
* Check to make sure your account is gone, and that you've been logged out
* You can also try /delete?token=[garbage] when signed into an account and make sure you get an error

- [x] Have support write articles explaining ownership transfer and other account-closing-specifics (ch5319, ch5320, ch5321, ch5322, ch5323) 
- [x] Fill in those links
- [ ] Take out feature switch? At least have a plan as to when to do so.
- [x] Is there a way to confirm that the feedback form is working and going where folks want it?
- [x] Make sure confirmation email is properly styled (ch6301)
- [x] Add analytics (ch3242)


src/components/delete-account/delete-account-modal.js
```diff
@@ -185,15 +199,15 @@ const EmailConfirm = ({ onClose, first, focusedOnMount, last }) => {
...
+        <p>Please click the link in the email to finish closing your account.</p>
         <p>
-          If you choose to delete your account,{' '}
+          If you choose to close your account,{' '}
           <b>
             <Pluralize count={soloProjects.length} singular="project" />, <Pluralize count={soloTeams.length} singular="team" />, and{' '}
             <Pluralize count={soloCollections.length} singular="collection" />
           </b>{' '}
-          will be deleted forever.
+          will be made unavailable forever.
```
jamtur01: Is 'forever' completely true here? Might this be easier without it?

src/components/delete-account/delete-account-modal.js
```diff
@@ -66,11 +71,25 @@ const TeamTransfer = ({ setPage, onClose, first, focusedOnMount, last }) => {
       onTeamSelection(singleAdminTeams[0].id);
     }
   }, [singleAdminTeams]);
+  async function triggerEmail() {
+    try {
+      await api.post(`/v1/users/${currentUser.id}/requestDeletion`);
```
sarahzinger: would it make sense to put this in api handlers so we can aim to move our api calls to one place? https://github.com/FogCreek/Glitch-Community/blob/c6dd0c9963b90b03aa134115719a77a1d69b97b4/src/state/api.js#L168

src/components/delete-account/delete-account-modal.js
```diff
@@ -66,11 +71,25 @@ const TeamTransfer = ({ setPage, onClose, first, focusedOnMount, last }) => {
...
+      setPage('emailConfirm');
+    } catch (error) {
+      createNotification('Unable to close account, try again later.', { type: 'error' });
+      captureException(error);
+    }
+  }
+  useEffect(() => {
+    if (singleAdminTeams.length === 0) {
```
sarahzinger: Is there ever a circumstance where the component mounts but current user is not available yet so we miss this call?

Hmm in testing it seems to make the call just fine! But I do see a flash of the teams page before I get to the email confirmation page I wonder if there's a smoother way to do this.

sarahzinger: one thought i have (and feel free to push back if this is overkill!) is to have some kind of function that determines what you should see next that is separate from the modals? It could make it easy to unit test if that's of interest! It also could mean that you could skip to the end modal if you don't have any teams or you don't have any projects. 

For example we could have something like:

```
const goToNextPage = ({ currentPage, setPage, singleAdminProjects, singleAdminTeams, triggerDeleteEmail }) => {
  // ready to delete 
  if (singleAdminProjects.length === 0 && singleAdminTeams.length === 0) {
    triggerDeleteEmail()
    setPage("emailConfirmation")
    return
  }
  
  // if should show project page, do that
  // if should show team page, do that
}
```

and then in your component:
```
<Button onClick={() =>  goToNextPage({ currentPage, setPage, singleAdminProjects, singleAdminTeams, triggerDeleteEmail })}>
          Continue to Close Account
</Button>
```

src/presenters/pages/delete.js
```diff
@@ -62,33 +65,50 @@ const ValidToken = () => {
...
 
-/* TODO:
-- if the token is valid we'll want to actually delete the account
-- which hopefully forces a logout, we may have to write some code to do that on the client side as well
-- we'll also probably need a loading state, will write this code once backend for it is complete
-*/
-const useTokenIsValid = (token) => token === 'good';
+async function deleteWithToken(token, api, setAccountStatus, signOut) {
+  try {
+    await api.delete(`/v1/users?token=${token}`);
```
sarahzinger: i'd put this in apihandlers as well

src/components/delete-account/delete-account-modal.js
```diff
@@ -185,15 +209,15 @@ const EmailConfirm = ({ onClose, first, focusedOnMount, last }) => {
...
+        <p>Please click the link in the email to finish closing your account.</p>
         <p>
-          If you choose to delete your account,{' '}
+          If you choose to close your account,{' '}
           <b>
             <Pluralize count={soloProjects.length} singular="project" />, <Pluralize count={soloTeams.length} singular="team" />, and{' '}
             <Pluralize count={soloCollections.length} singular="collection" />
           </b>{' '}
-          will be deleted forever.
+          will be made unavailable.
```
sarahzinger: so in testing I used a user who had not made anything and got:
```
If you choose to close your account, 0 projects, 0 teams, and 0 collections will be made unavailable.
```
which is a little confusing for a couple of reasons:
- "if you choose to close your account" - didn't I just choose to do it by pressing these buttons? I think we might want to rephrase to "Once you click the link in your email xprojects, teams, collections will no longer be available"
- Also maybe in the case of zeros we should just say your account will be closed rather than spell out the zeros?

sarahzinger:
Oh and to answer your question about the feedback form, you should be able to see responses in segment's debugger. if you want a walk through of how to use it let me know!

src/components/delete-account/delete-account-modal.js
whimsicallyson: I'm having trouble getting something like that to work, since I can't calculate the singleAdminX outside of modal functions, and it seems wrong to calculate all of those in each modal func? I'm likely missing something though.

src/presenters/pages/delete.js
```diff
@@ -13,23 +16,37 @@ import styles from './delete.styl';
...
 
   const submitFeedback = () => {
     setShowNotification(true);
-    tracker();
+    trackFeedback();
     setReason('');
   };
 
+  useEffect(() => {
+    trackDeletion({});
```
sarahzinger: do you have to pass an argument here?

src/presenters/pages/delete.js
```diff
@@ -13,23 +16,37 @@ import styles from './delete.styl';
...
-    reasonForLeaving,
-  }));
-  const isInitialized = useIsAnalyticsInitialized();
+  const trackDeletion = useTracker('Account Closed');
+  const trackFeedback = useTracker('Feedback Submitted', {
+    feedbackForm: 'Account Closure',
+    feedback: reasonForLeaving,
+    url: window.location.href,
+  });
+  const preferEmail = true;
```
sarahzinger: what's the thinking behind preferEmail?

src/presenters/pages/delete.js
```diff
@@ -13,23 +16,37 @@ import styles from './delete.styl';
...
 
   const submitFeedback = () => {
     setShowNotification(true);
-    tracker();
+    trackFeedback();
```
sarahzinger: if we're not using any of this stuff, perhaps we should delete it

src/state/api.js
```diff
@@ -198,6 +198,10 @@ export const useAPIHandlers = () => {
       addPinnedProject: ({ project, team, user }) => api.put(`/${entityPath({ team, user })}/pinnedProjects/${project.id}`),
       removePinnedProject: ({ project, team, user }) => api.delete(`/${entityPath({ team, user })}/pinnedProjects/${project.id}`),
 
+      // account closure
+      removeUserToken: (token) => api.delete(`/v1/users?token=${token}`),
```
sarahzinger: small nitpick to rename this to something like removeUserWithToken?

src/presenters/pages/delete.js
whimsicallyson: its what we agreed to in [ch3242](https://app.clubhouse.io/glitch/story/3242/add-analytics-for-user-deletion#activity-7319)

sarahzinger: I think my preference would be to remove any code we're not using right now as it will probably confuse the next person who works on it and it might not be relevant (ex the form looks super different with radio buttons etc). If we decide we want this code specifically we always will have a record of it here in github!

src/presenters/pages/delete.js
whimsicallyson: strangely, yes. it doesn't show up in Segment if I don't?

sarahzinger: hmmm that seems highly unlikely to me since we don't do that elsewhere right?

whimsicallyson: yeah I'm uncertain why its happening, I just know it wasn't working without an argument, even though we seem to be able to do that elsewhere...

whimsicallyson:
> you can now see the settings page if you're an anon user and you've enabled it on the /secret page

fixed!

PR #1149, opened 2020-01-06
Quixotic rocket: Only send the experiment viewed event if the experiment is active

GregWeil:
## Links

## Changes:
The 'Experiment Viewed' event doesn't get sent until the experiment is actually enabled

## How To Test:
Try disabling a feature in optimizely (this remix is using the dev optimizely config) and check that the segment.io/t request is not sent. It should be sent if the feature is enabled for any percent of users, even if it isn't active for you in particular


GregWeil:
@ezramechaber this will still send the event if you don't meet the requirements for the experiment, for ex if you are on the homepage and signed in/have projects. Is that ok?

ezramechaber:
Sorry, for people who don't meet the requirements, does that mean that it will send "Experiment Viewed" but with blank data, or that it will send "Experiment Viewed" as control viewed?

I don't think either is good for data hygiene, but the latter is definitely a non-starter.

GregWeil:
It'll send as control :/ I'll take a look through the docs and see what can be done

ezramechaber:
Got it. Both are definitely not preferable (since if you aren't part of an experiment we shouldn't say you're viewing it) but the control would be a huge issue.

GregWeil:
Updated, it now only sends events if you match the audience requirements. I also fixed a few things that could cause the experiment viewed event to fire more than once in a page load

ezramechaber:
Thanks Greg. Per our Slack conversation, that seems like a reasonable implementation!

PR #1154, opened 2020-01-09
Fallacious slice: Redux for local storage

GregWeil:
## Links
https://app.clubhouse.io/glitch/story/6671/convert-local-storage-to-use-redux

## Changes:
- Local storage is in the redux store, rather than having its own context
- Small behavioral change: components used to read from storage in their first render (excluding the initial full page render) but now the hook never reads from storage outside of redux handlers or useEffect calls

## How To Test:
Browse around the site and make sure things that use storage still work (and sync across tabs)
- Signing in and out
- Updating your name
- Dev toggles
- Optimizely overrides
- Pupdates

## Feedback I'm looking for:
It's apparently bad form to store dynamic stuff such as the local storage object in the redux state (it shows a warning). The getStorage function isn't node compatible, so I've opted to have a root level var the gets set by the appMounted handler

### DIFF
--- a/src/state/local-storage.js
+++ b/src/state/local-storage.js
@@ -57,13 +57,13 @@ export const writeToStorage = (storage, name, value) => {
 export const { reducer, actions } = createSlice({
   name: 'localStorage',
   initialState: {
-    storage: null,
     cache: new Map(),
+    ready: false,
   },
   reducers: {
-    setStorage: (state, { payload }) => ({
-      storage: payload,
+    storageFound: () => ({
       cache: new Map(),
+      ready: true,
     }),
     storageUpdated: ({ cache }, { payload }) => {
       cache.delete(payload);
@@ -71,10 +71,8 @@ export const { reducer, actions } = createSlice({
     storageCleared: ({ cache }) => {
       cache.clear();
     },
-    readValue: ({ storage, cache }, { payload }) => {
-      if (!cache.has(payload)) {
-        cache.set(payload, readFromStorage(storage, payload));
-      }
+    readValue: ({ cache }, { payload }) => {
+      cache.set(payload.name, payload.value);
     },
     writeValue: ({ cache }, { payload }) => {
       cache.set(payload.name, payload.value);
@@ -95,7 +93,7 @@ export const handlers = {
       }
     };
     window.addEventListener('storage', onStorage, { passive: true });
-    store.dispatch(actions.setStorage(storage));
+    store.dispatch(actions.storageFound());
   },
   [actions.writeValue]: ({ payload }, store) => {
     writeToStorage(store.localStorage.storage, payload.name, payload.value);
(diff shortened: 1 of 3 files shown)


src/state/local-storage.js
```diff
@@ -51,75 +53,69 @@ export const writeToStorage = (storage, name, value) => {
...
+    writeValue: ({ cache }, { payload }) => {
+      cache[payload.name] = payload.value;
+    },
+  },
+});
+
+let storage = null;
+
+export const handlers = {
+  [appMounted]: (_, store) => {
```
sarahzinger: can you tell me more about this appMounted thing while we're here? I don't totally follow what we're trying to do here. We create a redux action saying the app mounted whenever the file is called yeah? seems weird right? I guess similarly I don't really understand the need for the storageFound thing, is this related to ssr somehow?

src/state/store.js
```diff
@@ -3,14 +3,20 @@ import { configureStore, getDefaultMiddleware } from '@reduxjs/toolkit';
...
 const createStore = () =>
   configureStore({
     reducer: {
+      localStorage: localStorage.reducer,
       currentUser: currentUser.reducer,
     },
-    middleware: [...getDefaultMiddleware(), createHandlerMiddleware(currentUser.handlers)],
+    middleware: [
+      ...getDefaultMiddleware(),
+      createHandlerMiddleware(localStorage.handlers),
```
sarahzinger: can you talk more about where/how we use this middleware?

src/state/local-storage.js
GregWeil: appMounted is an event that gets fired just after the first page render, so it'll happen once in the browser and never in the server. Components won't read out of local storage until storageFound is called (by the appMounted event) and marks it as ready.

GregWeil:
> 
> 
> hiya, this looks really neat! Sorry to be a pain, but I'm gunna hold off on an approve not because the code is not right but because I'm not 💯 in following it just yet and would like to! Although if someone else is there, feel free to give an approve! Since this PR is not to my knowledge blocking any product stuff would you mind walking me through it a little bit more? Can you talk more about the rerendering you think this may fix?

This doesn't actually fix any specific bugs, but I think contexts will cause a rerender on the tree under the provider when they change (not just the components calling useContext). My goal was to get a feel for redux and get rid of one layer of nesting in app.js

src/state/local-storage.js
sarahzinger: is the variable `appMounted` here an object returned by createSlice? seems like a weird key value right? should it be `appMounted.name` (also I'm probably missing something sorry! if this is a stupid question)

sarahzinger: ah jk it's createAction not createSlice, but in that it case does it return a function?

src/state/store.js
sarahzinger: when do you put logic into a handler vs in a reducer? Is the handler for side effects and the reducer just for updating state?

GregWeil: Only reducers can update state, handlers are onAction listeners, and modify the store by dispatching other actions. (`dispatch(action())` will run the associated reducer)

src/state/local-storage.js
sarahzinger: oh neat I guess it overwrites tostring: https://redux-toolkit.js.org/api/createAction#createaction

PR #1155, opened 2020-01-10
classy-afrovenator: Add billing tab to /settings

taravancil:
## Links
* [classy-afrovenator.glitch.me](https://classy-afrovenator.glitch.me)
* [[ch6209]](https://app.clubhouse.io/glitch/story/6209/add-billing-tab-to-settings): Add billing tab to /settings

## GIF/Screenshots:
**To come. The backend is returning `500`s for the `/glitchPro` endpoints. Will update when Justin is in and can investigate.**

## Changes:
This adds a "Subscription" tab to /settings behind the `pufferfish` feature flag.

## How To Test:
*Note that this is the dogfood implementation of this feature, so review should be focused on "Does it work well enough to test internally and iterate on?" rather than polish and cleverness.*

### Testing Subscription Tab Visibility
To see the tab:

1. Log in
2. Make sure you're in [@testing-team](https://glitch.com/@testing-team)
3. Make sure the `pufferfish` flag is enabled on [classy-afrovenator.glitch.me/secret](https://classy-afrovenator.glitch.me/secret)

| Case  | Expected Outcome | 
| ------------- | ------------- |
| In Testing Team + Enabled the feature on /secret  | Subscription tab visible |
| In Testing Team + Disabled the feature on /secret  | Subscription tab NOT visible |
| NOT In Testing Team + Enabled the feature on /secret  | Subscription NOT tab visible  |
| NOT In Testing Team + Disabled the feature on /secret  | Subscription NOT tab visible |

### Testing Creating + Cancelling a Subscription
**To come. The backend is returning `500`s for the `/glitchPro` endpoints. Will update when Justin is in and can investigate.**

## Feedback I'm looking for:
- Can you create a subscription with test data?
- Can you cancel that subscription?
- Do the `useStripe` and `useSubscription` hooks feel ok to you?

PR #1156, opened 2020-01-10
verbena-moth: Add /pricing page

taravancil:
This is branched off of [classy-afrovenator](https://github.com/FogCreek/Glitch-Community/pull/1155), so maybe we should wait to review/merge this until that PR is merged.

## Links
* [verbena-moth.glitch.me/pricing](https://verbena-moth.glitch.me/pricing)
* [[ch6406]](https://app.clubhouse.io/glitch/story/6406/create-a-pricing-page)

## GIF/Screenshots:
<div>

</div>

## Changes:
This adds a /pricing page behind the `pufferfish` feature flag.

## How To Test:
*Note that this is the dogfood implementation of this feature, so review should be focused on "Does it work well enough to test internally and iterate on?" rather than polish and cleverness.*

### Testing /pricing Visibility
1. Log in
2. Make sure you're in [@testing-team](https://glitch.com/@testing-team)
3. Make sure the `pufferfish` flag is enabled on [verbena-moth.glitch.me/secret](https://verbena-moth.glitch.me/secret)

| Case  | Expected Outcome | 
| ------------- | ------------- |
| In Testing Team + Enabled the feature on /secret  | /pricing `200` |
| In Testing Team + Disabled the feature on /secret  | /pricing `404` |
| NOT In Testing Team + Enabled the feature on /secret  | /pricing `404`  |
| NOT In Testing Team + Disabled the feature on /secret  | /pricing `404` |

## Feedback I'm looking for:
- Can you create a subscription with test data?
- If you have an active subscription, does the CTA on /pricing say "Manage Your Subscription"?


src/presenters/pages/about/pricing.js
```diff
@@ -0,0 +1,98 @@
...
+  const subscriptionStatus = useSubscriptionStatus();
+  const { createSubscriptionSession } = useAPIHandlers();
+  const stripe = useStripe();
+  const { currentUser, fetched: currentUserFetched } = useCurrentUser();
+  const userHasPufferfishEnabled = useFeatureEnabled('pufferfish');
+
+  async function subscribe() {
+    try {
+      const { data } = await createSubscriptionSession({
+        successUrl: `https://glitch.com${getUserLink(currentUser)}`,
```
sarahzinger: might make sense to use APP_URL from our constants file here so this works nicely on staging and stuff

sarahzinger: Also is this subscribe func the same as the last one? Might be worth thinking through if there's a place we should put this... maybe `useUserEditor` or maybe we need something like `useCustomerEditor` ? or maybe we should punt for the sake of speeding things along. Hmm I guess I've talked myself into punting lol, but something to have in the back of our minds if it's feeling like a lot of duplicated logic in the future

modernserf: if you use `window.location.origin` it will work on remixes, too

PR #1158, opened 2020-01-13
Fix settings page flicker

sarahzinger:
## Links
* Remix link: https://sarah-staging.glitch.me/settings

## Changes:
* Don't show not found page until user has been fetched

## How To Test:
* go to /secret and enable various features like user deletion and auth stuff, see how the /settings page works when features are enabled/disabled and when you're logged in/not logged in etc.


taravancil:
Pulled these changes into verbena-moth so we can deploy them together.

PR #1161, opened 2020-01-14
Fix invite user to team/join team from email

clottman:
## Links
* no remix; pull locally to test
* https://app.clubhouse.io/glitch/story/6824/can-t-join-team-from-email-invite-404-from-v1-auth-jointeam

## Changes:
* Reverts some of the changes in [this PR](https://github.com/FogCreek/Glitch-Community/commit/b3e9b73e83481f2a8ae6dbed8ff7167d5c3061ec#diff-648ea48227c4deb64d3adb96257946ff)

## How To Test:
* Invite a user you control to a team, by login or email (try both!) 
* When you get an email with a join team link, have the running app up in a tab with requests to Glitch.com blocked
* Copy the URL from the email and paste it in the address bar of the tab with requests to glitch.com blocked.
* The fancy email link will redirect you to glitch.com/something/three-word-token. Replace the glitch.com part with localhost:3000 (or whatever your port is). 
* You should join the team

* expect it to fail if the user you invited has 2FA enabled (out of scope)


GregWeil:
It seems like if I try to accept the invite without already being signed in as the invited user the page breaks with a 'currentUser.projects is undefined' error. It should sign me in as the invited user (same as an email link would).

src/presenters/pages/join-team.js
```diff
@@ -20,7 +20,7 @@ const JoinTeamPage = ({ teamUrl, joinToken }) => {
   useEffect(() => {
     (async () => {
       try {
-        const { data } = await api.post(`v1/teams/join/${joinToken}`);
+        const { data } = await api.post(`/teams/join/${joinToken}`);
```
GregWeil: I can't highlight the exact line, but I think the issue is that `data` is treated as either `{ tfaToken }` or `{ user }` when a successful join responds with the user object directly.

GregWeil: Just tested, replacing `if (data.user) replaceCurrentUser(data.user)` with `if (data.login) replaceCurrentUser(data)` works

PR #1164, opened 2020-01-15
Buttery vacation: clean up glitch pro data management

modernserf:
## Links
* Remix link: https://buttery-vacation.glitch.me/
* Issue-tracker link: https://app.clubhouse.io/glitch/story/7125/fix-subscription-cancellation-page

## Changes:
* store state in context, so only one request is made for header, footer & pages
* share subscribe functions between pricing and settings page
* update subscription state when cancelling
* rename to `useGlitchPro`


src/components/account-settings-overlay/subscription-settings.js
```diff
@@ -3,69 +3,46 @@ import { Button, Loader } from '@fogcreek/shared-components';
...
-      const { id: sessionId } = data;
-      stripe.redirectToCheckout({ sessionId });
-    } catch (err) {
-      // TODO decide what kind of error handling we need here
-      console.log(err);
-    }
-  }
-
-  async function cancel() {
+  const cancelWithWait = async () => {
```
sarahzinger: just curious whats the `withWait` part here indicate to folks?

src/state/glitch-pro.js
```diff
@@ -0,0 +1,80 @@
...
+  const [loaded] = useScript(stripeJS);
+  useEffect(() => {
+    if (loaded) {
+      setStripe(window.Stripe('pk_test_WIJxKl0T1xT8zSb2StHu8zlo'));
+    }
+  }, [loaded]);
+  return stripe;
+}
+
+function useGlitchProState() {
```
sarahzinger: would it make sense at this point to add some tests? We don't have a lot but I don't think anything technically is stopping us from making that happen.

src/components/account-settings-overlay/subscription-settings.js
modernserf: maybe `disabledButtonAndCancel`?

sarahzinger: ah ok I see what you mean, yeah how you have is fine! just wanted to know if this is like a common pattern I should learn about or something

PR #1168, opened 2020-01-16
add filteredTag util to avoid passing custom props

whimsicallyson:
## Links
* https://wind-walnut.glitch.me
* https://app.clubhouse.io/glitch/story/6389/warnings-about-textwrap-being-passed-to-dom-elements
* https://app.clubhouse.io/glitch/story/7055/team-title-bleed-formatting-issue-in-search-results
* https://github.com/styled-components/styled-components/issues/135#issuecomment-485315176 (long thread about overarching issue in styled-components, scrolled to the solution we used)

## Changes:
* adds a `FilteredTag` utility function that you can use to remove props that styled components don't like to have passed through (namely, textWrap on a `Button` that wraps a `Link`, but there could be more uses in the future)
* Additionally, added a textWrap to `team-item` to address the button bleed in case 7055.

## How To Test:
* Click around the site with devtools open, see if you find any warnings about textWrap being passed to a DOM element as shown in the initial case

## Feedback I'm looking for:
Its very possible I missed other locations where we're getting this error, even though I tried to be thorough about visiting pages
Also, is there a better way to do this? I don't love it, but it seemed like one of the least messy/convoluted ways, and is still extensible for other uses if need be.


whimsicallyson:
let's cross that bridge if/when we get to it :)

PR #1172, opened 2020-01-17
remove X from all search boxes

whimsicallyson:
## Links
* debonair-quokka.glitch.me
* https://app.clubhouse.io/glitch/story/4137/textinput-with-type-search-from-shared-components-has-overlapping-symbols-in-chrome

## GIF/Screenshots:
Screen Shot 2020-01-17 at 2 07 11 PM

## Changes:
* Adds a global "don't show the X on search inputs"

## How To Test:
* go to a project page, add to collection, search for collection

## Feedback I'm looking for:
Is this too blunt of a hammer?


src/components/global.styl
```diff
@@ -51,6 +51,16 @@ article
 section
   margin-top: 2rem
 
+/* clears the 'X' from Internet Explorer */
+input[type=search]::-ms-clear {  display: none; width : 0; height: 0; }
```
clottman: So two things - I don't think this will work without search being in `"`, but also, I can't test it because if we are running in IE, we show a "Glitch doesn't support Internet Explorer yet" message. I'm fine with either fixing the selector, or removing the IE-specific sections altogether since we don't allow people to even try using the site with IE.

whimsicallyson: let's remove all the IE stuff, since a user shouldn't get to the point of seeing a search box anyhow.

PR #1174, opened 2020-01-22
separate and Title Case team and collection page titles

whimsicallyson:
## Links
* https://cookie-ricotta.glitch.me
* https://app.clubhouse.io/glitch/story/7676/change-format-of-title-tags-for-team-and-collection-names

## GIF/Screenshots:
Screen Shot 2020-01-22 at 2 19 35 PM
Screen Shot 2020-01-22 at 2 20 15 PM

## Changes:
* Separates words in team and collection domains (by hyphen only), converts them to Title Case

## How To Test:
* go to a team or collection page, see what's in the title bar

## Feedback I'm looking for:
we cool?

### DIFF
--- a/src/presenters/pages/collection.js
+++ b/src/presenters/pages/collection.js
@@ -16,6 +16,7 @@ import { useCurrentUser } from 'State/current-user';
 import { useCachedCollection } from 'State/api-cache';
 import { useCollectionEditor, userOrTeamIsAuthor } from 'State/collection';
 import useFocusFirst from 'Hooks/use-focus-first';
+import useTitleFormatter from 'Hooks/use-title-formatter';
 import { renderText } from 'Utils/markdown';
 import allCategories from 'Shared/categories';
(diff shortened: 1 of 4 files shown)


src/presenters/pages/collection.js
```diff
@@ -45,7 +45,7 @@ const CollectionPageContents = ({ collection: initialCollection }) => {
   return (
     <>
       <GlitchHelmet
-        title={collection.name}
+        title={collection.name.replace(/-/g, ' ').replace(/\w\S*/g, (txt) => txt.charAt(0).toUpperCase() + txt.substr(1))}
```
clottman: Now that we've got this in two places it might be nice to a shared place - maybe Utils/models.js (kind of a stretch) or (probably better) we could make it a hook - useTitleFormatter() that takes a string and returns the output after running this function.

whimsicallyson: ah good call!

PR #1181, opened 2020-01-23
fix team analytics page to use new rollouts hooks

keithk:
## GIF/Screenshots:
Screen Shot 2020-01-23 at 3 50 35 PM

## Changes:
* Updates to use useFeatureEnabled, since it will either be enabled/disabled, not per team

## How To Test:
* Pull down locally and get the dev key from optimizely (keith can invite) and add that to your .env
* Go to a teams page, it should show a


keithk:
Closed because Greg has a better fix in #1175

PR #1184, opened 2020-01-27
enable prod hosts

cori:
## Changes:
* this moves the ci changes to get the `master` branch building and deploying to the new prod infra. to whit:
  * update the CircleCI config
  * adds the ci scripts for bootstrapping new community hosts and deploying releases 

## Feedback I'm looking for:
* the bulk of the changes were reviewed in https://github.com/glitchdotcom/Glitch-Community/pull/1121 when they were merged into `staging`, so this is partly a sanity check
* there are some unrelated-to-ci changes in here that are in `staging`. Should we remove those from the PR?
* one of the larger changes to the CircleCI config is to use `npm` instead of `pnpm`. I don't think that should affect anything else, though, but maybe I'm missing something?


package.json
```diff
@@ -27,138 +27,139 @@
...
-    "@babel/core": "^7.5.5",
-    "@babel/plugin-proposal-class-properties": "^7.5.0",
-    "@babel/plugin-syntax-dynamic-import": "^7.2.0",
-    "@babel/plugin-transform-runtime": "^7.5.5",
-    "@babel/preset-env": "^7.5.5",
-    "@babel/preset-react": "^7.0.0",
-    "@babel/register": "7.5.5",
-    "@babel/runtime": "^7.5.4",
-    "@babel/runtime-corejs3": "^7.5.5",
+    "@babel/core": "^7.7.7",
```
oliviacpu: I am going to make a commit that reverts these package changes, but keeps your changes to the scripts above. I think we either have a story to update packages or can make one and add it to an upcoming sprint.

src/state/api.js
```diff
@@ -56,29 +56,37 @@ export function APIContextProvider({ children }) {
...
-      overrides[method] = async (...args) => {
-        const apiWithToken = await new Promise((resolve) => {
-          setPendingRequests((latestPendingRequests) => [...latestPendingRequests, resolve]);
-        });
-        return apiWithToken[method](...args);
-      };
-    });
-    return { ...rawApi, ...overrides };
-  }, [persistentToken]);
+  const api = useMemo(
```
oliviacpu: These look like just formatting changes?

.circleci/config.yml
```diff
@@ -3,75 +3,122 @@
...
       - restore_cache:
           keys:
-            - deps-{{ checksum "shrinkwrap.yaml" }}
+            - deps-{{ checksum "package-lock.json" }}
       - restore_cache:
           keys:
             - build-{{ .Revision }}
       - run:
           name: update-pnpm
           command: "sudo npm install -g pnpm@3.x"
```
eestolano: Are you still using pnpm elsewhere? Or can this go away?

cori: ah that's a good point; we probably _could_ get rid of this update. I'm going to defer that for now in the interests of getting the community hosts spinning, but will make a note.

PR #1188, opened 2020-01-28
Sitelinks searchbox support

clottman:
## Links
* https://search-seo.glitch.me 
* https://app.clubhouse.io/glitch/story/7718/add-markup-to-enable-searching-glitch-com-from-the-search-result-for-glitch
* [Google Search Rich Results test results](https://search.google.com/test/rich-results?id=24VCMl3snCH-ufYK3UyOag) - kept erroring on the full page, so this tests the raw HTML instead. 

## GIF/Screenshots:
results from Search Rich Results test results link above

## Changes:
* adds some ld-json that tells Google/other search engines how to search glitch.com

## How To Test:
* Make sure the page still looks fine
* Can run the code or page through https://search.google.com/test/rich-results 
* Take the url in `ldJson.potentialAction.target` (line 397), replace `{search_term_string}` with other characters, and you should arrive at the glitch.com search results page with results for those characters.
* *After deploy*, Cassey should  check the Google search console and confirm the configuration works in production, too.

## Feedback I'm looking for:
whatever


clottman:
No issue with the two helmets! Helmet supports it explicitly. If there were duplicate tags, (two `<title>` elements for example) the one defined later would take precedence. 

[from the docs](https://www.npmjs.com/package/react-helmet)
helmet docs excerpt about multiple helmet tags

PR #1191, opened 2020-01-29
lengthen HSTS maxAge

whimsicallyson:
## Links
* https://app.clubhouse.io/glitch/story/7707/add-hsts-header-to-traffic-on-glitch-com

## Changes:
* Updated maxAge to a month
* Removed Jenn's STS header so that we're not competing

Next step is to wait a month to make sure everything's good here, then bump to two years and submit to hstspreload!


server/routes.js
```diff
@@ -30,7 +30,6 @@ module.exports = function(EXTERNAL_ROUTES) {
     // security headers added by jenn to get mozilla observatory score up
     response.header('X-XSS-Protection', '1; mode=block');
     response.header('X-Content-Type-Options', 'nosniff');
-    response.header('Strict-Transport-Security', 'max-age=15768000');
```
clottman: Weird! When I ran a checker on glitch.com before making the ticket for this, it wasn't finding this header. 🤔

